PR #130, opened 2011-04-20
added alert, confirm, prompt to browser

timendum:
I know alert, confirm, prompt are not part of the DOM, but all browsers handle them


ianwalter:
I agree with @istvan-antal and @timendum
Putting confirm in `devel` makes no sense. It's a user interface.

lukeapage:
@ianwalter you are free to add alert and console to your globals if you are happy your project using them for user interface.

ianwalter:
@lukeapage I was speaking specifically about `confirm` although you could make the same argument for `alert` and `prompt`. They are browser-supported user interface APIs so I think putting them under the `browser` flag makes sense. Putting `console` under `devel` makes sense since its an API used by developers and doesn't have full browser support.

PR #1538, opened 2014-02-19
Make `arguments` not a reserved keyword.

yungsters:
This patch changes the handling of `arguments` so that it is not considered a
reserved keyword. It is not listed as a reserved keyword on either MDN or the
ES5.1 specification: http://www.ecma-international.org/ecma-262/5.1/#sec-7.6.1

I verified that the following code works in IE 8, Chrome 32, and Firefox 26.

``` javascript
function arguments() {
  return arguments[0] + arguments[1];
}
var object = {};
object.arguments = arguments(1, 2);
object.arguments === 3;
```

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2159,16 +2159,17 @@ var JSHINT = (function () {
   delim(":").reach = true;
   delim("#");
 
+  type("arguments", function (x) {
+    if (state.directive["use strict"] && funct["(global)"]) {
+      warning("E008", x);
+    }
+  }).identifier = true;
+
   reserve("else");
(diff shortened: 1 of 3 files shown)


metasansana:
Hello,

Maybe the warning should change at least, currently I see:

```sh
Expected an identifier and instead saw 'arguments' ( a reserved word).
```

when I use arguments as a parameter but `arguments` is not a keyword. 
Perhaps something like "Warning value of arguments replaced by function parameter" ? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords

PR #1647, opened 2014-04-29
Closes #1081 by adding the `anonfuncs` option

### LINKED ISSUE #1081: Optionally warn about anonymous functions
When using anonymous functions instead of named function expressions or function declarations it will make stacktraces and code less readable

It would be nice if jshint could warn about them.

hughes:
This option allows anonymous (unnamed) functions.
If false, anonymous functions will raise a warning (W126).


Starefossen:
+1

Issue #1462, opened 2014-01-07
Regression: referencing the `const`-defined variable can be done in a code before its declaration


mzgol:
Take the following code (from https://gist.github.com/mzgol/8305487):
```js
(function () {
  'use strict';
  const o = {
    f: function () {
      console.log(a);
    }
  };
  const a = 'aaa';
  o.f();
})();
```

In both IE11 & Chrome Canary with experimental JS flag on this code logs `'aaa'`. Referencing the `const`-defined variable in a code before its declaration is valid in ES6 as long as this code doesn't run before the declaration (I asked @rwaldron for confirmation).

jsHint used to not warn against this code when both `esnext` and `unused` were turned on (simply because `unused` didn't work for `const`-defined variables) but currently it errors on valid code, i.e. this is a regression.

lukeapage:
@jugglinmike I think this is no longer an error on master and can be closed

mzgol:
> @jugglinmike I think this is no longer an error on master and can be closed

I pasted this code into http://jshint.com/ with `/* jshint esnext: true */` added at the beginning and in the summary I see:

```
One unused variable
10	a
```

lukeapage:
@mzgol sorry you are right, I mis intrepretted the issue - there are multiple const/let scope bug reports which overlap considerably.

Re-reading through it again I see it is not fixed on master, but is fixed in my PR #2344 and duplicated in #2292

PR #1682, opened 2014-05-16
Allow const/let/var above declaration if not used.

mzgol:
(this is a re-opened PR #1658 against `master` instead of the removed `2.x` branch)

Ref gh-1462

@valueof I made it work for `const` but I have no idea how to make it work for `let`. From what I see, `var` & `const` are handled by one data structure (though with weirdnesses like `var` being written as `unused` when not used and `const` being just `const` in such a case just with a special `unused` property written in a different structure that is ignored in the `var` case...) but `let` is handled in a completely different way.

I spent some time looking at the code but I cannot figure out what's exactly going on here... Could you help me make it work for `let` as well?

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4847,6 +4847,8 @@ var JSHINT = (function () {
 
             if (context[name] === "unused")
               context[name] = "var";
+            else if (context[name] === "const" && getprop(context, name, "unused"))
+              setprop(context, name, { unused: false });
             else if (context[name] === "unction")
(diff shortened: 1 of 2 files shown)


lukeapage:
@jugglinmike this PR is now invalid and I think the issue related is fixed too

jugglinmike:
[As noted](https://github.com/jshint/jshint/issues/1462#issuecomment-100513693) in gh1462, the related issue is not resolved.

PR #1791, opened 2014-08-08
Added basic handling of (abandoned) ECMA 4 type annotations.

clstopher:
These are not allowed in standard Javascript or in later ECMA
versions, but some custom ECMA implementations do allow them.

This commit doesn't add any additional checking of types, it
just allows scripts that do have type annotations in them to be
checked for other standard errors/warnings.

I need this support  for a project I am working on, and can work
off of my own fork, however I thought maybe others could use it
too.

I was not able to find a formal definition of type annotations,
so I based them on the code I have.

Some examples:

``` javascript
function add(x : Number, y : Number) : Number {
  var sum : Number = x + y;
  return sum;
}
for (var i : Number = 0; i < 10; i++) {
  // do something.
}
```

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -127,6 +127,7 @@ var JSHINT = (function () {
         lastsemic   : true, // if semicolons may be ommitted for the trailing
         loopfunc    : true, // if functions should be allowed to be defined within
         expr        : true, // if ExpressionStatement should be allowed as Programs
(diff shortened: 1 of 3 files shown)


toddbluhm:
It would be really great to not have jshint mark this as an error. Is this still a rejected feature?

jugglinmike:
Yes

PR #1814, opened 2014-08-21
Multi-line template literal test case.

KevinGrandon:
I noticed that a function which returns a template literal throws a linting error. I don't think this should be the case. Oddly enough it works properly if you put an interpolated variable ${} in the first line of the literal. The following errors are thrown when I run the attached test case:

```
  [bold]{Line 4, Char 9} Missing semicolon.
  [bold]{Line 6, Char 1} Expected an assignment or function call and instead saw an expression.
  [bold]{Line 7, Char 2} Unnecessary semicolon.
```

@caitp @rwaldron - Do you guys have any pointers of where to look into to fix this?

### DIFF
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -775,6 +775,14 @@ exports.testES6TemplateLiteralsUndef = function (test) {
   test.done();
 };
 
+exports.testES6TemplateLiteralsMultiLine = function (test) {
+  var src = fs.readFileSync(__dirname + "/fixtures/es6-template-literal-multiline.js", "utf8");
+  TestRun(test)
+    .test(src, { esnext: true });
+
+  test.done();
+};
+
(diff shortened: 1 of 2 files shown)


spenceralger:
@caitp I am still running into this issue because I am returning the template literal.

output:
```sh
$ jshint -v
jshint v2.6.0
$ jshint ~/Desktop/test.js
/Users/spencer/Desktop/test.js: line 12, col 9, Missing semicolon.
/Users/spencer/Desktop/test.js: line 14, col 1, Expected an assignment or function call and instead saw an expression.

2 errors
```

test code:
```js
/* jshint esnext:true */

var foo = 'bar';

// fine
var one = `
  ${foo}
`;

// failes
var two = (function () {
  return `
    ${foo}
  `;
}());
```

caitp:
i'll take a look tonight

PR #1909, opened 2014-10-18
Support `--prereq` CLI options

### LINKED ISSUE #1685: ERROR: Unknown option --prereq
I couldn't get it working in v2.5.0, but 2.5.1 doesn't work either.

screen shot 2014-05-16 at 4 03 21 pm

What should I do to troubleshoot this? The [CLI docs](http://jshint.com/docs/cli/) make it look like `--prereq` is supported, but I can't find any other references to it.

I have a large application that makes use of multiple `namespace.js` files throughout the project to declare globals – how do I pass this to JSHint? Thanks!

satchmorun:
Since [the docs][1] mention it, and it's already supported via [config][2],
I thought I'd just pop the CLI support in.

This PR make `jshint` allow specifying the `--prereq` option on the command
line, accepting a comma-separated list of paths.

Test-related notes:

- Fixes the existing `testPrereq` test to check the right exit code. In its
  current incarnation, it could never fail (even when `lint` didn't pass)
- Added 2 tests for specifying the `--prereq` options from the command line.

Implementation notes:

I did the simplest thing possible (just merging the CLI `prereq` options in
with the config ones in the `run` method in `cli.js`). It's a little bit
awkward because there are 2 paths to `lint`:

1. `opts.useStdin`
2. `!opts.useStdin`

And each loads the config in a slightly different way. So the `mergeCLIPrereq`
function has to be in 2 code paths. So, slightly awkward, but not too terrible.

Fixes #1685.

[1]: http://jshint.com/docs/cli
[2]: https://github.com/jshint/jshint/blob/master/src/cli.js#L451-458

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -15,6 +15,13 @@ var defReporter       = require("./reporters/default").reporter;
 var OPTIONS = {
   "config": ["c", "Custom configuration file", "string", false ],
   "reporter": ["reporter", "Custom reporter (<PATH>|jslint|checkstyle)", "string", undefined ],
+  "prereq": [
+    "prereq",
+    "Comma-separate list of prerequisite (paths). E.g. files which include" +
+    "definitions of global variabls used throughout your project",
+    "string",
+    ""
+  ],
   "exclude": ["exclude",
     "Exclude files matching the given filename pattern (same as .jshintignore)", "string", null],
   "exclude-path": ["exclude-path", "Pass in a custom jshintignore file path", "string", null],
@@ -593,6 +600,12 @@ var exports = {
     var results = [];
     var data = [];
 
+    function mergeCLIPrereq(config) {
+      if (opts.prereq) {
+        config.prereq = (config.prereq || []).concat(opts.prereq.split(/\s*,\s*/));
+      }
+    }
+
     if (opts.useStdin) {
       cli.withStdin(function (code) {
         var config = opts.config;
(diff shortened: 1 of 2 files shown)


edslocomb:
Can I do anything to help move this along?

Support for prereqs on the command line is really rather important for a component that fits in the middle of a toolchain; e.g. when the list of dependencies is generated by a precompiler (or multiple precomilers), and jshint output sent along to a text editor.

adrianmoisey:
:bell: 
Would be rad if this got merged.

src/cli.js
```diff
@@ -593,6 +600,12 @@ var exports = {
     var results = [];
     var data = [];
 
+    function mergeCLIPrereq(config) {
+      if (opts.prereq) {
+        config.prereq = (config.prereq || []).concat(opts.prereq.split(/\s*,\s*/));
```
lukeapage: @jugglinmike 
this only supports whitespace where it has already been escaped. e.g. In my tests..

```
$ jshint a.js --prereq b.js ,c.js
```
will try and open `,c.js` but
```
$ jshint a.js --prereq "b.js, c.js"
```
and
```
$ jshint a.js --prereq b.js,\ c.js
```
will treat it as one argument.

So I think your comment is already resolved through pre-processing before it gets into this code.

Are you happy with that? I've resolved the other code comments.

lukeapage:
moved to https://github.com/jshint/jshint/pull/2474

src/cli.js
jugglinmike: Yeah, I think I was thrown off by the way `cli.interpret` works in the tests.

PR #1967, opened 2014-11-10
Don't delete lines following lex'd `ignore:end`. Fixes gh-1879

### LINKED ISSUE #1879: Ignore blocks confuse dead code analysis
It seams deadcode analysis doesn't take ignore blocks into account:

```javascript
function Foo() {   // unmatched '{'
    return ;
    // jshint ignore: start
    return [];  // unreachable 'return' after return
    // jshint ignore:end
} // Expected an identifier and instead saw '}'
```

caitp:
Previously, lines would be cleared (set to the empty string) even after `ignore:end`
is found, if the comment was not processed first.

This change performs some minor comment processing in the lexer itself in order to
solve this problem.

This is a hack, ideas are welcome.

Closes #1879

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -358,6 +358,7 @@ Lexer.prototype = {
     var rest = this.input.substr(2);
     var startLine = this.line;
     var startChar = this.char;
+    var self = this;
 
     // Create a comment token object and make sure it
     // has all the data JSHint needs to work with special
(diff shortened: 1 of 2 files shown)


lukeapage:
@caitp if you don't mind I'll grab this branch and build on top of it?

I just hit a problem with destructuring where it is using peek more and the lexer can throw errors on peek, so it works both ways.

I'd like to a) abstract option parsing to a new file, so it isn't repeated b) apply @jugglinmike suggestions, so removing ignore from state and jshint

lukeapage:
> I'd like to a) abstract option parsing to a new file, so it isn't repeated b) apply @jugglinmike suggestions, so removing ignore from state and jshint

on second thought (a) is alot more work than its worth

lukeapage:
moved to https://github.com/jshint/jshint/pull/2512

PR #1984, opened 2014-11-18
Don't warn fat arrow params used out of scope. Fixes #1983

### LINKED ISSUE #1983: incorrect "used out of scope" warning with arrow function
Test case:

```
'use strict';

(() => {
    for (var i = 0; i < 10; i++) {
        var x;
    }

    var arrow = (x) => {
        return x;
    };

    arrow();
})();
```

jshint 2.5.10 generates the following error:

```
test.js: line 8, col 18, 'x' used out of scope.
```

caitp:
This is a hack, but I couldn't find a quicker fix. There are probably some things
that ought to be done better here.

Fixes gh-1983

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2078,7 +2078,9 @@ var JSHINT = (function () {
         case "function":
         case "var":
         case "unused":
-          warning("W038", state.tokens.curr, v);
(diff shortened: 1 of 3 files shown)


rwaldron:
I landed the alternate version of this

PR #2010, opened 2014-11-27
AppVeyor: remove global cache and add node 0.11

XhmikosR:
We don't install any global modules.

/CC @jugglinmike


XhmikosR:
*Bump*

It seems there are a few tests that fail with Node.js 0.11 on Windows.

XhmikosR:
We need to get this in along with #1932 and then try to fix the remaining issues...

PR #2049, opened 2014-12-20
Test cleanup

jugglinmike:
This improves JSHint's CLI unit tests a bit. Aside from using Nodeunit (via `setUp` and `tearDown` method) and Sinon more effectively (via the "sandbox" feature), it also reduces shared state between test runs. The shared state forced tests to include irrelevant logic, but (more importantly) it could have led to false negatives.

[We first discovered this problem back in October](https://github.com/jshint/jshint/pull/1900#discussion_r18750300), but it's only with [the introduction of code coverage reporting](https://github.com/jshint/jshint/pull/2007) that I felt comfortable doing this refactoring.

### DIFF
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -3,15 +3,30 @@
 var path  = require("path");
 var shjs  = require("shelljs");
 var sinon = require("sinon");
-var cli   = require("../src/cli.js");
+
+var cliPath = path.resolve(__dirname, "../src/cli.js");
+var cli;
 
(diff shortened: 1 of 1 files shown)


rwaldron:

jugglinmike:
@rwaldron Reworded. The "[[DOCS]]" tag seemed most appropriate here, but maybe we should add "[[TESTS]]" to the list?

rwaldron:
Agreed.

cc @caitp

PR #2061, opened 2014-12-30
[[FIX]] Allow array, grouping and string form to follow spread operator in function call args.

rwaldron:


rwaldron:

rwaldron:
@caitp Updated to new commit message format and changes per your last review

caitp:
Looks good to me --- last nit might be adding a "TODO: allow all assignment expressions when parsing enables this" note or something

src/jshint.js
```diff
@@ -2260,7 +2260,12 @@ var JSHINT = (function () {
     if (!state.option.esnext) {
       warning("W119", this, "spread/rest operator");
     }
-    if (!state.tokens.next.identifier) {
+
+    if (!state.tokens.next.identifier &&
+        state.tokens.next.type !== "(string)" &&
+          state.tokens.next.value !== "[" &&
+            state.tokens.next.value !== "(") {
```
jugglinmike: I think this can just be `!checkPunctuators(state.tokens.next, ["[", "{"])`

rwaldron: `!checkPunctuators(state.tokens.next, ["[", "("])` yep.

src/jshint.js
```diff
@@ -2260,7 +2260,13 @@ var JSHINT = (function () {
     if (!state.option.esnext) {
       warning("W119", this, "spread/rest operator");
     }
-    if (!state.tokens.next.identifier) {
+
+    // TODO: Allow all AssignmentExpression
+    // once parsing permits.
+    if (!state.tokens.next.identifier &&
+        state.tokens.next.type !== "(string)" &&
+          !checkPunctuators(state.tokens.next, ["[", "("])) {
```
jugglinmike: [`checkPunctuators` ensures that the `type` is `"punctuator"`](https://github.com/jshint/jshint/blob/08c70e632f889feb8481acdc41e541ca49edfad4/src/jshint.js#L4476), so the explicit type check is now unnecessary. I don't *think* that it should be necessary to check if the token is an identifier, either since if it's a punctuator it can't be an identifier.

rwaldron:
@caitp extended note: https://github.com/jshint/jshint/pull/2061/files#diff-0

Issue #2064, opened 2014-12-31
"singleGroups" option gives spurious warnings for negative instanceof condition


Qantas94Heavy:
Using JSHint v2.5.11, when I try the following code:

```js
// jshint singleGroups:true
if (!(a instanceof b)) alert(1);
```

It will give out a spurious warning that "Grouping operator is unnecessary for lone expressions". It is in fact necessary, because the unary negation operator has a higher precedence than the `instanceof` operator.

Krinkle:
Should we exclude `!!` (boolean cast) as well?

```js
      4 |return !!( window.JSON && JSON.stringify && JSON.parse );
                                                                 ^ Grouping operator is unnecessary for lone expressions.
```

rwaldron:
> Grouping operator is unnecessary for lone expressions

In the example you gave, this error just seems wrong.

jugglinmike:
@Krinkle Definitely. @rwaldron is right, that is another bug. It's covered by my patch in gh-2065, but [I've added another regression test to be safe](https://github.com/jshint/jshint/commit/62cdf5bcdedd4efc1d67dfb3bffaae54c4bebcdd). Thanks!

gagle:
Same issue here:

```javascript
//Warning
path.resolve(alias.dir, alias.path) + (obj.trail ? '/' + obj.trail : '');

//No warning & invalid
path.resolve(alias.dir, alias.path) + obj.trail ? '/' + obj.trail : '';
```

dsl101:
Another example?

    var a = 2;
    var b = 3;
    var resultStr;

    // Correct answer (5), but jshint gives warning:
    resultStr = 'Total: ' + (a + b) + '.';
                                    ^ Grouping operator is unnecessary for lone expressions.

    // Incorrect answer (23), but jshint gives no warning:
    resultStr = 'Total: ' + a + b + '.';

Qantas94Heavy:
Also, should this trigger the warning or not?

```js
// jshint singleGroups:true
if ( (foo = 1) ) alert(1);
```

This is usually used to indicate that the assignment was deliberate, but currently triggers a warning with the `singleGroups` option enabled even though this silences the "assignment instead of conditional expression" warning.

gagle:
I still receive the warning from my previous post.

PR #2065, opened 2014-12-31
Fix single groups

jugglinmike:
This approach is much cleaner because it honors the binding power of expressions--note that it is no longer necessary to special-case "fat arrow" bodies--that falls out naturally.

@rwaldron In a separate commit, I enabled the option in some of our integration tests. In the case of Backbone.js, I had to white-list three instances where I believe the grouping operator is unnecessary. Can you confirm that? (I didn't enable this option for jQuery, Lodash, or Prototype because those libraries use grouping operators quite liberally, so we'd have to maintain giant white lists of expected errors for each.)

Resolves gh-2064

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -846,8 +846,7 @@ var JSHINT = (function () {
   // They are elements of the parsing method called Top Down Operator Precedence.
 
   function expression(rbp, initial) {
-    var left, isArray = false, isObject = false, isLetExpr = false,
-      isFatArrowBody = state.tokens.curr.value === "=>";
+    var left, isArray = false, isObject = false, isLetExpr = false;
 
     state.nameStack.push();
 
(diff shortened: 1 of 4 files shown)


jugglinmike:
That last commit is meant to be a `fixup!`. Please let me know when this is good to land, and I'll squash accordingly.

rwaldron:

jugglinmike:
@rwaldron Squashed and re-worded. Again, "[[DOCS]]" doesn't seem quite right for that commit, and I'm happy to change it to "[[TESTS]]" if you agree.

caitp:
+1 for [[TEST]]

rwaldron:
@jugglinmike go ahead and update to [[TEST]] (or [[TESTS]]?)

jugglinmike:
Done!

jugglinmike:
**Hold off on this**. The following form still incorrectly triggers warnings:

```js
if (a + (b + c)) {}
```

I'm on it, though...

jugglinmike:
Superseded by gh-2109

PR #2069, opened 2015-01-02
Add additional test cases to increase code coverage

hpshelton:
This should hit additional corner cases in lex.js, trivially increasing code coverage in scanNumericLiteral() and scanEscapeSequence(). Please verify that the test case behavior itself is correct since I'm asserting the actual behavior not necessarily the expected behavior for these cases.


rwaldron:

rwaldron:
Thanks!

PR #2072, opened 2015-01-03
Using an option object to pass doFunction params

sebv:
Avoids passing null or undefined parameters everywhere.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1172,7 +1172,7 @@ var JSHINT = (function () {
       nobreaknonadjacent(state.tokens.prev, state.tokens.curr);
 
       this.left = left;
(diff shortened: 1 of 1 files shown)


Globegitter:
Any reason this is not being merged? Would love to see support for this and more importantly https://github.com/jshint/jshint/pull/2077

jugglinmike:
@Globegitter As discussed in #2077, that feature is not under consideration for inclusion in JSHint at this time; it was shared as an exploration into a future plugin system.

nicolo-ribaudo:
I think this PR can be closed, as now [`doFunction` requires an object as parameter](https://github.com/jshint/jshint/blob/master/src/jshint.js#L2907-L2925).

lukeapage:
good spotting..

PR #1979, opened 2014-11-16
Added optional asyncawait support on top of esnext.

sebv:
Seems to work, passes the test suite, works well with arrow function and classes. 

Configure like in the following:
```
"esnext": true
"experimental": ["asyncawait"]   or "experimental": ["asyncawait", "asyncreqawait"]
```

To try it:
```
npm install -g -f https://github.com/sebv/jshint/archive/esnextnext.tar.gz
```

EDIT updated options.

PR #2077, opened 2015-01-04
AsyncAwait abstracted into experimental.js

sebv:
Depends on #2072.

This is a rewrite of #1979 more in line with the current structure.

I've moved all the AsyncAwait code to a separate file and added some hook to the core, the idea beng to extract plugins later.

It's  also working against @rwaldron disambiguation specs: 
```
function async() {
}

let a = async () => 1;

let b = async _ => 1;

let c = async async => 1; // that's an async arrow with a single formal parameter named "async"

async();

async 
();

async function async() {

}

async function async
()
{}

function async
()
{}
```

All the fixes applied to #1979 are included except the export generator, will do a separate PR for that and pull it later from master.

To install do:
```
npm install -gf https://github.com/sebv/jshint/archive/experimental.tar.gz
```


src/jshint.js
```diff
@@ -883,11 +892,15 @@ var JSHINT = (function () {
...
 
+    expressionHook.parse();
+
     if (initial) {
       funct["(verb)"] = state.tokens.curr.value;
     }
 
-    if (initial === true && state.tokens.curr.fud) {
+    var runFud = initial === true && state.tokens.curr.fud;
+    runFud = expressionHook.updateRunFud(runFud);
```
rwaldron: why not...

```js
var runFud = expressionHook.updateRunFud(initial === true && state.tokens.curr.fud);
```

src/jshint.js
```diff
@@ -102,6 +103,11 @@ var JSHINT = (function () {
     extraModules = [],
     emitter = new events.EventEmitter();
 
+  var experimental = new Experimental(addlabel, advance, blockstmt, doFunction, error,
+    expression, function () { return funct; }, function() { return inblock; },
+    isEndOfExpr, nobreaknonadjacent, nolinebreak, optionalidentifier, peek, prefix,
+    warning, warningAt);
```
rwaldron: my eyes are bleeding

<3 :D

rwaldron:
I just left you a funny image and a comment about a specific line, but I'm going to stop there because I don't want my review to be mistaken for tacit approval of this patch. I'm concerned about the amount of code that exists here, vs. the amount of discussion that lead to it (a lot vs. basically none).

So I'm putting on the breaks here.

rwaldron:
I appreciate the effort here, but I think we need to take a step back for a moment.
In reviewing your previous pull request for this feature, I mentioned my interest in implementing this via some sort of future JSHint "plugin" system. I applaud your efforts to implement the plugin system on your own, but I can't in good conscious accept your implementation as-is. Here's why:
Exposing JSHint's internals, while necessary for such a system, will take a lot of careful consideration. Ideally, it will start with a formal design process where no code is written. Here, we would get input from a whole bunch of stakeholders and formally scope the work, evaluate tradeoffs, and identify areas for future extensions.

Given the effort that clearly went in to this, I can imagine this feedback would be a little disappointing. I didn't mean to suggest that you take on this challenge alone, though! At the very least, this latest implementation of the "async/await" feature most clearly demonstrates your specific needs for the plugin system, and it will be really helpful in the design process.

src/jshint.js
sebv: I know but how do you expose method within the closure otherwise? JSHINT need to use prototype, but cannot really write this on my own.

sebv:
I agree, never expected this to be merged as it is, just wanted to separate the code from the previous request from the core to see how it could possibly work while still offering a solution for people using `async`. I think from here the only way forward to remove this Experimental madness is to objectify JSHINT unless someone see a better solution.

Is it ok to do that, probably lot of pain but not too hard?

sebv:
Another solution would be to use the same pattern as `api` for plugins. Would be a good intermediary step no matter what.

sebv:
Added a plugin logic.

stefanpenner:
@rwaldron now that chakra implements https://twitter.com/reybango/status/649306777218097152 can we revisit this?

PR #2087, opened 2015-01-07
Use `path.join` properly and in more places.

XhmikosR:
/CC @caitp

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -674,7 +674,7 @@ var exports = {
     cli.options = {};
 
     cli.enable("version", "glob", "help");
-    cli.setApp(path.resolve(__dirname + "/../package.json"));
+    cli.setApp(path.join(__dirname, "/../package.json"));
 
     var options = cli.parse(OPTIONS);
     // Use config file if specified
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -202,7 +202,7 @@ exports.group = {
   },
 
   testOverrides: function (test) {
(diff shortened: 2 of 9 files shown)


XhmikosR:
Updated one more time...

XhmikosR:
Bump.

Is anyone still interested in my PRs? Or should I drop them?

jugglinmike:
I'm still reluctant to make this change for the same reasons I cited when you initially proposed it: https://github.com/jshint/jshint/pull/1932 .

Restating: this change (and also gh-2298) is not motivated by a real-world problem. While I think your commitment to propriety is commendable, project maintainers have to balance the benefits of refactoring against concerns for stability. Please understand we have a responsibility to keep things stable (even though we may not like certain aspects of the code base, either!).

For instance, you can see how a seemingly-innocuous change like 2fc1e2377ac590eb0455f0a2b0ffef76f3f0b0ef caused actual problems without adding any value to the project itself: https://github.com/jshint/jshint/issues/2330 . I very much regret this and want to avoid situations like these whenever possible.

I can't speak for the other maintainers, though.

XhmikosR:
Well,  this change has nothing to do with the one you linked. In fact, this
is the proper way. Not said by me,  just check node docs.

Secondly, tests pass just fine plus I've tested this thoroughly.

Thirdly,  I was referring to all my PRs not this one only.

Don't get me wrong, but you are overthinking this specific patch.

PR #2100, opened 2015-01-13
[[FEAT]] `trailingcomma` warns for missing trailing commas

jacksonrayhamilton:
Hello JSHint team! I'm currently considering migrating from JSLint due to a recent "discovery" I made regarding trailing commas. I not only find their usage acceptable, but I have evidence that they are actually _better_ than the ubiquitous list with the trailing comma omitted. I'm not sure if this is widely-recognized in the JavaScript community, considering the dreadful stigma trailing commas have inherited from the ES3 days, but those days are long past... right?

Missing trailing commas in array and object literals can result in undetectable visual ambiguities--the same visual ambiguities that lead to bugs when a programmer relies on automatic semicolon insertion. Therefore, it might be a good practice to always use them, so this option provides warnings in their absence.

For example, this code might have worked last Tuesday:

``` js
[
    b + c
].forEach(print);
```

But if one adds an element to the array and forgets to compensate for the missing comma after the final element, no syntax error is thrown, and a linter cannot determine if this was a mistake or an intentional function invocation:

``` js
[
    b + c
    (d + e)
].forEach(print);

// Gets interpreted as:
// [
//    b + c(d + e)
// ].forEach(print);
```

If one always appends a list item with a comma, this ambiguity cannot occur:

``` js
[
    b + c,
].forEach(print);

[
    b + c,
    (d + e),
].forEach(print);
```

Does this look familiar to [the spec's justification](http://www.ecma-international.org/ecma-262/5.1/#sec-7.9.2) for the usage of semicolons?

> The source
> 
> ```
> a = b + c
> (d + e).print()
> ```
> 
> is not transformed by automatic semicolon insertion, because the parenthesised expression that begins the second line can be interpreted as an argument list for a function call:
> 
> ```
> a = b + c(d + e).print()
> ```
> 
> In the circumstance that an assignment statement must begin with a left parenthesis, it is a good idea for the programmer to provide an explicit semicolon at the end of the preceding statement rather than to rely on automatic semicolon insertion.

If the aforementioned suggestion could be used to justify the strict usage of semicolons to terminate statements, then the same advice could be applied to trailing commas. The above paragraph could easily be re-purposed to say, "In the circumstance that an **array element** must begin with a left parenthesis, it is a good idea for the programmer to provide an explicit **comma** at the end of the preceding statement rather than to rely on **a syntax error**."

The more I think about it, the sillier it seems that anyone programming in an ES5 environment would continue to follow the error-inducing practice of omitting trailing commas. Trailing commas should be the standard, not the other way around; but I guess we are all so broken from the <=IE8 days that we have largely shied away from this new feature.

So, I have added option `trailingcomma` to detect missing trailing commas. If their omission is warned against, the programmer can circumvent the aforementioned ambiguities and reduce the likelihood of error.

If this option is `true`, a warning will be thrown if a "," does not occur before the closing "]" or "}" in an array or object literal. If option `es3` is also `true`, this option does nothing, because otherwise the options would contradict each other, and the `es3` mandate is more pressing.

There is no inverse of this option, that would warn if an ES5 programmer omitted a trailing comma. Omission of a trailing comma in ES5 provides no benefit, unlike option `trailingcomma`, which actually helps prevent bugs. Such an inverse feature would be considered "stylistic." Do not disregard `trailingcomma` on the premise of it being a "style" option; it is just as "stylistic" as the default detection of semicolons. If these things are "style-related," then they are good styles, because they serve the linter's purpose of "detecting errors and potential problems."

Thanks for your consideration. I would love to see this land.


tests/unit/options.js
```diff
@@ -1803,3 +1803,36 @@ exports.elision = function (test) {
...
+    "var a = [];",
+    "var b = [1];",
+    "var c = [1,];",
+    "var d = [1,2];",
+    "var e = [1,2,];",
+    "var f = {};",
+    "var g = {a: 1};",
+    "var h = {a: 1,};",
+    "var i = {a: 1, b: 2};",
+    "var j = {a: 1, b: 2,};"
```
jacksonrayhamilton: Before you point out the irony in me *not* using a trailing comma here, I was simply following your contribution guidelines and keeping consistent with the dominant style in this file.

rwaldron:
I like this, but I want to hold off on landing until we have some time to see how it fares against existing code. Appreciate your patience

KylePDavis:
Seems worth mentioning that the `jscs` tool supports a similar check but allows a couple of exceptions for a couple of simple cases: `ignoreSingleValue` and `ignoreSingleLine`
Here is the section in their docs:   http://jscs.info/rules.html#requiretrailingcomma

jugglinmike:
Hi @jacksonrayhamilton, this looks good to me! Thanks for the patch (and
especially for the detailed write-up). There are just a few trivial conflicts
to address:

- The `inES5` method is now defined on the `state` object directly
- We'll want to invoke that check *without* setting the first argument to
  `true` so that ES6+ environments may also opt in to this functionality
- The warning ID will have to be updated to accommodate the warnings that have
  been introduced since you first submitted this
- This should be enforced for array initializers in ES3 (as I somewhat
  redundantly mentioned in an in-line comment)

That said, it's been quite a while since you offered this patch, and I can
imagine you've moved on to bigger and better things. Say the word and I'll do
the follow-up work on your behalf.

src/jshint.js
```diff
@@ -2767,6 +2767,9 @@ var JSHINT = (function() {
           break;
         }
       } else {
+        if (state.option.trailingcomma && state.option.inES5(true)) {
```
jacksonrayhamilton: That check was accommodate [IE < 9 trailing comma bugs](http://stackoverflow.com/a/11306770/1468130). I guess it isn't really a matter of the ES version, instead whether old IE bugs need to be taken into account. Old IE == ES3 to me, I guess.

jacksonrayhamilton:
Hi @jugglinmike, thanks for getting back.

Your assumption is correct, I've moved on from JSHint - onto a newer linter, with a horribly large set of useless options, where the defaults are not useful, and the config files are enormous. (ESLint.) I highly prefer a linter with good opinions (like JSHint), rather than no opinions (like ESLint). But, I made the mistake of tweaking that newer linter to my heart's content, so I may as well enjoy those configurations now.

These days, I'm also a little less concerned with mistakes like the one in `trailingcomma`. I don't think I even check for that error, because I just don't care if all my arrays and objects are "perfect" anymore. This feature may be more popular now that we have [trailing commas in parameters and calls](https://github.com/tc39/proposal-trailing-function-commas)... but there is only so much time and energy we can exert in our jobs, and I've come to realize that this sort of stuff is a complete waste of it. So, I can't be the one to push it forward anymore, and I'd prefer to see it die now.

jugglinmike:
I'm sorry you've become so frustrated with the state of things! I personally
think there is still value in the work you've done here, so I'll take up the
cause. I made the suggested changes and re-submitted as gh-3090. Thanks again
for your contribution; I hope you can see the value in it again some day.

PR #2101, opened 2015-01-13
[[TEST]] Introduce jscs for style guide enforcement

mikesherov:


.jscsrc
```diff
@@ -0,0 +1,45 @@
+{
+    "requireOperatorBeforeLineBreak": true,
+    "maximumLineLength": {
```
jugglinmike: Can you remove the `maxlen` option from our `.jshintrc`?

.jscsrc
```diff
@@ -0,0 +1,45 @@
...
+      "value": 100,
+      "allowComments": true,
+      "allowRegex": true
+    },
+    // disabled due to https://github.com/jscs-dev/node-jscs/issues/915
+    // "validateIndentation": 2,
```
jugglinmike: Think this will be a tough one to fix? I'd like to also remove `indent` from our `.jshintrc`, but we could do it in a future commit if it's non-trivial.

mikesherov: Not tough. Will do shortly.

.jscsrc
mikesherov: addressed.

.jscsrc
mikesherov: I'd like to land this as is, and remove later once I address the remaining issues.

src/lex.js
```diff
@@ -619,7 +619,7 @@ Lexer.prototype = {
     }.bind(this);
 
     function removeEscapeSequences(id) {
-      return id.replace(/\\u([0-9a-fA-F]{4})/g, function(m0, codepoint) {
+      return id.replace(/\\u([0-9a-fA-F]{4})/g, function (m0, codepoint) {
```
caitp: I never liked the space between `function` and `(` for anonymous functions :<

jugglinmike: Same here!

mikesherov: I change the rules later maybe?

rwaldron: > I never liked the space between function and ( for anonymous functions :<

Yeah... I hate that.

rwaldron: As long as we fix all of them immediately, this is ok for the moment

PR #2122, opened 2015-01-21
[[DOCS]] Document deprecated options

jugglinmike:
These options have long been considered deprecated by JSHint
maintainers. The intention to remove should be communicated in the
source code itself to give advance notice to end users.


jugglinmike:
I'll be following up with an extension to the website here: https://github.com/jshint/jshint.github.io/pull/8

PR #2127, opened 2015-01-23
Fix `.jshintrc` resolution

jonnyreeves:
The current layout of the `findConfig` appears to contain an error which results in no `.jshintrc` being found if the user does not have one of the defined 'home' environment variables; I'm sure 99.9% of people will never run into this, but there is a small percentage of us out there who work with very, very bespoke systems ;)

This is a really tricky one to test as the tests themselves fail if you unset your home environment variable - however this PR does not change any functionality, just makes it a bit more explicit :)

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -74,18 +74,18 @@ function deprecated(text, alt) {
 function findConfig(file) {
   var dir  = path.dirname(path.resolve(file));
   var envs = getHomeDir();
-
-  if (!envs)
-    return home;
-
-  var home = path.normalize(path.join(envs, ".jshintrc"));
-
   var proj = findFile(".jshintrc", dir);
+  var home;
+
   if (proj)
     return proj;
 
-  if (shjs.test("-e", home))
-    return home;
+  else if (envs) {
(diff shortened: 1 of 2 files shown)


jonnyreeves:
On an unrelated note, this project's AppVeyor linting rules / config appear to differ from those applied by the Travis build.

jonnyreeves:
*nudge*

lukeapage:
Your code makes much more sense anyway and a further change could be to not even get the home dir if the project matches something.

lukeapage:
A test would be nice, though I can see why it might be difficult. I can take a look at how a test might be written.

jonnyreeves:
Hey @lukeapage, add a test, rebased against master and squashed.  Let me know if there's anything else I can do to help this land, please :)

lukeapage:
Thanks,  lgtm, will land it in the next couple of days

lukeapage:
merged - https://github.com/jshint/jshint/commit/946af3e96f062f77d34646aab5622999d872dee1

Thanks

Issue #1993, opened 2014-11-20
JS file "for browser-like environments" doesn't seem to work


jscheid:
Hi,

http://www.jshint.com/install/ has a download link "for browser-like environments", which I think is what bower-based environments use, but that ceased to work for me. I currently have jshint hardwired to 2.5.6 in my `bower.json` because newer versions fail to load.

This minimal test case (which uses rawgit.com to get around CORS restrictions)...

``` HTML
<!DOCTYPE html>
<html><body><script src="https://rawgit.com/jshint/jshint/master/dist/jshint.js" type="text/javascript"></script>
```

... gives me (on Chrome 38.0.2125.111):

```
Uncaught Error: Cannot find module 'jshint'
jshint.js:6 s
jshint.js:6 s
jshint.js:6 (anonymous function)
jshint.js:11123 (anonymous function)
jshint.js:11125
```

Same for other recent versions, 2.5.6 was the most recent I could find that still works.

Am I doing something wrong, do I perhaps now need to include another JS file on my page before jshint? Is it perhaps related to #1969?

PR #2128, opened 2015-01-24
Test Browser Build

jugglinmike:
Back in November, we learned that the browser build of JSHint was broken
through issue gh-1993. This was caused when this project's dependency on
Browserify was updated to a new major version. Although this could have been
avoided with a more careful review process, fundamental problems like this can
(and should) be detected automatically.

There are many ways to test the build version of JSHint. The simplest would
involve assuring that, when loaded in a browser, a `JSHINT` function is defined
in the global scope. Even this simple test may have been enough to catch the
regression described above.

Although easy to implement, this test would fail to catch a wide array of bugs
that could come about when bundling the project and executing the bundle in a
browser. We could define some additional tests that assert basic functionality,
but those tests would be redundant with the project's unit tests and would
never provide the same coverage.

This changest introduces a testing process that bundles all of the project's
unit tests and executes them in the browser. Approaching the problem this way
provides the best code coverage possible while minimizing duplication, but it
comes at the expense of simplicity. I've tried to make the new testing
infrastructure as simple as possible, and I've documented the pieces that I
think need documentation.

Beyond simply running the tests, I had a few goals:

- **Report parity with the current unit test suite** - the output of `npm run
  test-browser` is identical to the output of `npm run test-unit`)
- **Ability to run the tests in a local browser for manual debugging** -
  running `npm run browser-test-server` initiates a simple web server whose
  index page executes the unit tests)
- **Avoidance of temporary files** - JSHint is re-built with every test run but
  not written to disk. This precludes mistakes like accidental inclusion of
  test builds in future commits without necessitating special `.gitignore`
  rules.

I think that's about it!

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -3,36 +3,36 @@
 
 "use strict";
 
-var browserify = require("browserify");
-var bundle     = browserify();
 var path       = require("path");
-var version    = require("../package.json").version;
+var build      = require("../scripts/build");
 require("shelljs/make");
 
 var distDir    = path.join(__dirname, "../dist");
-var srcDir     = path.join(__dirname, "../src");
 
 if (!test("-e", distDir))
   mkdir(distDir);
 
-bundle.require(srcDir + "/jshint.js", { expose: "jshint" });
-bundle.bundle(function (err, src) {
-  var web   = distDir + "/jshint.js";
-  var rhino = distDir + "/jshint-rhino.js";
+build("web", function(err, version, src) {
+  if (err) {
+    console.error(err);
+    process.exit(1);
+  }
 
-  [ "/*! " + version + " */",
-    "var JSHINT;",
-    "if (typeof window === 'undefined') window = {};",
-    "(function () {",
-      "var require;",
-      src,
-      "JSHINT = require('jshint').JSHINT;",
-      "if (typeof exports === 'object' && exports) exports.JSHINT = JSHINT;",
-    "}());"
-  ].join("\n").to(web);
+  src.to(distDir + "/jshint.js");
+  console.log("Built: " + version + " (web)");
+});
 
-  ("#!/usr/bin/env rhino\nvar window = {};\n" + cat(web, srcDir + "/platforms/rhino.js")).to(rhino);
-  chmod("+x", rhino);
+build("rhino", function(err, version, src) {
+  var dest;
 
-  echo("Built: " + version);
+  if (err) {
+    console.error(err);
+    process.exit(1);
+  }
+
+  dest  = distDir + "/jshint-rhino.js";
+  chmod("+x", dest);
+
+  src.to(dest);
+  console.log("Built: " + version + " (Rhino)");
 });
--- a/tests/helpers/browser/server.js
+++ b/tests/helpers/browser/server.js
@@ -1,0 +1,213 @@
+"use strict";
+var fs = require("fs");
+var http = require("http");
+var Stream = require("stream");
+var path = require("path");
+var url = require("url");
+
+var browserify = require("browserify");
+var buildJSHint = require(__dirname + "/../../../scripts/build");
+
+var mainPath = path.resolve(
+  __dirname + "/../../../" + require("../../../package.json").main
+);
+var contentTypes = {
+  ".html": "text/html",
+  ".js": "application/javascript"
+};
+
+var streams = {
+  fixtures: function() {
+    var fixtureDir = __dirname + "/../../unit/fixtures";
+    var fixtureStream = new Stream.Readable();
+    fixtureStream._read = fixtureStream.write = function() {};
+
+    fs.readdir(fixtureDir, function(err, files) {
+      var src = "";
+      var fsCache = {};
+
+      if (err) {
+        done(err);
+        return;
+      }
+
+      files.forEach(function(fileName) {
+        var relativeName = "/tests/unit/fixtures/" + fileName;
+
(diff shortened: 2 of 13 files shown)


jugglinmike:
@rwaldron I resolved the Windows issue by cheating--I'm not running the browser build tests in Windows (see the most recent commit for exact details). I'm half joking when I say "cheating", though I think this is acceptable/safe (and not just laziness).

When I first opened this pull request, there were one set of problems with the browser testing process in Windows. Now the issue is completely different. I could try to resolve one (or both) problems, but apparently PhantomJS/Windows integration is a moving target. Who knows when another change will occur and cause JSHint's CI builds to fail.

Keep in mind that we are still testing JSHint in Windows. The omission I'm talking about simply means that we are not testing *in PhantomJS in Windows*. I think it's very unlikely that we will make a change to JSHint that breaks Windows browsers specifically.

If that *is* a concern, it still may be worthwhile to land this as-is and work on Windows browser testing separately. As it stands, this patch only increases test coverage.

jugglinmike:
Actually, **let's hold off on this**. It occurred to me that without support for testing browsers from Windows, our Windows-running *contributors* cannot run the full test suite. It's possible that a contributor would author a patch, see that `npm test node`, and submit; only to find that their change breaks the browser build after the AppVeyor job completes. From there, debugging would be difficult.

I'll take another pass and shore up Windows support. If PhantomJS really is a moving target for that platform, then we'll just pin the devDependency.

jugglinmike:
@caitp @rwaldron I think this is good to go now. What do you think?

bin/build
```diff
@@ -3,36 +3,36 @@
...
+build("rhino", function(err, version, src) {
+  var dest;
+
+  if (err) {
+    console.error(err);
+    process.exit(1);
+  }
+
+  dest  = distDir + "/jshint-rhino.js";
+  chmod("+x", dest);
```
lukeapage: I guess chmod isn't needed because it isn't an attribute that is distributed?
Is it there to make testing easier?

bin/build
```diff
@@ -3,36 +3,36 @@
...
+  src.to(distDir + "/jshint.js");
+  console.log("Built: " + version + " (web)");
+});
+
+build("rhino", function(err, version, src) {
+  var dest;
+
+  if (err) {
+    console.error(err);
+    process.exit(1);
```
lukeapage: I've had problems in the past with process.exit - https://github.com/less/less.js/blob/master/bin/lessc#L48

but not sure if that applies to stderr as well as stdout - I guess so. Also not sure if this is something node has now fixed in some way.

anyway, doesn't matter too much since this is just for developers. Searching the code it seems jshint doesn't already use process.exit anywhere at the moment.

tests/helpers/browser/server.js
```diff
@@ -0,0 +1,213 @@
...
+        ));
+        return;
+      }
+
+      done(null, src);
+    });
+  }
+};
+
+module.exports = function(port, done) {
```
lukeapage: I guess I'm used to node modules generating to a tmp directory, then using an existing http server package to serve that, to avoid writing an http server to run the tests.. but I guess its not that much code... and not having a tmo directory is nice.

lukeapage: there are probably http servers that you can handle requests for, which might simplify the code - did you consider that?

lukeapage:
Have you thought how this solution might fit in with taking advantage of something like sauce labs (which is free for open source) in order to run the tests in multiple browsers as part of the build ? Maybe thats ott for a project that is mainly not used in the browser?

lukeapage:
Tested on windows, works well.

tests/helpers/browser/server.js
jugglinmike: Not really. The majority of the code here is in the build handlers which are application-specific. The server code itself is so minimal (lines 171-201) that I thought it was prudent to write it myself.

jugglinmike:
> Have you thought how this solution might fit in with taking advantage of something like sauce labs (which is free for open source) in order to run the tests in multiple browsers as part of the build ? Maybe thats ott for a project that is mainly not used in the browser?

Since this exposes an HTTP server, I imagine it would be fairly straightforward to configure a Sauce Labs tunnel. It's hard to say "no" to more testing (for me, anyway), but I think it would be good to leave that as an area for future research.

jugglinmike:
@lukeapage Thanks for digging this one up. Let me know if it looks good to you and I'll squash the `fixup!` commits

lukeapage:
yes, LGTM

jugglinmike:
In the time since I originally submitted this pull request, a test has landed in `master` that exposes a bug in PhantomJS. Now that I have rebased, the build fails consistently. I'll follow up with a workaround shortly.

jugglinmike:
Rebased onto `master`. Thanks to gh-2559, all tests are now passing!

lukeapage:
To re-iterate, this is good to land from my point of view.

lukeapage:
committed to master
https://github.com/jshint/jshint/commit/92acdd133cf744d6eaee05246c61c13aa8b4513c
https://github.com/jshint/jshint/commit/5c33ded31447dc59520c98ecb868e3258746e6f2
https://github.com/jshint/jshint/commit/2f98f9168cc48277b38a61c1553a37122ba35a67
https://github.com/jshint/jshint/commit/d09a219e59cd130831fd87a7a93bc3a5e39fc28c

jugglinmike:
Thank you, Luke!

PR #2142, opened 2015-02-01
[[FIX]] Handle multi-line template literal return statements

caitp:
Closes #1814

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4151,7 +4151,8 @@ var JSHINT = (function() {
 
 
   stmt("return", function() {
-    if (this.line === state.tokens.next.line) {
+    if (this.line === state.tokens.next.line ||
+        this.line === state.tokens.next.startLine) {
       if (state.tokens.next.id !== ";" && !state.tokens.next.reach) {
         this.first = expression(0);
 
--- a/src/lex.js
+++ b/src/lex.js
@@ -1005,7 +1005,8 @@ Lexer.prototype = {
           return {
             type: tokenType,
             value: value,
-            isUnclosed: true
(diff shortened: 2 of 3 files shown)


caitp:
@jugglinmike could you review this one? it was the quickest hack i could come up with to support multi-line tokens in return statements

jugglinmike:
Hey Cait! It looks like this is indicative of a larger problem that exists for *all* multi-line structures: https://github.com/jshint/jshint/blob/33612f8cedce779d0d22571b72da4216d18bee9c/src/lex.js#L1377-L1386

For instance, this also fails:

```js
/* jshint multistr: true */
function foo() {
  return 'start\
    end';
};
```

I know your patch is intended as a quick hack, but what would you say to trying to resolve the problem a bit more generally? Here's what I'm thinking that would look like:

Multiline structures would need to track their last line *and* first line. I think it would make the most sense to change the current semantics around and have `line` refer to the first line. A new `lastLine` property (and maybe `lastLineChar`?) would be defined for multi-line structures only. (We could also try keeping `line` behavior as-is and defining a new `firstLine` property, but I think that would be more confusing.)

Do you think it would be that easy? Or maybe not and better to try in a follow up patch?

caitp:
@jugglinmike in general, I don't think it is a problem. It is a problem for templates, because a single template token can span multiple lines. But as far as I'm aware, no other token fits that category

jugglinmike:
@caitp It's also a problem for multi-line string literals.

caitp:
few extra bits added to work for string literals too... i think this can be generalized more cleanly, but i'll do that in a different patch

PR #2144, opened 2015-02-03
[[FIX]] Support more cases of ES6 module usage

### LINKED ISSUE #2118: ES6 import default and named export handled incorrectly
Consider this ES6 import:

```js
import _, { each } from 'underscore';
```

This results in the error:

```
Expected 'from' and instead saw ','
```

while this is should be valid.

### LINKED ISSUE #2143: ES6 export from source
jshint errors when encountering an ES6 export from a source:

```js
export { foo } from "source"
```

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports

leebyron:
The ES6 module parsing is incomplete currently. This adds more cases.

Fixes #2118
Fixes #2143


leebyron:
Specced in https://people.mozilla.org/~jorendorff/es6-draft.html#sec-imports, this pull enables the remaining cases of *ImportClause* and from https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports enables *export ExportClause FromClause*

caitp:
I can see that it's specced, `ExportClause FromClause` was not spec'd last time this was touched. I'm trying to find when this changed. Anyways, the exports portion of the patch looks good to me (I'd prefer it if there were comments indicating the specific grammar a given block corresponds to, there are some things like that in the export statement function)

leebyron:
Sure thing, I'll add some comments pointing to spec

caitp:
huh, interestingly it seems to have been spec'd before, I don't remember that. anyways, looks ok

leebyron:
Just added some inline comments to help illustrate where portions of the parse rule are fulfilled for both import and export

src/jshint.js
```diff
@@ -4242,7 +4242,18 @@ var JSHINT = (function() {
       // ImportClause :: ImportedDefaultBinding
       this.name = identifier();
       addlabel(this.name, { type: "unused", token: state.tokens.curr });
-    } else if (state.tokens.next.id === "*") {
+      if (state.tokens.next.value === ",") {
+        // ImportClause :: ImportedDefaultBinding , NameSpaceImport
```
caitp: @rwalrdron can you explain the ImportDefaultBinding grammar? It doesn't make a whole lot of sense for that to be followed by comma NameSpaceImport (or comma NamedImports) --- and there's no example of this in section 15.2.1.15, which could just be an oversight but the grammar just looks wrong

leebyron: Sure, `ImportedDefaultBinding` allows writing a shorthand for the special `"default"` export name.

So say I have a module `"foo"`:

```js
export var x = "x";
export default "default";
```

In another file, I could *import ImportedDefaultBinding , NameSpaceImport*

```js
import fooDefault, * as fooAll from "foo"
console.log(fooDefault); // "default"
console.log(fooAll); // { default: "default", x: "x" }
```

Or I could *import ImportedDefaultBinding , NamedImports*

```js
import fooDefault, { x as fooX } from "foo"
console.log(fooDefault); // "default"
console.log(fooX); // "x"
```

leebyron: Another way to say this is that these are equivalent:

```js
import fooDefault, * as fooAll from "foo"
// vs
import * as fooAll from "foo"
const fooDefault = fooAll.default;
```

As are these:

```js
import fooDefault, { x as fooX } from "foo"
// vs
import { default as fooDefault, x as fooX } from "foo"
```

The grammar just provides a shorthand for these cases.

rwaldron: This is an oddball, but totally legit case. Coincidentally, @arv and I were talking about this grammar about two weeks ago (I had noticed that Traceur doesn't yet support it).

rwaldron:
LGTM

src/jshint.js
arv: Yeah, I don't like this grammar but I got voted down. Traceur should be fixed.

caitp:
LGTM (but I think someone should edit the word document to add some clarifying notes for the grammar points marked above, because it did not make a lot of sense to me)

leebyron:
Any other action you need from me?

rwaldron:
@leebyron can you add, somewhere near line 4246 and below this: `// ImportClause :: ImportedDefaultBinding , NameSpaceImport` a nice big block of comments that includes your excellent explanations [here](https://github.com/jshint/jshint/pull/2144#discussion_r23978460)? Once that's done, ping me here and I will land it. Thanks!!

leebyron:
How's that, @rwaldron?

src/jshint.js
```diff
@@ -4337,31 +4357,41 @@ var JSHINT = (function() {
           break;
         }
       }
+      if (state.tokens.next.value === "from") {
```
leebyron: In a follow-up diff, the linting behavior of *export* should be implemented more thoroughly.

The body of the *ExportClause* needs to be linted differently depending on this piece of information.

With "undef" and "unused" turned on:

Should pass:
```js
var foo = "foo";
export { foo };
```

Should fail "undef" (currently does not)
```js
export { foo };
```

Should pass (currently does, but because above is broken)
```js
export { foo } from "source";
```

Should fail "unused" (currently does not)
```js
var foo = "foo";
export { foo } from "source";
```

Issue #2148, opened 2015-02-04
Support ES6 export aliasing


leebyron:
This syntax is currently a parse error but is syntactically valid ES6:

```js
export { foo as bar };
export { biz as baz } from "source";
```

Issue #2147, opened 2015-02-04
ES6 Modules should impact undef and unused rules


leebyron:
See discussion: https://github.com/jshint/jshint/pull/2144#discussion_r24038445

PR #2149, opened 2015-02-04
[[FIX]] ES6 modules respect undef and unused

leebyron:
This adds scoping rules for ES6 modules, ensuring that undef and unused rules accurate reflect
the behavior of ES6 modules on the scope.

This also adds syntax parsing support for `export { foo as bar }` aliasing.

Fixes: #2147
Fixes: #2148

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1394,9 +1394,8 @@ var JSHINT = (function() {
   // fnparam means that this identifier is being defined as a function
   // argument (see identifier())
   // prop means that this identifier is that of an object property
-  // exported means that the identifier is part of a valid ES6 `export` declaration
 
-  function optionalidentifier(fnparam, prop, preserve, exported) {
+  function optionalidentifier(fnparam, prop, preserve) {
     if (!state.tokens.next.identifier) {
       return;
     }
@@ -1407,10 +1406,6 @@ var JSHINT = (function() {
 
     var curr = state.tokens.curr;
     var val  = state.tokens.curr.value;
-
-    if (exported) {
-      state.tokens.curr.exported = true;
-    }
 
     if (!isReserved(curr)) {
       return val;
@@ -1433,9 +1428,8 @@ var JSHINT = (function() {
   // fnparam means that this identifier is being defined as a function
   // argument
   // prop means that this identifier is that of an object property
-  // `exported` means that the identifier token should be exported.
-  function identifier(fnparam, prop, exported) {
-    var i = optionalidentifier(fnparam, prop, false, exported);
+  function identifier(fnparam, prop) {
+    var i = optionalidentifier(fnparam, prop, false);
     if (i) {
       return i;
     }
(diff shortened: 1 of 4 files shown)


leebyron:
The tables in https://people.mozilla.org/~jorendorff/es6-draft.html#sec-static-and-runtime-semantics-module-records do a nice job of illustrating local scope effects of es6 modules.

rwaldron:
@leebyron you're on fire :)

I need some time with this, I might be paranoid but I want to understand how this will affect expectations pertaining to the extant to this rule: 

```
exported

A directive for telling JSHint about global variables that are defined in the current file but used elsewhere. Use it together with the unused option.

/* exported EXPORTED_LIB */
```

Which appears (cursory look) to be changed by this patch (tell me if I'm misunderstood).

leebyron:
@rwaldron that rule's behavior should be unchanged. This diff only affects ES6 exports.

leebyron:
(I know this because I accidentally broke it while writing this patch a first time, and then carefully made sure I didn't alter it in this 2nd attempt!)

rwaldron:
Ah, that's great—thanks for the follow up. 

LGTM, pending @caitp and/or @jugglinmike

caitp:
I don't think this is very good, we should not be warning on unused module exports

rwaldron:
> I don't think this is very good, we should not be warning on unused module exports (since they're likely to be used when they're imported)

/me busted.

I haven't actually had a chance to test this locally. That LGTM was lazy.

caitp:
i might be wrong, it was from a quick glance, but that's how it looked to me

leebyron:
> I don't think this is very good, we should not be warning on unused module exports

Can you explain @caitp? Do you mean undefined module exports? What test case do I have that's incorrect or missing?

leebyron:
@caitp from your first comment, it sounds like you misunderstood this as adding warning when an exported variable was not being used. That's not what this is adding. You're definitely correct that the act of exporting a variable indicates usage, since it becomes available for another importer.

What this diff *does* do is look for exporting undefined variables, which are often the result of typos or bad refactors. You almost never want to export a variable you haven't defined in the module scope (and defining via `/* globals */` rule handles the almost).

rwaldron:
I think the diff shown for the changes to `testES6BlockExports` was misleading, but the full source makes sense.

leebyron:
@rwaldron I'm happy to add clarifying comments for those tests. Which test did you find misleading?

caitp:
I understood it --- I think this is pretty abysmally bad stuff to be spec'd though. It's good that folks are implementing modules proper now, but it's really not good if crap like this can't be changed :(

leebyron:
Can you be more clear, @caitp? I probably agree with you but just am not clear about what you're upset
about.

caitp:
ExportClauses not producing bindings in module scope is extremely dumb, all kinds of awful

leebyron:
You mean the through bindings? `export x from "x"`?

I agree that those are pretty confusing.

caitp:
anyways, it looks okay

leebyron:
I'm adding some comments to the tests, since this syntax is new and easy to misread.

leebyron:
Any other followups necessary here?

rwaldron:
@caitp 

> ExportClauses not producing bindings in module scope is extremely dumb, all kinds of awful (particularly when there is a FromClause)

How would the following know _what_ kind of binding to produce?

```js
export { foo } from "source";
```

Which should be created in this scope? `let`, `const` or `var`?

rwaldron:
> Any other followups necessary here?

No, it's good

leebyron:
Just to make sure it's clear (I added comments in the tests to this effect), the example you just wrote:

```js
export { foo } from "source";
```

Transpiles to:

```js
exports.foo = require("source").foo;
```

Which might help explain why a local binding isn't made for particular syntax.

Of course, in an actual ES6 capable browser (none yet), swap `exports` and `require()` for their native internal corollaries.

rwaldron:
@leebyron, sorry I wasn't asking in terms of transpilation, but purely in terms of scope contour design. But the resulting code makes sense—there is no `foo` in the module scope, it's created on the `exports` intermediary.

Issue #2151, opened 2015-02-04
Support nested ES6 template literals


leebyron:
ES6 template literals should be able to be nested:

```js
var nested = `Look and ${ `Nested ${ `whoaaa` } template` } listen`;
```

rwaldron:
@caitp should review this

Issue #2082, opened 2015-01-06
Object Literals in String Templates show E020


despairblue:
```javascript
console.log(`TypeError: expected "${ util.inspect(object, { depth: 0 }) }" to be of type "${typeSchema}"`);
```
shows `Expected '}' to match '{' from line 60 and instead saw '}) }" to be of type "${'. (E020)` although it is correct syntax. `esnext` is enabled.

rwaldron:
Thanks for the report!

PR #2152, opened 2015-02-04
[[FIX]] Parse nested templates

leebyron:
ES6 template literals are nestable, jshint's parser is capable, but it's lexer isn't. This patch adds the concept of a stack of contexts to the lexer, replacing the previous boolean flag. This context stack pattern is borrowed from the awesome Acorn parser project, where it's used to enable more than just template literals.

Fixes: #2151
Fixes: #2082
Fixes: #1814

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -35,7 +35,8 @@ var Token = {
 };
 
 var Context = {
-  Template: 1
+  Block: 1,
+  Template: 2
 };
 
 // Object that handles postponed lexing verifications that checks the parsed
@@ -224,13 +225,30 @@ Lexer.prototype = {
     case ")":
     case ";":
     case ",":
-    case "{":
-    case "}":
     case "[":
     case "]":
     case ":":
     case "~":
     case "?":
+      return {
+        type: Token.Punctuator,
+        value: ch1
+      };
+
+    // A block/object opener
+    case "{":
+      this.context.push(Context.Block);
+      return {
+        type: Token.Punctuator,
+        value: ch1
+      };
+
+    // A block/object closer
+    case "}":
+      if (!this.inContext(Context.Block)) {
+        return null;
+      }
+      this.context.pop();
       return {
         type: Token.Punctuator,
         value: ch1
--- a/src/lex.js
+++ b/src/lex.js
@@ -34,6 +34,10 @@ var Token = {
   TemplateTail: 12
 };
 
+var Context = {
+  Template: 1
+};
+
 // Object that handles postponed lexing verifications that checks the parsed
 // environment state.
 
@@ -111,7 +115,7 @@ function Lexer(source) {
   this.from = 1;
   this.input = "";
   this.inComment = false;
-  this.inTemplate = false;
+  this.context = [];
   this.templateLine = null;
   this.templateChar = null;
 
@@ -122,6 +126,10 @@ function Lexer(source) {
 
 Lexer.prototype = {
   _lines: [],
+
+  inContext: function(ctxType) {
+    return this.context.length > 0 && this.context[this.context.length - 1] === ctxType;
+  },
 
   getLines: function() {
     this._lines = state.lines;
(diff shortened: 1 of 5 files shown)


leebyron:
BTW, expanding the use of context can help solve #2082. Acorn uses the same approach to solve this problem.

leebyron:
Just also fixed #2082 in a 2nd commit in this patch, editing the description of the pull req now.

src/lex.js
```diff
@@ -1026,19 +1053,17 @@ Lexer.prototype = {
...
         value += ch;
         this.skip(1);
       }
     }
 
-    // Final value is either TokenTail or NoSubstititionTemplate --- essentially a string
-    tokenType = this.inTemplate ? Token.TemplateTail : Token.StringLiteral;
-    this.inTemplate = false;
+    // Final value is either StringLiteral or TemplateTail
+    tokenType = tokenType === Token.TemplateHead ? Token.StringLiteral : Token.TemplateTail;
```
caitp: since you're fixing all this, you should also fix this: don't use Token.StringLiteral for no-substitution templates, otherwise they could be treated as valid directives, which is not true.

caitp: test case:

```js
function fn() {
  `use strict`;
  return "\077";
} // should not yield any warnings
```

leebyron: Great point. I'll add that to this PR

caitp:
rest of it lgtm, the other thing could be a followup bug

leebyron:
If you're comfortable merging before I get to fixing the directive issue, go nuts. I'll still follow up with that.

caitp:
I think I will wait, I need to make sure the startLine thing can still work correctly with the stack structure, @jugglinmike or @rwaldron might also want to have a look

leebyron:
Just added your suggestion, @caitp. Let me know if that's in line with what you were looking for.

src/lex.js
```diff
@@ -31,7 +31,13 @@ var Token = {
   RegExp: 9,
   TemplateHead: 10,
   TemplateMiddle: 11,
-  TemplateTail: 12
+  TemplateTail: 12,
+  TemplateString: 13
```
caitp: nit: `NoSubstTemplate` maybe?

tests/unit/core.js
```diff
@@ -818,6 +818,21 @@ exports.testES6TemplateLiteralMultiline = function (test) {
...
+exports.testES6TemplateLiteralsAreNotDirectives = function (test) {
+  var src = [
+    "function fn() {",
+    "`use strict`;",
+    "return \"\\077\";",
+    "}"
+  ];
+
+  TestRun(test)
+    .addError(2, "Expected an assignment or function call and instead saw an expression.")
```
caitp: I don't think this is right, there should be no error for this

caitp: if it is going to be an error, the message should definitely be clearer

caitp: Another case we probably want to check is this:

```js
function fn() {
  `${"use strict"}`;
  return "something-strict-mode-only";
}
```

That will pass, but it's a nice test to have anyways

leebyron:
Changed to `NoSubstTemplate` and added that second test.

I think the error is expected, you get the same error for this example:

```js
function fn() {
  42;
}
```

So it's not specific to templates.

leebyron:
@caitp - I just added another commit to this patch that covers what you were doing in #2142 but maintaining awareness of the context, so you won't need to rebase over this PR.

I took your test cases from that PR verbatim, but added cases for `yield` expressions as well.

leebyron:
Anything else you guys would like me to cover here @caitp @rwaldron?

caitp:
good enough, can you try to figure out why coverage decreased? might need some extra test cases

caitp:
I'm guessing it's probably for the `yield` multi-line expressions

caitp:
hmm nope, I'm wrong about that :<

caitp:
oh, you can remove the quotmark handling for backticks, since templates aren't string tokens anymore

caitp:
missing a case for [this](https://coveralls.io/builds/1872194/source?filename=src%2Flex.js#L248) as well

leebyron:
Good catch, I'll do that. 

From looking at coverall's report, it's hard to tell what didn't get tested. It thinks coverage of style.js went down, but I didn't touch that file. Lex.js also went down but I can dig into that.

caitp:
Stuff in the last commit is okay, but I was referring to [this(]https://github.com/jshint/jshint/blob/8e680e6accfac7c52933b4bc20a8b25c8382171a/src/style.js#L66-L70) << block is never hit because templates are not strings

leebyron:
I just saw that too, and updated that commit :)

caitp:
coverage still seems to have gone down a tiny bit but I think everything relevant to this PR is covered... will merge

leebyron:
Thanks @caitp!

PR #2156, opened 2015-02-06
[[FIX]] remove unused var

leebyron:
Cleanup lint error from 7e80490 which removed a check using the `esnext` var.

cc @caitp


caitp:
looks good to me

PR #2160, opened 2015-02-06
[[TEST]] Improve lexer test coverage

leebyron:
This adds two test cases that were uncovered: numbers that appear to be the beginning of identifiers, and multi-line comments between ignore directives.


leebyron:
It looks like appveyor just timed out for unknown reasons. Doesn't seem related to this at all.

caitp:
i restarted it, the no build log looked flaky

PR #2166, opened 2015-02-07
Detect unused variables that are only ever assigned to, but never referenced

### LINKED ISSUE #1994: Unused detection misses out: Assigning a variable is not using it.
```
var x;
x = require( 'x' );
```

x is not used, albeit 'unused' is set to true in jshint's json file. jshint does not catch it.

Assigning a variable without ever reading it is not using it.

I like to catch these things, since I want to clean my code base of node modules I historically requried but do not seem to require any more.

Please don't tell me to write it all as `var x = require( 'x' );` yes it is catched then. But in many cases these requirements are inside conditionals, like same code used for node and for browser where in browser its a global but node a require assignment I put an if (SERVER) around it. 

Suggested solution: For undef checking, if a variable is simply assigned but never used as right-hand-side, it is not to be considered to be "used".

Kind regards,

univerio:
Variables first declared then assigned to are now detected:

    var x;
    x = 1;

A further, more complicated improvement not covered by this patch would be to keep track of the usage of each assigned value of every variable, to detect cases like:

    var x = 1;
    x = x + 1;

Fixes #1994

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1889,8 +1889,10 @@ var JSHINT = (function() {
         // the name is in a block scope.
         switch (block ? block[v]["(type)"] : funct[v]) {
         case "unused":
-          if (block) block[v]["(type)"] = "var";
-          else funct[v] = "var";
+          if (state.tokens.next.id !== "=") {
+            if (block) block[v]["(type)"] = "var";
+            else funct[v] = "var";
+          }
           break;
         case "unction":
           if (block) block[v]["(type)"] = "function";
--- a/tests/unit/fixtures/gh870.js
+++ b/tests/unit/fixtures/gh870.js
@@ -1,4 +1,4 @@
-var s;
+var s; // jshint ignore:line
 var iterator;
 
 while (s = iterator.next()) { // jshint ignore:line
--- a/tests/regression/thirdparty.js
+++ b/tests/regression/thirdparty.js
@@ -30,6 +30,7 @@ exports.jQuery_1_7 = function (test) {
     .addError(1607, "'table' is defined but never used.")
     .addError(1710, "'internalKey' is defined but never used.")
     .addError(1813, "'internalKey' is defined but never used.")
+    .addError(2174, "'object' is defined but never used.")
     .addError(2818, "Expected an assignment or function call and instead saw an expression.")
     .addError(2822, "Expected an assignment or function call and instead saw an expression.")
     .addError(2859, "'rnamespaces' is defined but never used.")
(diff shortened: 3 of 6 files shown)


lukeapage:
if only it was this simple!

here is a false positive unused:
```
var a = 2;
var b = (a = 1);
```
and of course there are plenty more uses of identifiers that can be writing and not reading (or both).

I think its a large amount of work.

univerio:
@lukeapage

I disagree that that's a false positive under the simplified semantics I have proposed. `a` is simply never used in your example. Yes, it's assigned to, and yes, it happens that the value of the assignment is used (assigned to `b`), but the value of `a` at any point is not ever used. As soon as you actually use the value, the warning disappears:

```
var a = 2;
var b = (a += 1);
```

lukeapage:
Your right,  sorry, should have thought a bit more..

lukeapage:
@jugglinmike I like the idea behind this and its a very popular request. WDYT ?

Do you think its too big a change to include in jshint.. one for jshint 3?

tests/unit/fixtures/gh870.js
```diff
@@ -1,4 +1,4 @@
-var s;
+var s; // jshint ignore:line
```
lukeapage: I think it might be better to not ignore lines and add the errors to the expected.. or change the test to actually use `s`.

src/jshint.js
```diff
@@ -1889,8 +1889,10 @@ var JSHINT = (function() {
         // the name is in a block scope.
         switch (block ? block[v]["(type)"] : funct[v]) {
         case "unused":
-          if (block) block[v]["(type)"] = "var";
-          else funct[v] = "var";
+          if (state.tokens.next.id !== "=") {
+            if (block) block[v]["(type)"] = "var";
```
lukeapage: although possibly not universal, most of the code wraps blocks in `{`

jugglinmike: Yes, that's our strong preference. @univerio would you mind adding in the braces?

jugglinmike:
I'm sure there are plenty of people who would find this new check helpful, but there are probably just as many who would be frustrated by sudden failures in continuous integration builds over a detail that is technically harmless. This is a great addition, but we should wait until JSHint 3.0.

We've been talking about that for a while now, so I think we should start getting serious about time frames. We'll keep you posted, @univerio!

adzellman:
@jugglinmike Are you considering taking this change for 3.0.0? When do you expect 3.0.0 to be released?
As an alternative, what about your Option 3 from #1994? That would be attainable without breaking existing functionality.

dvhx:
function main() {
      var a;
      a = 4;
      return 'Hello, World!';
    }
    main();

This also doesn't produce warning about unused "a"

Issue #2079, opened 2015-01-05
Figure out the next step toward plugin system.


sebv:
See the ugly (but working) #2077 PR to get an idea what is needed. @rwaldron was mentioning the need for a common design effort.

In short we need to expose a lot of internals, plugins would need to do stuff like registering prefix, statement call error/warnings, so the question is what is the best way to do that?

sebv:
On top of my head I see 2 options:

1/ expose an `plugin` hook similar to `api` within the `itself` object.
2/ expose method directly on the JSHINT object.

caitp:
Hooks you need:

1. scanner hooks
2. Hooks for each production

Other things you need:

1. Linting rules rewritten as items which match patterns of productions (if new productions are added, they don't break the existing lobbying rules)

At least, in my head, that's the most elegant and maybe workable solution, using the fewest hacks.

sebv:
What is  a `production`?

caitp:
A symbol representing a sequence of symbols scanned from the source code, part of the formal grammar of the language.

To add new parsing rules, you'd need to add rules for new productions. That would be easily pluggable. To add new linting rules, you'd add new rules for generating errors from grammar productions (or more complex analysis). I'm not sure there is a better way to make linting pluggable.

For simple rules, you could even define a domain specific Language for creating the rules for generating errors.

It would be a pretty substantial rewrite of the current architecture, but we've discussed it a bit

sebv:
@caitp sounds good, but how long is this going to take? A couple of things that struck me when I wrote the `async` extension:

- We need prefix("async") and stmt("async"), when both seem to do the same thing (at least for async case). The issue here seem to be that Jshint doesn't look enough token ahead. 
- The whole `state.tokens.curr`/`state.tokens.next`, being overuled by `peek`, sounds like state.tokens functionality should be extended.
- The whole `fat arrow` logic seems out of place, It may be possible to put the the function detection in one place (same thing, probably have to look more than one token ahead.).

caitp:
> We need prefix("async") and stmt("async"), when both seem to do the same thing (at least for async case). The issue here seem to be that Jshint doesn't look enough token ahead.

Arguably, this is another reason to move away from the current parsing strategy ;) It's hard for people to understand, and potentially leads to code duplication.

> The whole state.tokens.curr/state.tokens.next, being overuled by peek, sounds like state.tokens functionality should be extended.

Well sure, but ideally you shouldn't be dealing with tokens directly anyways.

[...]

PR #2167, opened 2015-02-09
[[FEAT]] Parse and lint JSX

leebyron:
JSX is a syntactical addition spec'd here: http://facebook.github.io/jsx/. It's frequently used alongside the React view rendering library, but is also used in other projects.

To avoid needing to transform before parsing, this adds the ability to parse and lint JSX laden code directly, provided that the `jsx` option is turned on.

Lint rules for undef follow the semantics of React if you ask it to by passing "react" to the "jsx" option, otherwise no assumptions are made about scoped variable usage.


leebyron:
Assembled this over the past couple days and looking for feedback, especially on some of the implementation details with the parser that I was less familiar with. The lexing and parsing was fairly straight-forward to add so that it's isolated, however a new feature to the lexer was necessary for disambiguating the `<` token. I borrowed the technique from SweetJS, where they use it for a lot of disambiguation - most famously to disambiguate `/` for RegExp.

rwaldron:
JSX is not a standard and will likely never be part of the Ecma-262 standard. While the JSHint project appreciates the other contributions you've made, there is no interest in extending JSHint with any more non-standard features than it already has. The SpiderMonkey specific extensions were a massive mistake and should never have been added to JSHint, but it is what it is and the goal now is preparedness for ES6. 

There is already a mechanism built into JSHint that perfectly tailors to the needs of JSX: [ignore delimiters](https://github.com/jshint/jshint/blob/master/tests/unit/fixtures/ignoreDelimiters.js)

leebyron:
I'm definitely aware JSX has no plans to be proposed as being part of Ecma-262. I tried to make that clear in the description for the option. I was also careful to implement this addition in such a way that it is not threaded throughout the lexer and parser but instead very self-contained. When the `jsx` option is off, the behavior (and performance) of jshint are completely unchanged. I saw the SpiderMonkey extension unraveling and understand why that was a mistake, but since this change only affects one part of the grammar (an additional case on PrimaryExpression), removing it in the future, should ES20XX conflict would be trivial.

Unfortunately, ignore delimiters are a fair escape hatch around arbitrary code, but that also means the baby goes out with the bathwater. The proposed value added by this patch is an actual lint of the JSX, checking for syntax mistakes, undefined identifiers, etc. Since JSX may contain arbitrary JS expressions, it also means linting those JS expressions. It also makes it easy to adopt JSHint on a codebase which already uses JSX, otherwise you must first go annotate all JSX regions to be ignored, which could be a burden for larger or more distributed codebases.

Is there any other way we can preserve the value of JSHint against JavaScript using JSX?

caitp:
@leebyron the idea that I had in mind for this sort of thing, was to switch to AST-based linting, and allow the parser to be extended using BNF as an API, so that it would be easy for people to add support for linting non-standard stuff.

But it's challenging to implement without a serious understanding of YACC and similar tools, and challenging to implement with good performance, and challenging to work around performance issues in a way that would work in a browser. So in other words, it's a very difficult thing to work on, but it's probably the best bet for supporting this stuff

leebyron:
@caitp that sounds awesome. JSHint with a plugin system would be a really great future for the project and open up the ability to write lint rules for arbitrary SweetJS macros as well.

leebyron:
The performance thing I'm really interested in preserving. ESLint recently added support for JSX, which is great, but their performance is quite dramatically worse than JSHint - especially noticeable when applied to large codebases. (ESLint is also quite further behind on supporting and linting ES6 at the moment).

leebyron:
Just following up here, is there any way we can preserve the value of JSHint against JavaScript using JSX? Is JSHint strictly supporting Ecma-262 approved features only as a project philosophy, or are these kinds of extensions still valuable to the project?

rwaldron:
> removing it in the future, should ES20XX conflict would be trivial.

That's exactly the problem. Once it becomes part of JSHint, it will likely be impossible to remove. Compared to something like the style related rules, which were easy because there was a tool to replace them and removal didn't break the ability lint source code—the old rules were simply no longer there. If JSX support lands today, then in 2018 there is some new syntax that conflicts, all the code that was written between today and that date, which uses JSX, will stop successfully linting.

> Is JSHint strictly supporting Ecma-262 approved features only as a project philosophy

It doesn't make sense to impose the burden of a special-case, non-standard syntax extension on the maintainers and users of JSHint—surely you agree with this on a pragmatic level. I haven't closed this yet because I don't know what to do. If JSHint allows this, then there is no more defense against everyone else's pet syntax. JSHint has already said no to "async/await" and that syntax extension has a better chance of language adoption than JSX. If JSX support is accepted, then JSHint has to accept Angular `@` annotations—even though it conflicts with an existing proposal ([which was already accepted](https://github.com/rwaldron/tc39-notes/blob/master/es6/2012-09/sept-19.md#syntactic-support-for-private-names), then deferred to ES7). That's only one example, but I think it illustrates the unfair position JSHint would be put in if it accepted JSX support. The way forward is a mechanism for defining plug-ins that allow these kinds of extensions.

leebyron:
> It doesn't make sense to impose the burden of a special-case, non-standard syntax extension on the maintainers and users of JSHint—surely you agree with this on a pragmatic level.

I definitely resonate with this.

> If JSHint allows this, then there is no more defense against everyone else's pet syntax.

I understand and appreciate the sentiment here, but also think that's a bit of a [logical fallacy](http://en.wikipedia.org/wiki/Slippery_slope).

> If JSX support is accepted, then JSHint has to accept Angular @ annotations—even though it conflicts with an existing proposal.

Not necessarily. That's up to you and the other maintainers (and obviously the burden is on someone from the Angular community to propose and actually draft a good PR). Since Angular annotations actively conflict with an existing proposal, then it seems fair to avoid.

> If JSX support lands today, then in 2018 there is some new syntax that conflicts, all the code that was written between today and that date, which uses JSX, will stop successfully linting.

Absolutely. The day that happens, I think JSHint support will be the least of JSX's problems. I can also imagine a fairly clear incremental path for removing JSX in that case. First, when a `es2018` option is added to JSHint, make a rule that the `jsx` option cannot be turned on at the same time. Then when the possible jsx-breaking `es2018` becomes the standard, then a major v. bump of JSHint could remove JSX support completely. It's also an important reason for why the `jsx` option should never be on by default.

> The way forward is a mechanism for defining plug-ins that allow these kinds of extensions.

I'm pretty psyched about that future. In lieu of having that plug-in system today, the question remains of how we might find address the value of linting this additional syntax today. I think it might be worthwhile to have something like this as an option in JSHint in the intermediate time, moving it into a plugin when the plugin based version of JSHint ships. I'm also happy to help build and host such a plugin. Hopefully this future comes before any possible future JSX-conflicting Ecma-262 proposals, so that there's a much clearer path forward for support in that hazy scenario.

Obviously I would be happy to get this into JSHint, so I'm biased, but I'm also trying to be pragmatic about this, because I think it's really easy to get stuck trying to predict every possible failure case so many years into the future. Software is always changing, and I'm sure JSHint will change in the future as well. I also very much understand the concern about adding options for supporting non-spec syntax.

From that point of view I look towards burden on maintenance - which is why I've tried to keep this change as self-contained as possible, JSX takes a very small syntactical surface area. I'm also happy to be on the hook to issue a fixing PR should anyone find bugs with the parser.

Then weighing that against value added, using precedent of other tools. While the [JSX Spec](https://github.com/facebook/jsx) explicitly is not a proposal to add to Ecma-262, it exists in the hopes that tools (like JSHint) will be able to incorporate it as an option. JetBrains has added it to their IDEs, VisualStudio is working on it (MS's annual release schedule makes this slow to launch), ESLint has added it as an option, 6to5 transpiler has JSX support (by default, oddly). So while not being an official part of Ecma-262, there's still a lot of precedent for value in it being incorporated as an option in tools like this one.

jbe456:
jshint should be extensible so that anyone can extend the rules or change the parser (like what they did for jscs : https://github.com/jscs-dev/node-jscs/issues/708).

Anyway, for those interested, you can try ESLint: http://eslint.org/ 
It is fully extensible and support ES6 and JSX : https://github.com/eslint/eslint/issues/1291

rwaldron:
> jshint should be extensible so that anyone can extend the rules or change the parser 

Apparently you've never looked at the source code, nor have you read any of the extensive discussions regarding this subject. 

Please feel free to contribute meaningfully.

jbe456:
Sorry, did not want to offend you! From what I know it looks like there is no easy extension/plugin API accessible: https://github.com/jshint/jshint/issues/2079 or https://github.com/marijnh/acorn/issues/168.

But I agree this can be a bit confusing because "extensible" can mean a lot of different things: existing rules? new rules? parsing engine?

Looking at the well-written documentation I can't find anything on that subject (except the possibility to define its own reporter). Furthermore, you seem to encourage using jshint ignore flags instead of building extensions : cf http://jshint.com/docs/ ignore block

> All code in between ignore:start and ignore:end won't be passed to JSHint so you can use any language extension such as Facebook React. 

Some research on google leads you to this topic, where it is not obvious that rules can be extended: http://stackoverflow.com/questions/21821528/jshint-create-custom-warnings-rules

Also from ESLint github main page https://github.com/eslint/eslint :
>The fact is that I've had a dire need for a JavaScript tool with pluggable linting rules. I had hoped JSHint would be able to do this, however after chatting with Anton, I found that the planned plugin infrastructure wasn't going to suit my purpose.

But I would be happy that you point me to docs on how to proceed. My goal is not to blame anyone, just to solve my issues. Thanks

PR #2169, opened 2015-02-11
[[FIX]] Use globals for global objects definitions

takenspc:
Use globals package for global objects of environments and libraries
except node, browserify, dojo and mootools. Global objects that are
related to jshint rules (ecma builtin, devel and typed) remain in jshint.


takenspc:
I hope this change will reduce duplicated works between linters ("Adding XXX to YYY envs").

Once jshint is in line with what globals defines for node, jshint can use the globals' definitions. 
https://github.com/jshint/jshint/pull/2163

tests/unit/options.js
```diff
@@ -1443,9 +1443,16 @@ exports.browser = function (test) {
...
+    .addError(31, "'alert' is not defined.")
     .test(src, {es3: true, undef: true });
 
-  TestRun(test).test(src, {es3: true, browser: true, undef: true });
+  TestRun(test)
+    .addError(30, "'console' is not defined.")
+    .addError(31, "'alert' is not defined.")
+    .test(src, {es3: true, browser: true, undef: true });
+
+  TestRun(test).test(src, {es3: true, browser: true, devel: true, undef: true });
```
rwaldron: This is a breaking change.

rwaldron:
Requiring a new flag `devel` for `console`, `alert`, etc. will cause **a lot** of currently passing lint tasks to fail.

tests/unit/options.js
takenspc: The intention of the above changes is not breaking but ensuring that jshint keeps current behavior. Please let me explain.

Currently, jshint reports errors for following

```
/* jshint browser:true, undef:true */
console.log("hello");
alert("hola");
```

but it doesn't report errors

```
/* jshint browser:true, undef:true, devel:true */
console.log("hello");
alert("hola");
```
as explained in http://jshint.com/docs/options/#devel

I would like to keep this behavior (or feature). Therefore I added  lines to test.

First I added console and alert to the test fixture.
https://github.com/jshint/jshint/pull/2169/files#diff-1d39308f00a655c961bd6636ed22e255R28

```
// check for devel
console.log("test");
alert("test");
```
Then I added code that ensures that jshint still reports errors for console, alert without devel option.
```
   .addError(30, "'console' is not defined.")
   .addError(31, "'alert' is not defined.")
     .test(src, {es3: true, undef: true });

 TestRun(test)
   .addError(30, "'console' is not defined.")
   .addError(31, "'alert' is not defined.")
   .test(src, {es3: true, browser: true, undef: true });
 ```

Finally I added codes that ensures that devel option still works correctly.

```
TestRun(test).test(src, {es3: true, browser: true, devel: true, undef: true });
```

If it is better that I create a separate test for `devel` option, please let me know.

**Comment by takenspc** [comment]
I added new test that ensures whether the `devel` option still work.

**Comment by takenspc** [comment]
I updated the pull request again not to decrease the coverage. I would like you to review the pull request.

File: src/jshint.js [hunk 2]
```diff
@@ -274,23 +277,32 @@ var JSHINT = (function() {
...
+      combine(predefined, globals.yui);
     }
 
     if (state.option.mocha) {
-      combine(predefined, vars.mocha);
+      combine(predefined, globals.mocha);
+    }
+
+    if (state.option.nonstandard) {
+      combine(predefined, globals.nonstandard);
```
lukeapage: I'm a little concerned about the dependency and the fact jshint would no longer be in control.
Case in point: nonstandard has been merged into builtin in the latest version - https://github.com/sindresorhus/globals/pull/36/files

So, yes, it would be nice, but once we switch we have to either slowly put things back if they break or else release a new major of jshint with breaking changes. So you can understand its a bit scary.

This is going to need the opinion of (another more) core team member I think.

Issue #1995, opened 2014-11-21
ES6 globals are activated even though esnext=false


hacker112:
Promise etc. are added to global predefined list even though "esnext" option is not true. Which just now hide the error that Promise that I used was undefined. (and that I should add var Promise = require('bluebird'); )

I think it was this change is the reason for why the Promise etc. are not seemed as undefined: 
https://github.com/jshint/jshint/commit/54c29007f26a719e3c100f4088d92d190dd0192a

coolaj86:
I just ran into this too.

Is there a way to explicitly ignore globals? That's the option I was actually looking for - like `{ "globals": { "Promise": null } }` or something (rather than true is rw and false is ro)

rwaldron:
I'm going to roll this change back.

jugglinmike:
Hey everybody,

JSHint 2.8.1 ([released today](https://github.com/jshint/jshint/releases/tag/2.6.1)) includes [a new option named `futurehostile`](http://jshint.com/docs/options/#futurehostile) which we designed to be a more understandable way to resolve issues like these.

lukeapage:
@jugglinmike can this be closed now?

jugglinmike:
@lukeapage yup! Thanks

PR #2171, opened 2015-02-12
Remove Set from the ECMAScript5 reserved words

AndrewLorente:
Set is an ECMAScript 6 proposal; in 5 it must be required or manually defined. This fixes an erroneous "Redefinition of..." warning.

(Set is already in the `newEcmaIdentifiers` object, so it didn't need to be added)

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -36,7 +36,6 @@ exports.ecmaIdentifiers = {
   RangeError         : false,
   ReferenceError     : false,
   RegExp             : false,
-  Set                : false,
   String             : false,
   SyntaxError        : false,
   TypeError          : false,


caitp:
s/proposal/object (it's not going anywhere).

For polyfills, you can simply do `/* jshint globals: -Set */` to prevent the redefinition warning, I believe.

@rwaldron / @jugglinmike does that seem fair to you (given 54c29007f26a719e3c100f4088d92d190dd0192a?) or does it make more sense to only issue the redefinition warning if esnext is used

jugglinmike:
@caitp Your recommendation is definitely consistent with that commit, so if we want to "stay the course", then I agree with that.

But if we're opening this up for more general discussion, I think that it's a bit inconsistent for JSHint to require explicit opt in for certain parts of ES6 but implicitly enable other parts of the spec. See also Rick's comments on gh-1995

rwaldron:
> For polyfills, you can simply do /* jshint globals: -Set */ to prevent the redefinition warning, I believe.

Right. JSHint doesn't know what environment the source its operating on will run in, or what its capabilities are. 

>  I think that it's a bit inconsistent for JSHint to require explicit opt in for certain parts of ES6 but implicitly enable other parts of the spec. 

It seems wise to warn about redefinition of any built-ins, but take a look at this case: for `Set`, any code would be paving it V8, SpiderMonkey and JSC (not sure about Chakra)—regardless of whether or not ny `esnext` flag is set. Does that make sense?

(This is why I haven't rolled anything back yet)

caitp:
> any code would be paving it V8, SpiderMonkey and JSC (not sure about Chakra)

https://status.modern.ie/set?term=set --- IE11+, apparently

jugglinmike:
@rwaldron Oh, good point! In making that commit, your intention was to prevent
users from writing code that wasn't portable to future versions of the
language. I completely agree.

That said, that motivation honestly never occurred to me. You can even tell
from my previous response: I called the current behavior "inconsistent" because
I assumed that the change was simply assuming users were in an ES6-compliant
environment.

I think the reason I made that assumption is the same reason we have seen
multiple issues filed against this behavior. A warning like "Redefinition of
'Promise'." is inaccurate in ES5 environments and does nothing to explain the
logical motivation you just described.

I think this is simply a documentation problem. I think we should keep the
current behavior, but we should also (1) do a better job helping users
understand *why* JSHint is issuing a warning and (2) force them to think about
what it means to suppress that warning. I've opened gh-2172 to demonstrate what
I have in mind: a new option named `futurehostile`.

rwaldron:
> (1) do a better job helping users understand why JSHint is issuing a warning and (2) force them to think about what it means to suppress that warning. I've opened gh-2172 to demonstrate what I have in mind: a new option named futurehostile.

Great!

jugglinmike:
JSHint 2.8.1 ([released today](https://github.com/jshint/jshint/releases/tag/2.6.1)) includes [a new option named `futurehostile`](http://jshint.com/docs/options/#futurehostile) which we designed to be a more explicit way for users to enable the overwriting of future-reserved identifiers.

AndrewLorente:
Thanks for handling this, y'all! :)

Issue #1747, opened 2014-06-30
redefinition of Promise (W079)


helmus:
I just switched from 2.4.4 to 2.5.1
But now all my requires are messed up because:

`var Promise = require('bluebird')`
Will warn about:
`Redefinition of Promise (W079)`

Is there anyway I can fix this ? I'm using bluebird all over the place.

fluidsonic:
I wonder why the warning is also thrown for block scoping:

```js
const Promise = require('bluebird');  // same for `let`
```

The global variable will be untouched in this case.

If you can use ES6 features (node 0.11+) and #2090 is fixed then renaming the variable should not be necessary anymore.

rwaldron:
> The global variable will be untouched in this case.

True, and that should be filed as a different issue—which you did, thanks!

But...

> If you can use ES6 features (node 0.11+) and #2090 is fixed then renaming the variable should not be necessary anymore.

`let` and `const` in v8 (as seen in node 0.11.x) are a prototype implementation that is not longer correct (as in, several years out of date) and therefore _not_ an ES6 feature, so you _really_ shouldn't be using them, because they don't adhere to the specification semantics.

caitp:
> so you really shouldn't be using them, because they don't adhere to the specification semantics.

In fact, please, please do not use these (in either v8 or in web applications) --- otherwise we might never be able to ship "correct" behaviour ;_; and that would be sad. If you see legacy let/const behaviour in the wild, please tell the appropriate people to remove it

rwaldron:
@caitp +9001

fluidsonic:
Can anyone reference info on what changed about `const` and `let`?
I wasn't able to find any info on that and I see no reason for not using it in Node.

rwaldron:
@fluidsonic it looks like the version of v8 running in node 0.10.35 has [TDZ semantics](https://gist.github.com/rwaldron/f0807a758aa03bcdd58a), but let and const are still intentionally (and incorrectly) limited to strict mode code.

jugglinmike:
Hey everybody,

JSHint 2.8.1 ([released today](https://github.com/jshint/jshint/releases/tag/2.6.1)) includes [a new option named `futurehostile`](http://jshint.com/docs/options/#futurehostile) which we designed to be a more understandable alternative to `predef: ['-Promise']`. Hope that helps!

JackCA:
^ *2.6.1*

aexmachina:
[...]

PR #2172, opened 2015-02-13
[[FEAT]] Implement new option `futurehostile`

jugglinmike:
A recent change to JSHint included the ECMAScript 6 global identifiers
in JSHint's collection of "predefined" variables. This disallowed the
definition of those global values even in contexts that did not
implement them.

The intent of this change was to help users avoid bugs when migrating to
the new version of the language. Re-using an existing warning mechanism
caused confusion among many users [1][2][3] for two reasons:

1. The generated warning messages were somewhat misleading (i.e. the
   warning "Redefinition of 'Promise'." is inaccurate in ES5
   environments)
2. The method of disabling the warnings did not accurately communicate
   the intent of the developer (i.e. setting `predef: -Promise` does not
   prompt the user to consider the user to consider if creating their
   own global variable named `Promise` is appropriate)

A dedicated warning message more clearly describes the best practice
that motivates this particular constraint on global variable names. An
explicit option makes it less likely that users will disable this
warning without understanding the motivation.

Intoduce a new option, `futurehostile`, and dedicated warning message to
better communicate the intent of restricting usage of future-reserved
identifiers.

[1] GH-1747 redefinition of Promise (W079)
[2] GH-1995 ES6 globals are activated even though esnext=false
[3] GH-2171 Remove Set from the ECMAScript5 reserved words

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -10,51 +10,47 @@ exports.reservedVars = {
 };
 
 exports.ecmaIdentifiers = {
-  Array              : false,
-  Boolean            : false,
-  Date               : false,
-  decodeURI          : false,
-  decodeURIComponent : false,
-  encodeURI          : false,
-  encodeURIComponent : false,
-  Error              : false,
-  "eval"             : false,
-  EvalError          : false,
-  Function           : false,
-  hasOwnProperty     : false,
-  isFinite           : false,
-  isNaN              : false,
-  JSON               : false,
-  Map                : false,
-  Math               : false,
-  Number             : false,
-  Object             : false,
-  Proxy              : false,
-  Promise            : false,
-  parseInt           : false,
-  parseFloat         : false,
-  RangeError         : false,
-  ReferenceError     : false,
-  RegExp             : false,
-  Set                : false,
-  String             : false,
-  SyntaxError        : false,
-  TypeError          : false,
-  URIError           : false,
-  WeakMap            : false,
-  WeakSet            : false
-};
-
-exports.newEcmaIdentifiers = {
-  Set     : false,
-  Map     : false,
-  WeakMap : false,
-  WeakSet : false,
-  Proxy   : false,
-  Promise : false,
-  Reflect : false,
-  Symbol  : false,
-  System  : false
+  3: {
+    Array              : false,
+    Boolean            : false,
+    Date               : false,
+    decodeURI          : false,
+    decodeURIComponent : false,
+    encodeURI          : false,
+    encodeURIComponent : false,
+    Error              : false,
+    "eval"             : false,
+    EvalError          : false,
+    Function           : false,
+    hasOwnProperty     : false,
+    isFinite           : false,
+    isNaN              : false,
+    Math               : false,
+    Number             : false,
+    Object             : false,
+    parseInt           : false,
+    parseFloat         : false,
+    RangeError         : false,
+    ReferenceError     : false,
+    RegExp             : false,
+    String             : false,
+    SyntaxError        : false,
+    TypeError          : false,
+    URIError           : false
+  },
+  5: {
+    JSON               : false
+  },
+  6: {
+    Map                : false,
+    Promise            : false,
+    Proxy              : false,
+    Reflect            : false,
+    Set                : false,
+    Symbol             : false,
+    WeakMap            : false,
+    WeakSet            : false
+  }
 };
 
 // Global variables commonly provided by a web browser environment.
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -1818,8 +1818,7 @@ exports.esnextPredefs = function (test) {
     '/* global alert: true */',
     'var mySym = Symbol("name");',
     'var myBadSym = new Symbol("name");',
-    'alert(Reflect);',
-    'alert(System);'
+    'alert(Reflect);'
   ];
 
   TestRun(test)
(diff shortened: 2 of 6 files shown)


src/vars.js
```diff
@@ -25,36 +25,32 @@ exports.ecmaIdentifiers = {
...
+    Set     : false,
+    Map     : false,
+    WeakMap : false,
+    WeakSet : false,
+    Proxy   : false,
+    Promise : false,
+    Reflect : false,
+    Symbol  : false,
+    System  : false
+  }
```
rwaldron: I definitely like this

nit: alphabetize lists like this as we go along—thanks!

rwaldron: Should add: 

```js
5: {
  JSON: false
}
```

I don't think we need to go pre-ES3

tests/unit/options.js
```diff
@@ -2102,3 +2102,79 @@ exports.elision = function (test) {
...
+    .addError(2, "'Map' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(3, "'WeakMap' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(4, "'WeakSet' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(5, "'Proxy' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(6, "'Promise' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(7, "'Reflect' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(8, "'Symbol' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(9, "'System' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .test(code, { es5: false });
```
rwaldron: Should test `JSON` and fail on it.

src/vars.js
caitp: does the System object still exist? It was removed from the draft, and I thought it was going to be defined in a separate document but I haven't seen it anywhere if that happened

tests/unit/options.js
```diff
@@ -2102,3 +2102,79 @@ exports.elision = function (test) {
...
+    .addError(2, "'Map' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(3, "'WeakMap' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(4, "'WeakSet' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(5, "'Proxy' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(6, "'Promise' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(7, "'Reflect' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(8, "'Symbol' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .addError(9, "'System' is defined in a future version of JavaScript. Use a different variable name to avoid migration issues.")
+    .test(code, {});
```
rwaldron: Should test `JSON` and fail on it. 

(Sorry, I originally put this comment on the wrong line)

src/vars.js
rwaldron: Indeed, it was removed—I wonder if JSHint should claim domain over the separate spec as well?

jugglinmike:
@rwaldron @caitp Thanks for the review! A couple things:

- Removing `System` is a breaking change. I think this is fair because developers should generally recognize that ES6 is a moving target and that enabling `esnext` will tend to decrease stability
- The most recent commit repurposes the existing `ecmaIdentifiers` attribute. Nobody asked for this, but after moving `JSON` to ES5, it only seemed appropriate to put everything else in ES3 so we'd have a consistent interface for referencing all of these globals.

tests/unit/options.js
jugglinmike: Yes! Awesome! I was really hoping for an excuse to justify the additional indirection that `esVersion[6]` has over, say, `es6Identifiers`

src/vars.js
```diff
@@ -10,51 +10,47 @@ exports.reservedVars = {
...
+    ReferenceError     : false,
+    RegExp             : false,
+    String             : false,
+    SyntaxError        : false,
+    TypeError          : false,
+    URIError           : false
+  },
+  5: {
+    JSON               : false
+  },
```
rwaldron: If you want to really correct, split up 3 into 1, 2 & 3... But that might be a bit much

jugglinmike: I *do* want to be really correct, but I agree that it would be overkill since (currently) there is no way to tell JSHint that you are operating in a pre-ES3 environment.

caitp: Is this convention going to confuse people if the year-based scheme becomes widely known/popular?

caitp: nevermind, I guess the numbers aren't really exposed via options or anything yet

rwaldron:
@jugglinmike this needs a rebase

jugglinmike:
@rwaldron Sure thing; should I squash while I'm at it?

src/vars.js
rwaldron: > Is this (numbered versions) convention going to confuse people if the year-based (eg ES2016™) scheme becomes widely known/popular?

The spec will still have numbered editions, only the title will change. For example: "ECMAScript 2015 Language Specification" is the same document as "ECMA-262, 6th Edition"

rwaldron:
Sure, that will be nice and clean

jugglinmike:
@rwaldron All set

PR #2176, opened 2015-02-15
informative error when jshintrc is invalid

david-gang:
When jshintrc is incorrect, i want to know what happened.
Therefore i am adding the error message to the error and not just saying that the file couldn't be parsed

### DIFF
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -115,8 +115,10 @@ exports.group = {
         ]);
       } catch (err) {
         var msg = out.args[1][0];
+        console.log(msg);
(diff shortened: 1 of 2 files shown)


tests/cli.js
```diff
@@ -115,8 +115,10 @@ exports.group = {
         ]);
       } catch (err) {
         var msg = out.args[1][0];
+        console.log(msg);
```
rwaldron: Is this intentional?

david-gang: No.
I will delete it.

PR #2184, opened 2015-02-19
[[FIX]] Assignment is not bad invocation

dpatti:
There are a multitude of operators that are tested for the left side of
an invocation, as they will evaluate into something that could be a
function. One such missing operator was the assignment operator.

Aside: I know this is something that our CoffeeScript-generated application fails on, and it seems like a reasonable addition.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2430,7 +2430,8 @@ var JSHINT = (function() {
       }
       if (!left.identifier && left.id !== "." && left.id !== "[" &&
           left.id !== "(" && left.id !== "&&" && left.id !== "||" &&
-          left.id !== "?" && !(state.option.esnext && left["(name)"])) {
+          left.id !== "?" && !(state.option.esnext && left["(name)"]) &&
+          left.id !== "=") {
         warning("W067", that);
       }
     }
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -1139,3 +1139,11 @@ exports.testUnCleanedForinifcheckneeded = function (test) {
 
   test.done();
 };
+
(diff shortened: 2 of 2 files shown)


lukeapage:
Hi,

I'm not sure about this.. @jugglinmike closed a similiar bug- #2487.

jshint isn't ideal for linting transpiled code and this particular warning could be disabled.

The question is whether jshint should never warn for this and accept all potentially valid code or warn because its not very clear code.. Personally, in a non-transpiled case, I'd want jshint to warn so that the developer can refactor
```
(foo = function(){})();
```
into the clearer (IMO)
```
foo = function(){};
foo();
```

dpatti:
Did you link the wrong bug there? It looks like it was merged. 

I believe your argument for not allowing it falls more into "style for readability" than "style for safety", and I was under the impression that jshint was trying to move away from the former.

The warning here is simply "illegal invocation", which is not true. My change is among a rather large list of other acceptable invocation targets, some of which could be just as questionable as this.

nicolo-ribaudo:
Probably @lukeapage meant issue https://github.com/jshint/jshint/issues/2485#issuecomment-115420259

Related: http://stackoverflow.com/questions/12358927/how-to-concisely-assign-and-immediately-invoke-a-function-variable

lukeapage:
> Did you link the wrong bug there?

yeah I did sorry.

> The warning here is simply "illegal invocation", which is not true. My change is among a rather large list of other acceptable invocation targets, some of which could be just as questionable as this.

Yes, you may well be right, I am just trying to play devils advocate and I'm unsure, so I need to leave this for someone who's been around here longer...

PR #2050, opened 2014-12-21
nocomma: fix false positives

denis-sokolov:
Fix #2044.

PR #2186, opened 2015-02-22
[[FIX]] Remove tautological condition

jugglinmike:
The parsing logic for the `(` prefix token returns early if the token is
determined to mark the beginning of an arrow function parameter list.
Logic that follows can safely assume that the token is *not* part of an
arrow function parameter lists, and explicit conditions to verify this
are unnecessary.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2480,7 +2480,7 @@ var JSHINT = (function() {
           break;
         }
 
-        if (pn1.value !== "=>" && state.option.nocomma) {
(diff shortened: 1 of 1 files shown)


jugglinmike:
[I mentioned the need for this change while reviewing that patch that introduced the code](https://github.com/jshint/jshint/pull/2050#issuecomment-67786638), but the change was dependent on another patch that was under review, and I forgot to follow up once it landed.

Issue #738, opened 2012-11-09
Eval error when eval method exists in a object


arunoda:
Following line cause an error and It is a legitimate and It has no relationship to eval.

~~~js
result =  schemaEvaluator.eval(experimentId, schema);
~~~

Error is shown here: http://arunoda.me/blog/jshint-evil-eval-check.html

rwaldron:
This _should've_ been corrected for ES5

mbildner:
this pr should fix that https://github.com/jshint/jshint/pull/2188

PR #2188, opened 2015-02-23
[[FIX]] Permit "eval" as object key

mbildner:
Makes it possible to use "eval" as a key in an object. Previously this
would have emitted a "W061" warning. "eval" used as the key for a
presumed global ("window" in "browser:true" and "global" in
"node:true") will still emit the warning.

Fixes https://github.com/jshint/jshint/issues/738

Adds unit tests for permitting using "eval" as object key

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1242,6 +1242,27 @@ var JSHINT = (function() {
     return false;
   }
 
+  function isGlobalEval(left, state, funct) {
+    var isGlobal = false;
+
+    // permit methods to refer to an "eval" key in their own context
+    if (left.type === "this" && funct["(context)"] === null) {
+      isGlobal = true;
+    }
+    // permit use of "eval" members of objects
+    else if (left.type === "(identifier)") {
+      if (state.option.node && left.value === "global") {
+        isGlobal = true;
+      }
+
+      else if (state.option.browser && (left.value === "window" || left.value === "document")) {
+        isGlobal = true;
+      }
+    }
+
+    return isGlobal;
+  }
+
   function findNativePrototype(left) {
     var natives = [
       "Array", "ArrayBuffer", "Boolean", "Collator", "DataView", "Date",
@@ -2396,7 +2417,9 @@ var JSHINT = (function() {
     }
 
     if (!state.option.evil && (m === "eval" || m === "execScript")) {
-      warning("W061");
+      if (isGlobalEval(left, state, funct)) {
+        warning("W061");
+      }
     }
 
     return that;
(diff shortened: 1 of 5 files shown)


src/jshint.js
```diff
@@ -1242,6 +1242,30 @@ var JSHINT = (function() {
     return false;
   }
 
+  function isGlobalEval(left, state, funct) {
+    var isGlobal = false;
+
+    // permit methods to refer to an "eval" key in their own context
+    if (left.type === "this") {
```
jugglinmike: These two `if` statements should be merged and the following `if` changed to an `else if` (we want to avoid creating unnecessary code paths where possible).

src/jshint.js
```diff
@@ -1242,6 +1242,30 @@ var JSHINT = (function() {
...
+      if (_.isNull(funct["(context)"])) {
```
jugglinmike: For simplicity, just use strict equality with `null` here

tests/unit/core.js
```diff
@@ -1139,3 +1139,26 @@ exports.testUnCleanedForinifcheckneeded = function (test) {
...
+  TestRun(test)
+    .addError(21, "eval can be harmful.")
+    .addError(22, "eval can be harmful.")
+    .addError(23, "eval can be harmful.")
+    .addError(25, "eval can be harmful.")
+    .addError(26, "eval can be harmful.")
+    .addError(28, "eval can be harmful.")
+    .addError(29, "eval can be harmful.")
+    .addError(31, "eval can be harmful.")
+    .test(src, { browser: true, node: true });
```
jugglinmike: Enabling both the `browser` and `node` is somewhat rare. It's more important to assert what happens when only one of these options is enabled (especially because the results will be different). We can keep this test, but we also need separate tests for `{ browser: true }` and `{ node: true }`. Be sure to remove the in-line directives from the source code.

mbildner:
@jugglinmike made the changes.

tests/unit/fixtures/gh-738.js
```diff
@@ -0,0 +1,31 @@
...
+	},
+	wrapper: function (str) {
+		// method calling "eval" key from context
+		// permitted use
+		return this.eval(str);
+	}
+};
+
+// object-key use, permitted
+obj["eval"]('console.log(\'hello world\');');
```
jugglinmike: Using double quotes consistently here and below will be more in-line with the project's code style and will allow you to avoid escaping quote marks.

jugglinmike:
@mbildner Thanks! In the future, please make your changes in separate commits--it helps your reviewers keep track of the work as it evolves

mbildner:
@jugglinmike I made the changes in separate commits this time ;)

PR #2192, opened 2015-02-24
[[DOCS]] Introduce a mail mapping file

jugglinmike:
Some contributors have authored commits using different e-mail
addresses. Add a `.mailmap` file to allow git to correlate the identity
of these contributors across e-mail addresses.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,25 @@
+Anton Kovalyov <anton@kovalyov.net> Anton Kovalyov <anton@medium.com>
+Rob Wu <gwnRob@gmail.com> Rob Wu <rob@robwu.nl>
+XhmikosR <xhmikosr@gmail.com> XhmikosR <xhmikosr@users.sourceforge.net>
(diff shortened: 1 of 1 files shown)


caitp:
Does CLAHub make use of this? Or just for legal reasons? Either way lgtm

rwaldron:
> Either way lgtm

same

jugglinmike:
Nah, CLAHub doesn't use it. It's partially for legal reasons, although I think that it's just nice for general book-keeping and attribution. It makes the output of commands like `git shortlog` much more accurate/helpful:

    git shortlog -nse

As far as I know, it's not used by `git log`, but I imagine in the future it could be nice to update [the changelog generator](https://github.com/jshint/jshint/blob/278ad75d5550365fbfcb03c1fc4fb271b870d9e8/bin/changelog) to honor this file, too.

lukeapage:
Is this good to land or is it awaiting something?

jugglinmike:
@lukeapage I had been making some tweaks while still contacting folks, and I since forgot about this. The changes are pretty much settled now (if I find any more, I'll just submit a new pull request). Squashed; I'll merge in a few minutes

PR #2196, opened 2015-02-26
[[FEAT]] parse and lint tagged template literals

### LINKED ISSUE #2000: Tagged Template Strings are not linting correctly
I'm getting a ```Missing semicolon.``` error when using a tagged template string. If I have time I'll try to submit a test case for this soon.

My code looks roughly like:
```
function tag() {}

var testStr = 'bar';
function foo() {
  return tag `some
    ${testStr} here`;
}
foo();

```

caitp:
Closes #2000

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -875,10 +875,6 @@ var JSHINT = (function() {
     if (state.tokens.next.id === "(end)")
       error("E006", state.tokens.curr);
 
-    if (state.tokens.next.type === "(template)") {
-      doTemplateLiteral();
-    }
-
     var isDangerous =
       state.option.asi &&
       state.tokens.prev.line !== startLine(state.tokens.curr) &&
@@ -904,7 +900,9 @@ var JSHINT = (function() {
         error("E030", state.tokens.curr, state.tokens.curr.id);
       }
 
-      while (rbp < state.tokens.next.lbp && !isEndOfExpr()) {
+      // TODO: use pratt mechanics rather than special casing template tokens
+      while ((rbp < state.tokens.next.lbp || state.tokens.next.type === "(template)") &&
+              !isEndOfExpr()) {
         isArray = state.tokens.curr.value === "Array";
         isObject = state.tokens.curr.value === "Object";
 
@@ -1491,6 +1489,8 @@ var JSHINT = (function() {
 
   function parseFinalSemicolon() {
     if (state.tokens.next.id !== ";") {
+      // don't complain about unclosed templates / strings
+      if (state.tokens.next.isUnclosed) return advance();
       if (!state.option.asi) {
         // If this is the last statement in a block that ends on
         // the same line *and* option lastsemic is on, ignore the warning.
@@ -1985,24 +1985,37 @@ var JSHINT = (function() {
     }
   };
 
-  state.syntax["(template)"] = {
-    type: "(template)",
+  var baseTemplateSyntax = {
     lbp: 0,
     identifier: false,
-    nud: doTemplateLiteral
+    template: true,
   };
+  state.syntax["(template)"] = _.extend({
+    type: "(template)",
+    nud: doTemplateLiteral,
+    led: doTemplateLiteral,
+    noSubst: false
+  }, baseTemplateSyntax);
 
-  type("(template middle)", function() {
-    return this;
-  });
+  state.syntax["(template middle)"] = _.extend({
+    type: "(template middle)",
+    middle: true,
+    noSubst: false
+  }, baseTemplateSyntax);
 
-  type("(template tail)", function() {
-    return this;
-  });
+  state.syntax["(template tail)"] = _.extend({
+    type: "(template tail)",
+    tail: true,
+    noSubst: false
+  }, baseTemplateSyntax);
 
-  type("(no subst template)", function() {
-    return this;
-  });
+  state.syntax["(no subst template)"] = _.extend({
+    type: "(template)",
+    nud: doTemplateLiteral,
+    led: doTemplateLiteral,
+    noSubst: true,
+    tail: true // mark as tail, since it's always the last component
+  }, baseTemplateSyntax);
 
   type("(regexp)", function() {
     return this;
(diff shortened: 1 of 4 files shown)


caitp:
This was really hard to make work, the pratt parser is danged confusing =) anyways, there's probably a lot of gunk in this patch that isn't really needed, so lets try to figure out what can be safely removed before merging

src/jshint.js
```diff
@@ -1490,7 +1488,19 @@ var JSHINT = (function() {
   }
 
   function parseFinalSemicolon() {
+    if (lex.inContext(Context.Template)) return;
```
caitp: I'm not sure if this is really needed... technically template expressions aren't parsed using statement() so they shouldn't need this...

src/jshint.js
```diff
@@ -1490,7 +1488,19 @@ var JSHINT = (function() {
...
     if (state.tokens.next.id !== ";") {
+      if (state.tokens.next.isUnclosed) {
+        advance();
+        return;
+      }
+      if (state.tokens.next.id === "(end)") {
+        // If we got here via an unclosed template literal, don't complain.
+        if ((state.tokens.curr.template && state.tokens.curr.isUnclosed) ||
```
caitp: these could probably be organized better ._.

src/jshint.js
```diff
@@ -1985,24 +1996,37 @@ var JSHINT = (function() {
...
-
-  type("(no subst template)", function() {
-    return this;
-  });
+  state.syntax["(template)"] = _.extend({
+    type: "(template)",
+    nud: doTemplateLiteral,
+    led: doTemplateLiteral,
+    noSubst: false
+  }, baseTemplateSyntax);
```
caitp: i like the metadata more than relying on just the "type" property. wdyt?

src/jshint.js
```diff
@@ -2903,20 +2927,40 @@ var JSHINT = (function() {
...
-                 state.tokens.next.type !== "(end)") {
-        expression(10); // should probably have different rbp?
+  function doTemplateLiteral(left) {
+    // ASSERT: this.type === "(template)"
+    // jshint validthis: true
+    var ctx = this.context;
+    var noSubst = this.noSubst;
+    var depth = this.depth;
+
+    if (this === state.tokens.next) advance();
```
caitp: I don't think this check is needed anymore, since expression() will always use led/nud

src/jshint.js
```diff
@@ -2903,20 +2927,40 @@ var JSHINT = (function() {
...
+        }
       }
     }
+
+    var tag = left;
     return {
       id: "(template)",
-      type: "(template)"
+      type: "(template)",
+      tag: tag
```
caitp: maybe we can do something useful with this at some point

src/jshint.js
```diff
@@ -2903,20 +2927,40 @@ var JSHINT = (function() {
...
+        if (!state.tokens.next.template || state.tokens.next.depth > depth) {
+          expression(0); // should probably have different rbp?
+        } else {
+          // skip template start / middle
+          advance();
+        }
       }
     }
+
+    var tag = left;
```
caitp: admittedly, declaring this isn't particularly useful

src/jshint.js
```diff
@@ -2903,20 +2927,40 @@ var JSHINT = (function() {
...
     return {
       id: "(template)",
-      type: "(template)"
+      type: "(template)",
+      tag: tag
     };
+
+    function end() {
+      if (state.tokens.curr.template && state.tokens.curr.tail &&
```
caitp: I know this is confusing, but there are a lot of things to juggle around u_U

src/lex.js
```diff
@@ -129,7 +129,23 @@ Lexer.prototype = {
...
+
+  popContext: function() {
+    return this.context.pop();
+  },
+
+  isContext: function(context) {
+    return this.context.length > 0 && this.context[this.context.length - 1] === context;
+  },
+
+  currentContext: function() {
```
caitp: currentContext() is probably not needed

src/lex.js
```diff
@@ -1620,6 +1642,18 @@ Lexer.prototype = {
       if (token && token.startLine && token.startLine !== this.line) {
         obj.startLine = token.startLine;
       }
+      if (token && token.context) {
```
caitp: there's a lot of extra metadata that the parser needs to use

caitp:
did some self review, another set of eyes would be good though

src/lex.js
caitp: nevermind, I guess it is used

src/jshint.js
```diff
@@ -904,7 +900,8 @@ var JSHINT = (function() {
         error("E030", state.tokens.curr, state.tokens.curr.id);
       }
 
-      while (rbp < state.tokens.next.lbp && !isEndOfExpr()) {
+      while ((rbp < state.tokens.next.lbp || state.tokens.next.type === "(template)") &&
```
caitp: todo: don't special case "(template)", just use the pratt mechanics for this

src/jshint.js
rwaldron: I like it. Better metadata is always a win.

src/jshint.js
rwaldron: Ping me when you're ready for another look

caitp: it's a TODO, i'm not sure how to make it fit into the pratt mechanics. I'll just add a comment marking it as such.

rwaldron: On Thu Feb 26 2015 at 12:17:05 PM ⭐caitp⭐ <notifications@github.com> wrote:

> In src/jshint.js
> <https://github.com/jshint/jshint/pull/2196#discussion_r25445080>:
>
> > @@ -904,7 +900,8 @@ var JSHINT = (function() {
> >          error("E030", state.tokens.curr, state.tokens.curr.id);
> >        }
> >
> > -      while (rbp < state.tokens.next.lbp && !isEndOfExpr()) {
> > +      while ((rbp < state.tokens.next.lbp || state.tokens.next.type === "(template)") &&
>
> it's a TODO, i'm not sure how to make it fit into the pratt mechanics.
> I'll just add a comment marking it as such.
>
Got it.

>

src/jshint.js
```diff
@@ -1985,24 +1985,37 @@ var JSHINT = (function() {
...
+    noSubst: false
+  }, baseTemplateSyntax);
+
+  state.syntax["(no subst template)"] = _.extend({
+    type: "(template)",
+    nud: doTemplateLiteral,
+    led: doTemplateLiteral,
+    noSubst: true,
+    tail: true // mark as tail, since it's always the last component
+  }, baseTemplateSyntax);
```
rwaldron: This is really nice separation.

ruimarinho:
@caitp, found a small issue and ticketed it as https://github.com/jshint/jshint/issues/2223. Thank you for adding this feature!

Issue #2197, opened 2015-02-26
ES6 (esnext:true), jshint says "Foo is not defined"... but it's defined!


trusktr:
I have some code that looks like

```js
import _ from 'underscore';
import template from 'header.hbs';
import ViewBase from 'ViewBase';

/**
 * @class Header
 * @extends {ViewBase}
 */
export default
class Header extends ViewBase {

  /*
   * Method documentation.
   */
  someMethod() {
    // ...
  }
}

// Other stuff to document for the Header class, using underscore to extend the prototype.
_(Header.prototype).extend({

  /*
   * @property {Function} template A handlebars template function.
   */
  template: template
});
```

and jshint says

```
/path/to/views/header.js
  line 21  col 3  'Header' is not defined.

  ⚠  1 warning
```

The app works, Header is defined, and the template loads. This seems like a bug with the esnext feature.

trusktr:
I found the problem. jshint fails with

```js
export default class Name extends Other {}
console.log(Name); // error, Name is not defined.
```

but works with

```js
class Name extends Other {}
console.log(Name); // no error
export default Name;
```

trusktr:
Is that a problem with jshint? Or is there something in the ES6 spec that prevents the default export from being used after being exported?

jugglinmike:
When part of an `export` statement, the class becomes a ClassExpression, not a ClassDeclaration. ClassExpressions do not create a new entry in the environment record.

rwaldron:
@jugglinmike This was the case, but changed at some point in the last 6 months (I don't have the bandwidth to search at the moment), but this should suffice: https://people.mozilla.org/~jorendorff/es6-draft.html#sec-exports-runtime-semantics-evaluation

PR #2198, opened 2015-02-26
[[Fix]] export default * are defined in lexical module scope. Fixes gh-2197

rwaldron:
cc @jugglinmike @caitp 

- `class default function Identifier() {}`
- `class default class Identifier() {}`
- `class default class Identifier extends Heritage() {}`

In all cases, there is a lexical binding, `Identifier`, created.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3561,12 +3561,13 @@ var JSHINT = (function() {
     return classdef.call(this, true);
   });
 
-  function classdef(stmt) {
+  function classdef(isStatement) {
+
     /*jshint validthis:true */
     if (!state.option.inESNext()) {
       warning("W104", state.tokens.curr, "class");
     }
-    if (stmt) {
+    if (isStatement) {
       // BindingIdentifier
       this.name = identifier();
       addlabel(this.name, { type: "unused", token: state.tokens.curr });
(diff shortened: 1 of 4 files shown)


src/jshint.js
```diff
@@ -3561,12 +3561,13 @@ var JSHINT = (function() {
...
 
-  function classdef(stmt) {
+  function classdef(isStatement) {
+
     /*jshint validthis:true */
     if (!state.option.inESNext()) {
       warning("W104", state.tokens.curr, "class");
     }
-    if (stmt) {
+    if (isStatement) {
```
rwaldron: This is unrelated, but I couldn't help myself. Feel free to reject, I'll update accordingly.

src/jshint.js
```diff
@@ -4449,7 +4453,20 @@ var JSHINT = (function() {
       if (state.tokens.next.id === "function" || state.tokens.next.id === "class") {
         this.block = true;
       }
-      this.exportee = expression(10);
```
rwaldron: `this.exportee` wasn't used for anything.

src/jshint.js
```diff
@@ -4449,7 +4453,20 @@ var JSHINT = (function() {
...
+
+      token = peek(0);
```
rwaldron: Needed to get a name for the function case.

jugglinmike: `0` is unnecessary

PR #2222, opened 2015-02-28
[[FIX]] allow trailing comma in ArrayBindingPattern

caitp:
Also, disallow commas after rest elements in ArrayBindingPattern.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1467,6 +1467,9 @@ var JSHINT = (function() {
     if (state.tokens.next.value === "...") {
       if (norest === true) {
         warning("E024", state.tokens.next, "...");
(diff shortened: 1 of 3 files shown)


rwaldron:
I read right past this earlier today:

caitp:
yup, me too... that last `opt` was invisible :(

rwaldron:
As ridiculous as 4 releases in two days is, I think this is important to have in the wild.

caitp:
I'm going to try to fix the other issue in #2223 first, if you want to wait a few hours

caitp:
fixup removes handling of "..." in advance(), since it's handled specially everywhere that it's valid.

PR #2230, opened 2015-03-03
[[FIX]] Correctly enforce maxparams:0

mbildner:
It protects against a relatively benign edge case, but it is absolutely incorrect behavior for the parser to fail to enforce `maxparams` when it is set to 0.

Previously setting maxparams:0 would cause the check to short-circuit,
permitting any number of parameters without issuing a warning. This PR
fixes the bug by testing whether maxparams is a number, instead of
whether it is truthy.


mbildner:
@mikesherov can we merge this in? It's a quick and easy fix to a silly bug.

Issue #1552, opened 2014-02-26
--exclude-path doesn't behave correctly with absolute paths


iffy:
Here are steps to duplicate the problem:

```
cat <<EOF > ignorefile
*.js
**/*.js
EOF
echo "not really javascript" >> foo.js
```

This works as expected:

```
jshint --exclude-path ignorefile .
```

This doesn't:

```
jshint --exclude-path ignorefile $(pwd)
```

producing

```
/tmp/bar/foo.js: line 1, col 1, Expected an assignment or function call and instead saw an expression.
/tmp/bar/foo.js: line 1, col 4, Missing semicolon.
/tmp/bar/foo.js: line 1, col 5, Expected an assignment or function call and instead saw an expression.
/tmp/bar/foo.js: line 1, col 11, Missing semicolon.
/tmp/bar/foo.js: line 1, col 12, Expected an assignment or function call and instead saw an expression.
/tmp/bar/foo.js: line 1, col 22, Missing semicolon.

6 errors
```

jodonnell-broadsoft:
I know this is old but I still see it happening. It makes jshint ignore quite confusing. 

Can we get this fixed?

PR #2232, opened 2015-03-03
[[FIX]] support absolute path on --exclude and --exclude-path and symbolic links

lejenome:
add support for absolute path on `--exclude` and `--exclude-path` plus better symbolic links handling for these two options and for symbolic link to files to lint.

This should fix https://github.com/jshint/jshint/issues/1552 and https://github.com/plone/plone.recipe.codeanalysis/issues/73

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -198,7 +198,13 @@ function loadIgnores(params) {
   }
 
   var lines = (file ? shjs.cat(file) : "").split("\n");
-  lines.unshift(params.exclude || "");
+  if(params.exclude && path.isAbsolute(params.exclude)) {
+    var exclude = fs.realpathSync(params.exclude);
+    exclude = exclude.slice(path.resolve(path.dirname(file)).length + 1);
+    lines.unshift(exclude);
+  } else {
+    lines.unshift(params.exclude || "");
+  }
 
   return lines
     .filter(function(line) {
--- a/src/cli.js
+++ b/src/cli.js
@@ -196,10 +196,13 @@ function loadIgnores(params) {
   if (!file && !params.exclude) {
     return [];
   }
-  file = fs.realpathSync(file);
+  if (file)
+    file = fs.realpathSync(file);
 
   var lines = (file ? shjs.cat(file) : "").split("\n");
-  if(params.exclude && path.isAbsolute(params.exclude)) {
+  if (params.exclude && path.resolve(params.exclude) ===
+      path.normalize(params.exclude).replace(RegExp(path.sep + "$"), "")) {
+      // params.exclude is an absolute path
     var exclude = fs.realpathSync(params.exclude);
     exclude = exclude.slice(path.dirname(file).length + 1);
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -196,9 +196,19 @@ function loadIgnores(params) {
   if (!file && !params.exclude) {
     return [];
   }
+  if (file)
+    file = fs.realpathSync(file);
 
   var lines = (file ? shjs.cat(file) : "").split("\n");
-  lines.unshift(params.exclude || "");
+  if (params.exclude && path.resolve(params.exclude) ===
```
lukeapage: could you use `path.isAbsolute` here?

src/cli.js
```diff
@@ -196,9 +196,19 @@ function loadIgnores(params) {
   if (!file && !params.exclude) {
     return [];
   }
+  if (file)
+    file = fs.realpathSync(file);
```
lukeapage: is this common? Sorry if it is, I just haven't seen other node build tools use realpathSync before.

src/cli.js
```diff
@@ -196,9 +196,19 @@ function loadIgnores(params) {
   if (!file && !params.exclude) {
     return [];
   }
+  if (file)
```
lukeapage: most existing code uses curly brackets for blocks.. please could you add?

src/cli.js
lukeapage: this could be simplified by assigning exclude e.g.
```
var exclude = params.exclude || "";
if (exclude && path.isAbsolute(exclude)) {
  exclude = ...
}
lines.unshift(exclude);
```

src/cli.js
```diff
@@ -196,9 +196,19 @@ function loadIgnores(params) {
...
+    file = fs.realpathSync(file);
 
   var lines = (file ? shjs.cat(file) : "").split("\n");
-  lines.unshift(params.exclude || "");
+  if (params.exclude && path.resolve(params.exclude) ===
+      path.normalize(params.exclude).replace(RegExp(path.sep + "$"), "")) {
+      // params.exclude is an absolute path
+    var exclude = fs.realpathSync(params.exclude);
+    exclude = exclude.slice(path.dirname(file).length + 1);
```
lukeapage: sorry, not sure what this line is doing? please could you add a comment?

lukeapage:
my main concern is around the use of `realpath` and how "normal" it is to be resolving symlinks for excluded paths.
@iffy I know it was a while ago you raised #1552 and left https://github.com/jshint/jshint/issues/1552#issuecomment-36172168 but would you care to comment?

lejenome:
After submiting PR, I remeber I found that the change while cause issues on some use cases. until I rember them again, I will fix them

lukeapage:
@lejenome ok thanks for letting us know

lejenome:
Closing the issue because I couldn't reproduce the issue last time (since last comment).

PR #2233, opened 2015-03-03
Reduce array size by encoding sequences.

gloryknight:
Reduce output size about 30 times by encoding number sequences into an array!
The resulting arrays are reconstructed during execution.
The encoded array is composed from sequence of [starting number, number of incremental values in a sequence].

### DIFF
--- a/scripts/generate-identifier-data.js
+++ b/scripts/generate-identifier-data.js
@@ -44,13 +44,26 @@ var generateES5Data = function() { // ES 5.1
 
 var fs = require('fs');
 var writeFile = function(fileName, data) {
-  fs.writeFileSync(
+  var old=data[0], n=1, s="[", c=old;
+  for (var i=1;i<data.length;i++){ // encode continuous sequences of numbers into elements of array [number, sequence length]
+  	p=data[i];
+  	if (old!=p){
+  		s=s+"["+c+","+ n +"],";
+  		n=1;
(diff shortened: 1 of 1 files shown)


gloryknight:
It is too much time consuming  to sign the Contributor License Agreement. Just take my word that I stick to the conditions.

caitp:
> It is too much time consuming to sign the Contributor License Agreement. Just take my word that I stick to the conditions.

That doesn't work.

gloryknight:
Now CLA is signed

scripts/generate-identifier-data.js
```diff
@@ -44,13 +44,26 @@ var generateES5Data = function() { // ES 5.1
 
 var fs = require('fs');
 var writeFile = function(fileName, data) {
-  fs.writeFileSync(
+  var old=data[0], n=1, s="[", c=old;
+  for (var i=1;i<data.length;i++){ // encode continuous sequences of numbers into elements of array [number, sequence length]
```
lukeapage: It would be good to use descriptive variable names, not just n/s/c.
It would also be good to follow the style from the rest of the project - it isn't enforced in this file, but may be in the future, e.g. missing spaces, like between `n=1`, tab size should be 2 spaces.

gloryknight:
style is adjusted.

PR #2234, opened 2015-03-04
[[FEAT]] add `varstmt` enforcement option to disallow use of VariableStatements

### LINKED ISSUE #1549: Option to disallow "var" keyword with esnext: true
`esnext: true` adds support for `let` and `const` which make `var` obsolete in most cases, esp. in node.js environments. It would be great to have an option to warn about variables declared with `var` since its usage is most likely a mistake in ES6 files.

caitp:
Closes #1549

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -268,6 +268,16 @@ exports.bool = {
     singleGroups: false,
 
     /**
+     * When set to true, the use of VariableStatements are forbidden.
+     * For example:
+     *
+     *     // jshint varstmt: true
+     *
+     *     var a; // Warning: `var` declarations are forbidden. Use `let` or `const` instead.
+     */
+    varstmt: false,
+
+    /**
      * This option is a short hand for the most strict JSHint configuration. It
      * enables all enforcing options and disables all relaxing options.
      *
@@ -944,4 +954,10 @@ exports.removed = {
   gcl: true,
   smarttabs: true,
   trailing: true
+};
+
+// Add options here which should not be automatically enforced by
+// `enforceall`.
+exports.noenforceall = {
+  varstmt: true
 };
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -178,7 +178,8 @@ var JSHINT = (function() {
   function processenforceall() {
     if (state.option.enforceall) {
       for (var enforceopt in options.bool.enforcing) {
-        if (state.option[enforceopt] === undefined) {
+        if (state.option[enforceopt] === undefined &&
+            !options.noenforceall[enforceopt]) {
           state.option[enforceopt] = true;
         }
       }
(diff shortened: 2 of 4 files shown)


rwaldron:
I love this so much.

src/options.js
```diff
@@ -268,6 +268,16 @@ exports.bool = {
     singleGroups: false,
 
     /**
+     * When set to true, the use of VariableStatements are forbidden.
+     * For example:
```
jugglinmike: Although not all the option descriptions include this, I think it's helpful if the explanation also has some text explaining the motivation for the feature. For example:

> `let` and `const` statements are block scoped, so they offer more precise control over the availability of a binding and more informative static errors. The function scoping of `var` statements enables usage patterns that are difficult to read and error-prone.

...but I'd be happy with whatever text you'd like to include.

src/options.js
```diff
@@ -945,3 +955,9 @@ exports.removed = {
   smarttabs: true,
   trailing: true
 };
+
+// Add options here which should not be automatically enforced by
+// `enforceall`.
+exports.noenforceall = {
```
jugglinmike: This is obviously necessary, but it does make the "enforceall" option a little confusing ("enforcesome"?). What do you think about adding to the documentation of the `enforceall` option that (in addition to being deprecated) it only enforces the options that were available in the 2.6.3 release of JSHint?

src/options.js
```diff
@@ -268,6 +268,16 @@ exports.bool = {
...
 
     /**
+     * When set to true, the use of VariableStatements are forbidden.
+     * For example:
+     *
+     *     // jshint varstmt: true
+     *
+     *     var a; // Warning: `var` declarations are forbidden. Use `let` or `const` instead.
+     */
+    varstmt: false,
```
jugglinmike: I think this name will make configuration files a little confusing since `varstmt: true` means "no `var` statements allowed". What about `novarstmt` or just `novar`?

jugglinmike:
@rwaldron @caitp Should JSHint throw a warning if this is enabled *without* `esnext`? I don't know if it would be a common enough mistake to warrant the extra code, but it probably couldn't hurt.

caitp:
> @rwaldron @caitp Should JSHint throw a warning if this is enabled without esnext? I don't know if it would be a common enough mistake to warrant the extra code, but it probably couldn't hurt.

I thought about emitting a warning, but it's awkward since you can have function-scoped options, seems awkward to do it. Another way to go would be making it implicate `esnext`, but I'm not keen on that

rwaldron:
> Should JSHint throw a warning if this is enabled without esnext?

...

> I thought about emitting a warning, but it's awkward since you can have function-scoped options, seems awkward to do it. Another way to go would be making it implicate esnext, but I'm not keen on that


I'm not really sure—let's take some time to think it over. No rush, right?

caitp:
yup

Issue #1825, opened 2014-08-26
use of `!` with `instanceof` should require parentheses


phpnode:
There is no legitimate reason to use `!foo instanceof Foo`, it requires parens and must be written as `!(foo instanceof Foo)`. Currently jshint does not warn on this syntax even though it is *guaranteed* to be a bug/typo. 

For instance, I could not get jshint to reject the following:

```js
function Test () {}

var y;

if (!y instanceof Test) {
   Test(); 
}
```

#437 mentions this but is closed, so maybe this is a regression?

julienw:
Is it as simple as adding the case with "instanceof" in https://github.com/jshint/jshint/blob/6bf7653ec8eedcb35ea6825fd8b5a0abd2a4410e/src/jshint.js#L1150 ?

If yes I can give it a try.

rwaldron:
@julienw I'm not sure and I would also have to try to answer. Give it a shot!

julienw:
Crap, you didn't fall in my evil trap. :)
PR https://github.com/jshint/jshint/pull/2237

phpnode:
The fix has been merged in #1825 so I'm closing this, thanks @julienw

PR #2237, opened 2015-03-06
Fix issue #1825: use of `!` with `instanceof` should require parentheses

julienw:


rwaldron:
LGTM

@caitp @jugglinmike ?

caitp:
looks good to me, I'm surprised this wasn't already happening =) genuinely shocked

rwaldron:
Probably because Crockford wasn't using `new` and so never really used the code pattern that this would most likely appear in, therefore never considered :)

caitp:
so, the only thing is, it's possible this will break peoples builds --- there's a trivial workaround, but it's not very nice to ship something like that in a patch release.

Maybe, this error should get its own message to at least make it obvious how to fix any build failures that it triggers (and give it its own relaxation option to turn it off)

rwaldron:
> so, the only thing is, it's possible this will break peoples builds --- there's a trivial workaround, but it's not very nice to ship something like that in a patch release.

> Maybe, this error should get its own message to at least make it obvious how to fix any build failures that it triggers (and give it its own relaxation option to turn it off)


Nice catch :)

julienw:
I'm in holidays for 1 week (the kind of holidays without a computer) so feel free to take it from here :)

julienw:
> so, the only thing is, it's possible this will break peoples builds --- there's a trivial workaround, but it's not very nice to ship something like that in a patch release.

If this shipped in a minor release instead, would it be ok ?

caitp:
@julienw npm would still pull it in when people use the hat operator (like `^2.6.3` in package.json).

Maybe we could merge this and I could add some quick fixes to it tonight?

julienw:
ok, I get it: minor versions include only backward compatible changes, and this is not one.

This is not that urgent, this is waiting for several months, so I can do the needed change when I come back if necessary. No need to rush.

In case you intend to do a major version this week, feel free to merge :p

julienw:
\o/ thanks. And sorry for not taking the time to fix the comments ..

PR #2242, opened 2015-03-08
Add support for typeof symbol === "symbol"

### LINKED ISSUE #2241: Invalid typeof value 'symbol'
As of ES6 'symbol' is a valid.

thejameskyle:
Fixes #2241

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1230,7 +1230,7 @@ var JSHINT = (function() {
 
     var values = [
       "undefined", "object", "boolean", "number",
-      "string", "function", "xml", "object", "unknown"
+      "string", "function", "xml", "object", "unknown", "symbol"
     ];
 
     if (right.type === "(identifier)" && right.value === "typeof" && left.type === "(string)")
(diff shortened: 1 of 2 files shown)


caitp:
I think you're going to need to add another assertion to the tests that use that fixture, but otherwise it looks p. good. Will merge when tests are fixed

caitp:
actually, I guess you probably won't need to add an assertion, since it's testing that an error isn't thrown... I guess it's okay :>

thejameskyle:
It does error if I remove the fix

caitp:
yeah it's fine, we might want to do a followup to support the new futurehostile feature with it, which would call for some assertions --- but I think this is fine for now

thejameskyle:
> to support the new futurehostile feature with it,

I'm not sure I understand what that'd consist of.

caitp:
> I'm not sure I understand what that'd consist of.

It just means making it a bit more complicated so that people authoring for ES3/5 environments can keep new features out of their code

thejameskyle:
Should it selectively add `"symbol"` to the typeof list when this option is enabled?

caitp:
something like that, I'm not sure exactly. @jugglinmike is the expert on it, but it's probably something we can hold back on until someone complains

jugglinmike:
I don't think the follow-up should use `futurehostile` because that option is intended to help people avoid authoring code that is difficult to upgrade to future versions of the language. This feature doesn't raise any future portability concerns; in this case, we just want to continue warning about "symbol" in pre-ES6 environments.

Doing this will require additional abstraction around the `values` array to key off the configured language version.

PR #2243, opened 2015-03-08
[[Fix]] Only accept "symbol" as a type in ES6 envs

jugglinmike:
As of gh-2242, JSHint unconditionally accepts expressions that compare `typeof` results to the string "symbol". These expressions are only valid in ES6, so they should continue to trigger W122 warnings ("Invalid typeof value") in pre-ES6 code.

I've taken this as an opportunity to extend the internal `isTypoTypeof` function tto accept the `options` object as a formal parameter. This more functional approach avoids  relying on the `state` object through a closure, so the code becomes easier to read and easier to move. This is not strictly necessary for this patch, but it seemed like an appropriate improvement to make while making the change.

Commit message:

> The ES6 specification extended the `typeof` operator to return the
> string "symbol" when when applied to Symbol instances. In pre-ES6
> environments, the operator will never return this value, so comparisons
> that involve it cannot possibly evaluate to `true`.
>
> Update the internal `isTypoTypeof` function to allow a different set of
> values for different language versions.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1219,8 +1219,8 @@ var JSHINT = (function() {
 
   var typeofValues = {};
   typeofValues.es3 = [
-    "undefined", "object", "boolean", "number", "string", "function", "xml",
-    "object", "unknown"
+    "undefined", "boolean", "number", "string", "function", "xml", "object",
+    "unknown"
   ];
   typeofValues.es6 = typeofValues.es3.concat("symbol");
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1217,21 +1217,26 @@ var JSHINT = (function() {
         node.type === "undefined");
   }
 
+  var typeofValues = {};
+  typeofValues.es3 = [
+    "undefined", "object", "boolean", "number", "string", "function", "xml",
+    "object", "unknown"
+  ];
+  typeofValues.es6 = typeofValues.es3.concat("symbol");
+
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -1217,21 +1217,26 @@ var JSHINT = (function() {
         node.type === "undefined");
   }
 
+  var typeofValues = {};
+  typeofValues.es3 = [
+    "undefined", "object", "boolean", "number", "string", "function", "xml",
+    "object", "unknown"
```
rwaldron: - `xml` and `unknown` are not valid. 
- `object` appears twice.

caitp:
we could delete the invalid ones just to see if anyone notices :D

jugglinmike:
> - object appears twice.

Removed

> - xml and unknown are not valid.
> note: I'm aware that you didn't write this, but we should figure out why these invalid types have been included. It might make sense to include a special typeofValues.bullshitIEAdded

Done. Info on these was surprisingly tough to track down

> we could delete the invalid ones just to see if anyone notices :D

I was very tempted to do just that. A slightly less aggressive (though still breaking) change would be to *only* add these values to the `es3` array--this would disallow those values for anyone with `esnext: true`.

I actually think that would be justifiable. It's true that the change may produce new errors in previously-passing code, but in the case of ES6, that code is technically incorrect. (Contrast this with ES3 code which may be running in IE8--here, "unknown" is unfortunate but necessary.) Thoughts on that?

PR #2249, opened 2015-03-11
Reduce array size by encoding sequences.

gloryknight:
Just included changes proposed in the previous post. Observe the huge reduction in the file size!


jugglinmike:
> Just included changes proposed in the previous post.

Are you referring to gh-2233?

gloryknight:
Yes I do refer to gh-2233.

jugglinmike:
Okay, cool. These changes are to files that are automatically built as a result of the scripts modified in gh-2233. We don't generally accept changesets to built files. It wouldn't make sense to accept this patch without gh-2233 (the very next build would un-do the changes). If gh-2233 is accepted, you can expect the built file to automatically be updated in the next release

PR #2252, opened 2015-03-12
[[FIX]] es6: export on var, let, const exports all variables, not just the first one

### LINKED ISSUE #2248: defined but never used on exported vars
```
export var a = "1", b = "2";
```

gives `'a' is defined but never used` - which is true, but it is exported.

I could do
```
var a = "1", b = "2";
export {a, b};
```

but its basically the same code, just more verbose

Same occurs with let

lukeapage:
Fixes #2248 

Looks like code will have to be expanded again to work with export and de-structuring in the same var/let/const but I think its alot more work. This looks to me like a move in the right direction and fixes the bug and add tests.

I've signed the CLA.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -863,7 +863,7 @@ var JSHINT = (function() {
       funct["(blockscope)"].stack();
       advance("let");
       advance("(");
-      state.syntax["let"].fud.call(state.syntax["let"].fud);
+      state.tokens.prev.fud();
       advance(")");
     }
 
@@ -3415,12 +3415,12 @@ var JSHINT = (function() {
         tokens = destructuringExpression();
         lone = false;
       } else {
-        if (inexport) {
-          exported[state.tokens.next.value] = true;
-          state.tokens.next.exported = true;
-        }
         tokens = [ { id: identifier(), token: state.tokens.curr } ];
         lone = true;
+        if (inexport) {
+          exported[state.tokens.curr.value] = true;
+          state.tokens.curr.exported = true;
+        }
       }
       for (var t in tokens) {
         if (tokens.hasOwnProperty(t)) {
@@ -3486,12 +3486,12 @@ var JSHINT = (function() {
         tokens = destructuringExpression();
         lone = false;
       } else {
-        if (inexport) {
-          exported[state.tokens.next.value] = true;
-          state.tokens.next.exported = true;
-        }
         tokens = [ { id: identifier(), token: state.tokens.curr } ];
         lone = true;
+        if (inexport) {
+          exported[state.tokens.curr.value] = true;
+          state.tokens.curr.exported = true;
+        }
       }
       for (var t in tokens) {
         if (tokens.hasOwnProperty(t)) {
@@ -3576,12 +3576,12 @@ var JSHINT = (function() {
         tokens = destructuringExpression();
         lone = false;
       } else {
-        if (inexport) {
-          exported[state.tokens.next.value] = true;
-          state.tokens.next.exported = true;
-        }
         tokens = [ { id: identifier(), token: state.tokens.curr.value } ];
         lone = true;
+        if (inexport) {
+          exported[state.tokens.curr.value] = true;
+          state.tokens.curr.exported = true;
+        }
       }
       for (var t in tokens) {
         if (tokens.hasOwnProperty(t)) {
@@ -4169,13 +4169,13 @@ var JSHINT = (function() {
 
       if (state.tokens.next.id === "var") {
         advance("var");
-        state.syntax["var"].fud.call(state.syntax["var"].fud, { prefix: true });
+        state.tokens.curr.fud({ prefix: true });
       } else if (state.tokens.next.id === "let") {
         advance("let");
         // create a new block scope
         letscope = true;
         funct["(blockscope)"].stack();
-        state.syntax["let"].fud.call(state.syntax["let"].fud, { prefix: true });
+        state.tokens.curr.fud({ prefix: true });
       } else if (!state.tokens.next.identifier) {
         error("E030", state.tokens.next, state.tokens.next.type);
         advance();
@@ -4242,13 +4242,13 @@ var JSHINT = (function() {
       if (state.tokens.next.id !== ";") {
         if (state.tokens.next.id === "var") {
           advance("var");
-          state.syntax["var"].fud.call(state.syntax["var"].fud);
+          state.tokens.curr.fud();
         } else if (state.tokens.next.id === "let") {
           advance("let");
           // create a new block scope
           letscope = true;
           funct["(blockscope)"].stack();
-          state.syntax["let"].fud.call(state.syntax["let"].fud);
+          state.tokens.curr.fud();
         } else {
           for (;;) {
             expression(0, "for");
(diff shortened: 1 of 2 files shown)


caitp:
oh man, I thought this was filed on traceur for some reason. Mark this as fixing the issue I just opened then, hah =)

src/jshint.js
```diff
@@ -863,7 +863,7 @@ var JSHINT = (function() {
       funct["(blockscope)"].stack();
       advance("let");
       advance("(");
-      state.syntax["let"].fud.call(state.syntax["let"].fud, false);
+      state.syntax["let"].fud.call(state.syntax["let"].fud);
```
caitp: @rwaldron do you have any idea why we're calling it this way instead of `state.tokens.curr.fud(false);`?

lukeapage: seems like it was a common pattern - found lots of commits using it - https://github.com/jshint/jshint/commit/5924b2aa5aafdf8fede525b7156bd1962f824a14#diff-180beb5359ee54413144b3d8200d7504R3975

one thing is I didn't even notice it because its hard to compare & notice that `state.syntax["let"].fud` is repeated identically.

removing it doesn't have any effect, further notice that commit has a mix of different styles. I can't find a single committer introducing one style or the other.

Most of the other non-let/var/const don't use call. My vote would be to remove the call. I have an extra commit waiting, just say the word.

caitp: The problem is if any of these constructs ever mutate their `this` binding, then you end up with problems if calling them that way, so I would prefer we didn't do that

caitp: that, and it's just harder to read the `.call()` invocations =)

lukeapage: you want me to push the change onto this PR?

caitp: I'd prefer it yeah

src/jshint.js
```diff
@@ -863,7 +863,7 @@ var JSHINT = (function() {
       funct["(blockscope)"].stack();
       advance("let");
       advance("(");
-      state.syntax["let"].fud.call(state.syntax["let"].fud, false);
+      state.syntax["let"].fud();
```
caitp: sorry, I meant I'd prefer `state.tokens.prev.nud()` --- where "prev" is the "let" token at this point

lukeapage: ah - then we get "Missing semicolon." errors... ?

caitp:
```diff
-      state.syntax["let"].fud();
+      state.tokens.prev.fud();
```

is what I meant --- that will give it the appropriate context

src/jshint.js
```diff
@@ -3562,6 +3576,10 @@ var JSHINT = (function() {
         tokens = destructuringExpression();
         lone = false;
       } else {
+        if (inexport) {
+          exported[state.tokens.next.value] = true;
+          state.tokens.next.exported = true;
+        }
```
caitp: setting the exported flag should happen after the call to `identifier()` below (which will make sure that it's actually an identifier)

src/jshint.js
```diff
@@ -3478,6 +3486,10 @@ var JSHINT = (function() {
...
```
caitp: same up here imo

src/jshint.js
```diff
@@ -4151,13 +4169,13 @@ var JSHINT = (function() {
 
       if (state.tokens.next.id === "var") {
         advance("var");
-        state.syntax["var"].fud.call(state.syntax["var"].fud, true);
+        state.syntax["var"].fud({ prefix: true });
```
caitp: we should really be using `state.tokens.curr.fud({ prefix: true });` here

src/jshint.js
```diff
@@ -4151,13 +4169,13 @@ var JSHINT = (function() {
...
       } else if (state.tokens.next.id === "let") {
         advance("let");
         // create a new block scope
         letscope = true;
         funct["(blockscope)"].stack();
-        state.syntax["let"].fud.call(state.syntax["let"].fud, true);
+        state.syntax["let"].fud({ prefix: true });
```
caitp: same idea here, `state.tokens.curr.fud({ prefix: true });`

src/jshint.js
```diff
@@ -4224,13 +4242,13 @@ var JSHINT = (function() {
       if (state.tokens.next.id !== ";") {
         if (state.tokens.next.id === "var") {
           advance("var");
-          state.syntax["var"].fud.call(state.syntax["var"].fud);
+          state.syntax["var"].fud();
```
caitp: and again, `state.tokens.curr.fud();`

src/jshint.js
```diff
@@ -4224,13 +4242,13 @@ var JSHINT = (function() {
...
         } else if (state.tokens.next.id === "let") {
           advance("let");
           // create a new block scope
           letscope = true;
           funct["(blockscope)"].stack();
-          state.syntax["let"].fud.call(state.syntax["let"].fud);
+          state.syntax["let"].fud();
```
caitp: and another, `state.tokens.curr.fud();`

src/jshint.js
```diff
@@ -4583,21 +4601,15 @@ var JSHINT = (function() {
     if (state.tokens.next.id === "var") {
       // ExportDeclaration :: export VariableStatement
       advance("var");
-      exported[state.tokens.next.value] = ok;
-      state.tokens.next.exported = true;
-      state.syntax["var"].fud.call(state.syntax["var"].fud);
+      state.syntax["var"].fud({ inexport:true });
```
caitp: `state.tokens.curr.fud({ inexport: true });`

src/jshint.js
```diff
@@ -4583,21 +4601,15 @@ var JSHINT = (function() {
...
     } else if (state.tokens.next.id === "let") {
       // ExportDeclaration :: export VariableStatement
       advance("let");
-      exported[state.tokens.next.value] = ok;
-      state.tokens.next.exported = true;
-      state.syntax["let"].fud.call(state.syntax["let"].fud);
+      state.syntax["let"].fud({ inexport:true });
```
caitp: `state.tokens.curr.fud({ inexport: true });`

src/jshint.js
```diff
@@ -4583,21 +4601,15 @@ var JSHINT = (function() {
...
     } else if (state.tokens.next.id === "const") {
       // ExportDeclaration :: export VariableStatement
       advance("const");
-      exported[state.tokens.next.value] = ok;
-      state.tokens.next.exported = true;
-      state.syntax["const"].fud.call(state.syntax["const"].fud);
+      state.syntax["const"].fud({ inexport:true });
```
caitp: `state.tokens.curr.fud({ inexport: true });`

caitp:
other than those comments it looks good to me

lukeapage:
Thanks, I understand now - was my first time looking at the code, so I didn't realise you could get curr and prev and that was the same as lookup on statements.

tests/unit/core.js
```diff
@@ -722,7 +722,10 @@ exports.testES6ModulesNamedExportsAffectUnused = function (test) {
     "export { a, x };",
     "export var b = { baz: 'baz' };",
     "export function boo() { return z; }",
-    "export class MyClass { }"
+    "export class MyClass { }",
+    "export var varone = 1, vartwo = 2;",
+    "export const constone = 1, consttwo = 2;",
+    "export let letone = 1, lettwo = 2;"
```
rwaldron: :+1:

caitp:
lgtm, thanks

PR #2264, opened 2015-03-18
[[FIX]] Incorrect 'Unclosed string' when the closing quote is the first character after a newline

### LINKED ISSUE #1319: Spurious 'Unclosed string' when strings are properly closed after a newline.
'two' in the code example below will trigger 'unclosed string' when the string is correctly terminated.

    /*jshint multistr:true */

    var one = 'one';

    var two = 'two \
    ';

    console.log(one, two)

No particular reason for the style of 'two' but it did pop up in some code and jshint doesn't handle it well at the moment.

### LINKED ISSUE #1532: Incorrect handling of multistr lines with closing quote on empty line
test = "closing quote on next line\
    ";

Gives erroneous warning "Unclosed string."

Added it to the `strings.js` file. I've been poking around the lexer but don't have high hopes for quickly finding the issue, although it might be related to `nextLine()`

jacksonmj:
Previously, the end-of-line loop was immediately followed by the non-quote character code. This
meant that the first character after an EOL was always treated as a non-quote character.

Control flow now returns to the quote character check in the while loop condition immediately
after each EOL or character has been handled, so that this check also runs for the first character
on a line.

Closes #1532. (The commit includes the test case provided in #1532 by DelvarWorld.)
Closes #1319.

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -1147,7 +1147,7 @@ Lexer.prototype = {
     this.skip();
 
     while (this.peek() !== quote) {
-      while (this.peek() === "") { // End Of Line
+      if (this.peek() === "") { // End Of Line
 
         // If an EOL is not preceded by a backslash, show a warning
         // and proceed like it was a legit multi-line string where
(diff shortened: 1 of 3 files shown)


caitp:
this patch basically looks okay, but technically even escaped line terminators are not really valid per spec, and don't seem to be covered in the annex.

@rwaldron wdyt?

jacksonmj:
Escaped line terminators seem to be valid based on my reading of the spec...

According to http://www.ecma-international.org/ecma-262/5.1/#sec-7.8.4 , a LineContinuation is a valid SingleStringCharacter or DoubleStringCharacter, so is allowed in a string (and has a string value of the empty character sequence). 
LineContinuation is defined as "\ LineTerminatorSequence", where LineTerminatorSequence is defined in http://www.ecma-international.org/ecma-262/5.1/#sec-7.3 and includes \n, \r\n, and \r. Therefore, LineContinuation is an escaped line terminator.

It's slightly non-obvious, because escaped line terminators are not defined in the same way as other escape sequences ("\ EscapeSequence"), presumably because that makes defining semantics easier, but I believe they are valid.


The docs for the jshint multistr option also seem to indicate that escaped line terminators are a style issue rather than a code correctness issue. (http://jshint.com/docs/options/#multistr , "this option allows correct multi-line strings", and describes them as dangerous but not incorrect).

caitp:
you're right, LineContinuation is valid. LGTM

PR #2266, opened 2015-03-19
[[FIX]] default to empty string in src/cli.js loadIgnores

stringparser:
Related to https://github.com/iojs/io.js/pull/1153.

They added type checking for ~~`path.dirname`~~ the `path` library and this line below fails

````js
return path.join(path.dirname(file), line.trim());
````


stringparser:
ping

Issue #2260, opened 2015-03-15
Multiplication and Division are not associative, grouping IS necessary


mMerlin:
Environments:
http://jshint.com/ as of 2015/03/15
Windows 7, FireFox 36.0.1
Fedora 21, Google Chrome 41.0.2272.76

```javascript
/* jshint singleGroups: true */
var s1 = 10, s0 = 5, dt = 0.5, a0, a1;
a0 = (s1 - s0) / (dt * dt);
a1 = s1 / (dt * dt);
console.log(a0);
console.log(a1);
a1 = s1 / dt * dt;
a0 = (s1 - s0) / dt * dt;
console.log(a0);
console.log(a1);
```

jslint.com complains about "Unnecessary grouping operator" for lines 3 and 4.  The grouping IS necessary.  Changing to … / dt / dt would work, but as long as multiple is used, the order of operation needs to be controlled to get the correct result.

rwaldron:
Thanks for the report. jslint.com is not this project, but I assume you actually meant jshint.com? Either way, I can repro this issue. 

cc @jugglinmike

mMerlin:
yes, jshint.  I use both, and fingers got on 'autopilot' in the description.  The environment info and code block both show jshint.  jslint does not report anything about the grouping.  IIRC, in some cases, jslint will recommend adding extra brackets for clarity.

jugglinmike:
On it

mMerlin:
Correct results with my simple test case at least, as of v2.7.0

PR #2270, opened 2015-03-21
Fix singleGroups bugs

jugglinmike:
This addresses two separate issues in everyone's favorite option, `singleGroups`.

The first commit resolves gh-2260
The second commit resolves an as-yet unreported bug regarding the option's behavior with relation operators.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1180,7 +1180,8 @@ var JSHINT = (function() {
 
     x.led = function(left) {
       nobreaknonadjacent(state.tokens.prev, state.tokens.curr);
-      var right = expression(100);
+      this.left = left;
+      var right = this.right = expression(100);
 
       if (isIdentifier(left, "NaN") || isIdentifier(right, "NaN")) {
         warning("W019", this);
@@ -1200,8 +1201,6 @@ var JSHINT = (function() {
         warning("W018", right, "!");
       }
 
-      this.left = left;
-      this.right = right;
       return this;
     };
     return x;
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2609,10 +2609,7 @@ var JSHINT = (function() {
           // begin with the `{` and `function` tokens
           (opening.beginsStmt && (ret.id === "{" || triggerFnExpr || isFunctor(ret))) ||
           // Used as the return value of a single-statement arrow function
-          (ret.id === "{" && preceeding.id === "=>") ||
-          // Used to prevent left-to-right application of adjacent addition
-          // operators (the order of which effect type)
-          (first.id === "+" && preceeding.id === "+");
+          (ret.id === "{" && preceeding.id === "=>");
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -2622,7 +2618,7 @@ var JSHINT = (function() {
       // first expression *or* the current group contains multiple expressions)
       if (!isNecessary && (first.left || ret.exprs)) {
         isNecessary =
-          (!isBeginOfExpr(preceeding) && first.lbp < preceeding.lbp) ||
+          (!isBeginOfExpr(preceeding) && first.lbp <= preceeding.lbp) ||
```
caitp: do we want to do it this way, or does it make more sense to adjust the lbp marker of the arithmetic operators?

jugglinmike: I initially tried that, but it failed to account for cases where like operators were being used on either side of the grouping operator. Since usages like `x = 1 / (2 / 3)` are valid, I think `<=` is the correct approach here.

apfelbox:
Another issue (not sure whether this is the same as above): the associativity between binary operators and the conditional operator is wrong:

```js
var a = 4 * (4 > 5) ? -1 : 1; // === 1
var a = 4 + (4 > 5) ? -1 : 1; // === -1

// ... and ...

var a = 4 * ((4 > 5) ? -1 : 1); // === 4
var a = 4 + ((4 > 5) ? -1 : 1); // === 5
```

But both are reported as violations of this rule:

    test.js: line 2, col 14, Unnecessary grouping operator.
    test.js: line 3, col 14, Unnecessary grouping operator.

jugglinmike:
@apfelbox  I believe the behavior you are reporting is already correct. Here's the code I'm linting:

```js
/* jshint singleGroups: true, unused: false */
var a = 4 * (4 > 5) ? -1 : 1; // === 1
var a = 4 + (4 > 5) ? -1 : 1; // === -1

// ... and ...

var a = 4 * ((4 > 5) ? -1 : 1); // === 4
var a = 4 + ((4 > 5) ? -1 : 1); // === 5
```

...and the output:

```
$ jshint tmp2.js
tmp2.js: line 7, col 14, Unnecessary grouping operator.
tmp2.js: line 8, col 14, Unnecessary grouping operator.

2 errors
$
```

I think this is correct because in that example, the inner grouping operators on line 7 and 8 are both unnecessary.

```js
4 * ((4 > 5) ? -1 : 1) // 4
4 * (4 > 5 ? -1 : 1) // 4
4 + ((4 > 5) ? -1 : 1) // 5
4 + (4 > 5 ? -1 : 1) // 5
```

Removing them precludes the error:

```js
/* jshint singleGroups: true, unused: false */
var a = 4 * (4 > 5) ? -1 : 1; // === 1
var a = 4 + (4 > 5) ? -1 : 1; // === -1

// ... and ...

var a = 4 * (4 > 5 ? -1 : 1); // === 4
var a = 4 + (4 > 5 ? -1 : 1); // === 5
```

```
$ jshint tmp2.js
$
```

Does that seem right to you?

apfelbox:
@jugglinmike lol, I didn't even think about the inner braces around the condition. I just looked at the outer braces, the ones around the complete ternary operator. :see_no_evil:

Thanks and yes, the current behavior is correct, of course.

jugglinmike:
No worries! I appreciate having more eyes on this feature, as it's been much trickier than I originally expected.

jugglinmike:
The "as-yet unreported" bug that the second commit addresses has just been reported: gh-2280 .

PR #2275, opened 2015-03-24
[[FEAT]] use /etc/.jshintrc as fallback

### LINKED ISSUE #2274: Look for jshintrc in /etc/
We have multiple users using `jshint` on a central development server.

To enforce the corporate coding standard for all users in the same way, it would be nice if jshint would look at `/etc/jshintrc`.

caitp:
Closes #2274


caitp:
ping @rwaldron / @jugglinmike / @mikesherov --- doesn't seem unreasonable to me, thoughts?

jugglinmike:
The recent discussion in gh-2274 seems reasonable to me. What do you think about switching this patch to look for `/etc/jshintrc` instead of `/etc/.jshintrc`?

The details regarding the XDG spec probably make sense to implement in a separate commit... although we should be careful to release both features together.

caitp:
yeah, not looking for a hidden file is fine --- but should it be jshintrc or jshint.rc? choices choices.

jugglinmike:
Arrrg

rwaldron:
Can we do both and land this?

PR #2285, opened 2015-04-02
[[FEAT]] `nobom` option to disallow BOM sequences

### LINKED ISSUE #2284: New option: nobom (Disallow ByteOrderMark)
Downstream: https://phabricator.wikimedia.org/T40233

PHPCS equivalent: [Sniffs/Generic.Files.ByteOrderMark](https://github.com/squizlabs/PHP_CodeSniffer/blob/2.3.0/CodeSniffer/Standards/Generic/Sniffs/Files/ByteOrderMarkSniff.php)

Ref #187, #253.
Ref jscs-dev/node-jscs#1225

caitp:
Closes #2284

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -5230,6 +5230,70 @@ var JSHINT = (function() {
       });
     }
 
+    if (o && o.nobom) {
+      // nobom:
+      // Report error unicode BOM sequence is found
+      s = (function(s) {
+        var kBOM = /(?:\uFFFE)|(?:\uFEFF)/g;
+        var lines;
+
+        if (Array.isArray(s)) {
+          lines = s;
+          return s.map(warnAndRemoveBOM);
+        } else {
+          return warnAndRemoveBOM(s);
+        }
+
+        function warnAndRemoveBOM(s, lineIndex) {
+          return s.replace(kBOM, function byteOrderMarkError(m, offset) {
+            var off = 0;
+            var didSplit = false;
+            if (!lines) {
+              didSplit = true;
+              lines = s.split(/([^\r\n]*?\r?\n)/);
+            }
+            if (isString(lines[0])) {
+              lines = lines.map(mapLinesToPositions);
+              if (didSplit) {
+                lines = lines.filter(filterPositions);
+              }
+            }
+            warning("W132", offsetToToken(lines, offset, lineIndex));
+
+            // Replace it with spaces, so that it doesn't break linting the rest
+            // of the file.
+            return m.replace(/./g, " ");
+
+            function mapLinesToPositions(l) {
+              return { start: off, end: (off = off + l.length) - 1 };
+            }
+
+            function filterPositions(p) {
+              // Split produces empty items at every even index in the array,
+              // filter them out.
+              return p.end > p.start;
+            }
+          });
+        }
+
+        function offsetToToken(lines, offset, hint) {
+          var line;
+          if (typeof hint === "number") {
+            line = lines[hint];
+            return { line: hint + 1, from: offset + 1 };
+          } else {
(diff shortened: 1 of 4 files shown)


caitp:
PTAL ---it's a lot of code (mostly to deal with the array input case :(

It probably needs a more "e2e"-style test with a fixture too, to make sure it works with utf8 as well as utf16 (unit tests only deal with utf16, since that's JS for you). But I'm not sure how hard it will be to keep the file from being mysteriously re-encoded or mysteriously have its BOM removed

jugglinmike:
The lexer [currently calls its `scanUnsafeChars` method](https://github.com/jshint/jshint/blob/12811e797a264f323b920b314db965b8adcb207e/src/lex.js#L1519) which executes [a regular expression search](https://github.com/jshint/jshint/blob/12811e797a264f323b920b314db965b8adcb207e/src/lex.js#L1424) for [a set of characters that includes both `\uFEFF` and `\uFFFE`](https://github.com/jshint/jshint/blob/12811e797a264f323b920b314db965b8adcb207e/src/reg.js#L13-L15). This is done unconditionally for all input, regardless of configuration options or source code input method (i.e. string or array).

As far as I can tell (and [preliminary tests confirm](https://github.com/jugglinmike/jshint/commit/a5614589aba9375113c225f08e59dfcd22b33609)), JSHint already warns about BOM sequences. This contradicts the assumption of [the original feature request](https://github.com/jshint/jshint/issues/2284), so I suspect I'm missing something important here. @caitp / @Krinkle: could you help me understand?

rwaldron:
@caitp @krinkle any follow up?

Krinkle:
I test, I created a file in Sublime Text 3 and used "Save with Encoding > UTF-8 with BOM" (verified with `cat test.js | wc` in bash that the file did in fact grew 3 bytes after this operation).

Passing it to jshint v2.8.0 didn't produce any warnings.

jugglinmike:
Based on the discussion so far, I would expect the test case that @caitp provided to fail without the patch. Since they do not ([as I mentioned above](https://github.com/jshint/jshint/pull/2285#issuecomment-89662440)), I think we are still missing some part of the problem. @Krinkle could you upload that file somewhere? That will help make the failing test case.

Krinkle:
@jugglinmike I've committed the file to git, and pushed here as pull #2478. It seems it is failing already actually, though it could certainly do with a more descriptive warning. 
>   [bold]{Line 1, Char 0} This character may get silently deleted by one or more browsers.

jugglinmike:
@Krinkle Your explicit test (along with your original report) has helped me to identify the issue. The unit test already fails on `master` not because we recently resolved it, but because the unit tests are run against the `JSHINT` function directly. [The behavior your originally reported is limited to the CLI](https://github.com/jshint/jshint/blob/5f88aa76eb1b4ec5acfba4a539e8ea33c965dc78/src/cli.js#L485):

```js
  buffer.push(code);
  buffer = buffer.join("\n");
  buffer = buffer.replace(/^\uFEFF/, ""); // Remove potential Unicode BOM.
```

There are not tests or documentation for this, but I've traced its introduction to [a pull request against the long-deprecated `node-jshint` project](https://github.com/jshint/node-jshint/pull/50).

Although the proposed `nobom` option could address your use-case, I think the real issue here is the inconsistency between the API and the CLI. If our confused, months-long conversation is any indication, the inconsistency causes more trouble than it is worth.

I think it would be better to remove the offending line from the CLI code. Users who aren't concerned about the issue can suppress the `W100` warning. I've opened gh-2486 to track that. This would be a breaking change, though, so we'll plan for the next major release.

Thanks for following through on this!

PR #2296, opened 2015-04-10
Extends overrides section of base config

arikon:
It was overwritten before

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "minimatch":           "2.0.x",
     "shelljs":             "0.3.x",
     "strip-json-comments": "1.0.x",
-    "underscore":          "1.8.x"
+    "lodash":              "3.6.x"
   },
 
   "devDependencies": {
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var _                 = require("underscore");
+var _                 = require("lodash");
 var fs                = require("fs");
 var cli               = require("cli");
 var path              = require("path");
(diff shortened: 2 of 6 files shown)


jugglinmike:
Thanks for the pull request! Could you add a test for this?

arikon:
@jugglinmike There are no tests for `extends` at all. Could you spot me, where to add it?

jugglinmike:
https://github.com/jshint/jshint/blob/master/tests/cli.js . That file [currently uses `extends` on lines 45 through 48](https://github.com/jshint/jshint/blob/12811e797a264f323b920b314db965b8adcb207e/tests/cli.js#L45-L48) for use in later tests, although a new test would probably be appropriate.

arikon:
@jugglinmike Done.

I've replaced `underscore` with `lodash` because:
- it is a bit faster
- it has `_.merge()` function for deep object extending (less code to extend configs and so to maintain)

And added two more tests.

package.json
```diff
@@ -41,7 +41,7 @@
     "minimatch":           "2.0.x",
     "shelljs":             "0.3.x",
     "strip-json-comments": "1.0.x",
-    "underscore":          "1.8.x"
+    "lodash":              "3.6.x"
```
rwaldron: :D

rwaldron:
https://github.com/jshint/jshint/commit/928f19dbf3378734ef2119689dc51acfa37edfc2

arikon:
@rwaldron :+1:

arikon:
@rwaldron @jugglinmike Could you please `npm publish` `jshint@2.7.0`

jugglinmike:
All set

PR #2298, opened 2015-04-11
Update dependencies.

XhmikosR:
Note that the new JSCS version fails for indented comments. From a quick look, I couldn't find a fix; any feedback is welcomed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,22 +38,22 @@
     "console-browserify":  "1.1.x",
     "exit":                "0.1.x",
     "htmlparser2":         "3.8.x",
+    "lodash":              "3.8.x",
     "minimatch":           "2.0.x",
-    "shelljs":             "0.3.x",
-    "strip-json-comments": "1.0.x",
-    "lodash":              "3.7.x"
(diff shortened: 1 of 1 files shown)


XhmikosR:
Hmmm, Travis is running a different command than AppVeyor.

@rwaldron @caitp: can we make this consistent? Also, if no one is using coveralls locally, we could skip the devDependency and just install it for CI.

XhmikosR:
@ mikesherov: isn't this a bug?  https://ci.appveyor.com/project/jshint/jshint/build/465/job/h5pk71jaj30476le

XhmikosR:
@rwaldron @caitp: I skipped JSCS for now so that everything passes. I still believe we should run the same command in Travis as AppVeyor (well, maybe except for coverage; no need to duplicate that).

XhmikosR:
So, does anybody have any idea how to fix [that JSCS error](https://ci.appveyor.com/project/jshint/jshint/build/465/job/h5pk71jaj30476le)? We can always silence JSCS for those lines as a last resort.

PR #2302, opened 2015-04-13
[[FEAT]] Add option 'brkret' to suppress warning about unreachable break...

### LINKED ISSUE #757: add option or comment hint to ignore 'Unreachable break after return'
I specifically ask about this condition because any straight-forward lex-compatible parser generator will use this pattern.

lukaszzdanikowski:
[[FEAT]] Add option 'brkret' to suppress warning about unreachable break after return statement

This options suppresses warnings about unreachable break after return. It is useful especially
with auto generated code.

This issue was already closed but with workaround 

Closes #757

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1529,9 +1529,13 @@ var JSHINT = (function() {
           }
           break;
         }
-        if (!state.option.brkret) {
-          warning("W027", t, t.value, controlToken.value);
+
+        if (t.value === "break" && controlToken.value === "return") {
+          if (state.option.brkret === true) {
+            return;
+          }
         }
+        warning("W027", t, t.value, controlToken.value);
         break;
       }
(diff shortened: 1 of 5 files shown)


lukeapage:
This option can be turned off globally using `W027`. The only reason to add a named option as I understand it, is if it is a common thing to want to turn off.
So the closed issue was not given a work-around, it was given the the way of turning off that option (the issue mentions the directive being function scoped - it is, but nothing stops you defining it globally or in your config)

Personally, I don't see this adds much value. Core team members may feel differently, but given how stale this is, I doubt it.

jugglinmike:
I agree--disabling `W027` is just as easy as enabling a new `brkret` option, but the former doesn't require any additional documentation or tests. Thanks anyway!

Issue #1636, opened 2014-04-22
ES6 Modules and strict option


DonutEspresso:
ES6 modules are strict by default. Is there a way to trigger strict parsing without actually having to output the 'use strict' statement within ES6 files? It's not clear to me if ESNext triggers that. Using globalstrict: true causes the parser to complain about missing use strict statements. My understanding is that ES6 assumes strict by default, and does not require the 'use strict' identifier. 

(Context: source code does not have 'use strict', but ES6 transpiler adds 'use strict'. I want to avoid having two 'use strict' statements in transpiled code.)

wuct:
+1

jugglinmike:
What are you +1'ing?

arcanis:
JSHint warns about missing strict mode declarations even when the file is an ES6 module. So we +1'ing this :

> This is correct. Detection should be trivial, any file with an export must be treated as a module.

kikonen:
So, has there been in progress on this. I.e. adding extra "use strict" into .es6 files is a bit clumsy workaround.

jugglinmike:
Many people have posted "+1", but there has been no progress that I am aware of. If anyone is interested in helping out, I will update the "Assignee" attribute on this issue to make that more clear

caitp:
so, there's a bit of a problem with this --- we don't know if we need to parse as a script or as a module.

yeah, we could sniff the file and see if it contains any module-specific syntax, but that doesn't help much (that's not how engines will behave, most likely, especially if we end up with the `<script module>` or whatever that has been proposed before. If a file containing no module syntax is still a module, then we'd be dishonest in doing this.

I think what it really needs, is like an extra setting in .jshintrc or something, to list files which are to be parsed as modules --- or to enable "parse as module by default". This would also allow us to report module syntax in scripts as syntax errors, I think. @rwaldron would know better, but otoh that seems right

rwaldron:
> I think what it really needs, is like an extra setting in .jshintrc or something, to list files which are to be parsed as modules --- or to enable "parse as module by default".

That's about the only unambiguous solution I've seen so far. I can support that. @jugglinmike ?

kikonen:
Detecting it by file extension surely sounds rather robust to me (I presume you are meaning that). Such file extension list could even default into something like [".es6'"] (unless something else specified by the .jshintrc). However, I think that "parse as module by default" option could have some trouble with projects containing both es5 and es6 files.

caitp:
[...]

PR #2306, opened 2015-04-13
[[FEAT]] Impement `module` option

jugglinmike:
@rwaldron @caitp I took a few liberties while implementing this. These aren't (*ahem*) "strictly" necessary, but they've always bugged me, and improving them now made the "module" option easier to implement. They seem pretty low-risk to me, but I've thought that before, so I'd appreciate your review of those changes in particular. (And if you don't think they're improvements, don't hesitate to say so.) They're contained in separate commits to help review (and so I can use commit messages to share justification).

I've introduced a new warning to help people avoid using the new option incorrectly. I've tried to make this generic enough for re-use with other ES6 options, and even for use in future versions.

Note that I've declared this under the "environments" collection of options. My first choice was "enforcing", but I changed my mind while working on the patch. Although it doesn't introduce new identifiers, it does describe an environmental change, so it kind of makes more sense to go here. It's not really "enforcing" *per se*, and people using `enforceall` would be especially surprised if the next release started treating all their code as strict mode code. (I can't wait to get rid of that option, by the way.)

Ideas for future extensions:

- Any minor release - Make this a "value" option and allow three different values: `true`, `false`, and `"infer"`. That last one would use a heuristic to automatically detect whether the code was module code. As Cait pointed out, this could never be foolproof, but by exposing the behavior through an explicitly-named option value, we can have *some* confidence that the user understands the edges before opting in. The benefit here would be that users wouldn't have to special-case their configuration according to their file structure.
- Any major release - Warn on `import` and `export` statements in any code that does not use `module: true` (or `module: "infer"` as the case may be).

If you think that first idea is worthwhile, I could implement it as part of this pull request. But the second one is backwards-breaking, so it will have to wait until 3.0.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -202,6 +202,22 @@ var JSHINT = (function() {
       combine(predefined, vars.ecmaIdentifiers[6]);
     }
 
+    if (state.option.module) {
+      /**
+       * TODO: Extend this restriction to *all* "environmental" options.
+       */
+      if (!hasParsedCode(funct)) {
+        error("E055", state.tokens.next, "module");
+      }
+
+      /**
+       * TODO: Extend this restriction to *all* ES6-specific options.
+       */
+      if (!state.inESNext()) {
+        warning("W134", state.tokens.next, "module", 6);
+      }
+    }
+
     if (state.option.couch) {
       combine(predefined, vars.couch);
     }
@@ -2954,6 +2970,17 @@ var JSHINT = (function() {
     return "(scope)" in token;
   }
 
+  /**
+   * Determine if the parser has begun parsing executable code.
+   *
+   * @param {Token} funct - The current "functor" token
+   *
+   * @returns {boolean}
+   */
+  function hasParsedCode(funct) {
+    return funct["(global)"] && !funct["(verb)"];
+  }
+
   function doTemplateLiteral(left) {
     // ASSERT: this.type === "(template)"
     // jshint validthis: true
@@ -5317,7 +5344,8 @@ var JSHINT = (function() {
 
         if (state.isStrict()) {
           if (!state.option.globalstrict) {
-            if (!(state.option.node || state.option.phantom || state.option.browserify)) {
+            if (!(state.option.module || state.option.node || state.option.phantom ||
+              state.option.browserify)) {
               warning("W097", state.tokens.prev);
             }
           }
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -148,7 +148,7 @@ var JSHINT = (function() {
       // Some ES5 FutureReservedWord identifiers are active only
       // within a strict mode environment.
       if (meta.strictOnly) {
-        if (!state.option.strict && !state.directive["use strict"]) {
+        if (!state.option.strict && !state.isStrict()) {
           return false;
         }
       }
@@ -1337,7 +1337,7 @@ var JSHINT = (function() {
         if (left.id === ".") {
           if (!left.left) {
             warning("E031", that);
-          } else if (left.left.value === "arguments" && !state.directive["use strict"]) {
+          } else if (left.left.value === "arguments" && !state.isStrict()) {
             warning("E031", that);
           }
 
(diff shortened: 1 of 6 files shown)


caitp:
The patch looks good.

I'm really not keen on the "infer" idea, I think this could lead to confusion. Explicit is better, imo.

Can you make sure `// jshint module: true` works properly (eg at the head of the file)?

rwaldron:
This is surprisingly simple, I like it. 

> I'm really not keen on the "infer" idea, I think this could lead to confusion. Explicit is better, imo.

I'm with @caitp on this. 


> Any major release - Warn on import and export statements in any code that does not use module: true (or module: "infer" as the case may be).

Very cool.


------

Things...

- @caitp asked for confirmation of `// jshint module: true` inline/in-file, which I just tried and it works, but please add a test for this. 

- When I add `"module": true`  to a .jshintrc, all files will be treated as a module, which is what I would expect. 
- What about allowing (in some future extension of the feature)...
  - `"module": [array of file name strings]`
  - `"module": "glob expression"`
  - `"module": [array of file name strings or glob expressions]`
- Maybe that ^^^ is an option called `"modules"` (with an "s")?

jugglinmike:
> What about allowing (in some future extension of the feature)...
>
> - "module": [array of file name strings]
> - "module": "glob expression"
> - "module": [array of file name strings or glob expressions]
>
> Maybe that ^^^ is an option called "modules" (with an "s")?

It sounds like you have gh-1636 in mind.

The good news is that JSHint already exposes a general mechanism for this--`overrides`. (I only learned about this through the recent gh-2296, but there is [an open issue to document it](https://github.com/jshint/jshint.github.io/issues/17).) This means that we can land this as-is without including additional filesystem code :)

jugglinmike:
Thanks for the review! As requested, I've added an additional test for in-line directive. This raises another question: should we forbid use of this directive in function scope? It doesn't make much sense to write:

```js
class Foo {
  bar() {
    /* jshint module: true */
  }
}
```

...but that's possible with this patch in its current state. Is it worth adding code to disallow weird uses like that? Is it likely that doing so would help anyone?

caitp:
> This raises another question: should we forbid use of this directive in function scope? It doesn't make much sense to write:

I was saying this above --- I think the module directive is kind of special, in that you really need it to be one of the first items in the file (before any non-comment tokens)

jugglinmike:
> I was saying this above

Sorry, I completely missed that. I'll see what I can do about this tonight. I think this would generally be fitting for *all* "environment" options. Obviously we can't do that in a minor release, but I'll try to build it so we can restrict `node: true` (for example) this way later.

jugglinmike:
Note: we just received a feature request to turn `devel` off on a line-by-line basis (in the #jshint IRC channel), so this idea:

> I think this would generally be fitting for all "environment" options.

May not be appropriate after all.

caitp:
I think you need to distinguish between certain methods being "safe" or " allowed", and an operating environment like "this is a module or script, browser or node, web worker or what" etc. Allowing reference to certain globals could certainly be enabled on a line by line basis, but nothing will change the intended environment of a file

jugglinmike:
Okay, as discussed, the option is now disallowed outside of ES6 code and after executable code has been parsed. Sorry for the delay.

jugglinmike:
Rebased! I haven't received any feedback on the additional restrictions added by the two `fixup!`- commits. If those look good, I'll squash an merge.

src/jshint.js
```diff
@@ -855,7 +844,7 @@ var JSHINT = (function() {
 
     // if current expression is a let expression
     if (!initial && state.tokens.next.value === "let" && peek(0).value === "(") {
-      if (!state.option.inMoz(true)) {
+      if (!state.inMoz(true)) {
```
rwaldron: Since all occurrences have to be touched in this patch, maybe also refactor this Boolean arg into an options object?

rwaldron:
LGTM, I'll let you make the call regarding my comment about refactoring the Boolean arg.

jugglinmike:
Nice call, @rwaldron . In some cases, the parameter was completely unnecessary!

rwaldron:
Oh, that's great news

jugglinmike:
Triggering one last test build against the squashed history...

jugglinmike:
Thanks, Cait! Thanks, Rick!

caitp:
You did it!

rwaldron:
Nice work!

impinball:
Yay!

PR #2316, opened 2015-04-16
[[DOCS]] Added an example for the globals option

countermeasure:
The `globals` option requires an object, but this doesn't appear to be documented.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -760,7 +760,11 @@ exports.val = {
    * Setting an entry to `true` enables reading and writing to that variable.
    * Setting it to `false` will trigger JSHint to consider that variable
    * read-only.
-   *
+   * 
+   * This option accepts an object, like so:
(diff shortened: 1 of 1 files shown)


src/options.js
```diff
@@ -760,7 +760,11 @@ exports.val = {
    * Setting an entry to `true` enables reading and writing to that variable.
    * Setting it to `false` will trigger JSHint to consider that variable
    * read-only.
-   *
+   * 
+   * This option accepts an object, like so:
+   * 
+   *   // jshint globals:{"React": false, "_": false}
```
lukeapage: this code doesn't work.. you need this to set the option
```
/* globals a:false */
```

In addition this is documentation inside jshint for the option, which can be set inline, or through config, so I am not sure this is the appropriate place to mention the format of the inline form

PR #2318, opened 2015-04-20
Separate package.json tasks.

XhmikosR:
Also, remove JSHint devDependency; use `bin/jshint` directly, instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,9 @@
     "build": "node bin/build",
     "coverage": "istanbul -- cover ./node_modules/.bin/nodeunit tests/unit",
     "data": "node scripts/generate-identifier-data",
-    "pretest": "jshint src && jscs src",
+    "jscs": "jscs src",
+    "jshint": "node bin/jshint src",
+    "pretest": "npm run jshint && npm run jscs",
     "test": "nodeunit tests tests/regression tests/unit"
   },
 
@@ -49,7 +51,6 @@
     "coveralls":     "2.11.x",
     "istanbul":      "0.3.x",
     "jscs":          "1.11.x",
-    "jshint":        "2.6.x",
     "mock-stdin":    "0.3.x",
     "nodeunit":      "0.9.x",
     "regenerate":    "1.2.x",


package.json
```diff
@@ -27,7 +27,9 @@
     "build": "node bin/build",
     "coverage": "istanbul -- cover ./node_modules/.bin/nodeunit tests/unit",
     "data": "node scripts/generate-identifier-data",
-    "pretest": "jshint src && jscs src",
+    "jscs": "jscs src",
+    "jshint": "node bin/jshint src",
```
lukeapage: I'm not sure about this.. when developing you use jshint as a tool to check its valid javascript. if it is *not* valid, then this will run your invalid code, error, and you won't get the benefit of jshint telling you where you typed something wrong...

lukeapage: Also, why do this?

XhmikosR: Please check the whole branch and my other PRs. You've missed the point.

XhmikosR: Again you clearly miss the point. It's at least stupid to have jshint as a
dependency for its own code. Instead, we use it directly.

lukeapage: I've glanced through the other PR's from you and I cannot see a connection.

> It's at least stupid to have jshint as a
> dependency for its own code.

You cannot just say something is stupid and a justification. I've provided an example where the current setup is *more* useful than your proposed setup. Why is your proposed setup more useful to people developing jshint?

> Again you clearly miss the point.

Sorry if I have, I'm just trying to sort out all the PR's, understand and review them. Please either explain within the PR, link to an explanation or you will have to rely on someone spending the time to try and understand what you meant, which is unlikely if you are not helpful and no benefit is obvious.

XhmikosR: Sorry, if you cannot see the benefit from removing jshint and using
directly, just leave it here.

Anyway, I've lost any interest in JSHint's development in any way and I
have no interest nor time to explain, obvious to me, things.

Consider my PRs closed and I'll close them later myself if no one else does
it.

lukeapage: > Anyway, I've lost any interest in JSHint's development in any way and I
> have no interest nor time to explain, obvious to me, things.
> 
> Consider my PRs closed and I'll close them later myself if no one else does
> it.

Sorry you were discouraged, I'm currently trying to sort out the PR's so people do not get discouraged in the future.

Your other PR's, marked `Needs review` I am not discounting, it was just this one I was questioning to try and understand the motive behind it.

Issue #2292, opened 2015-04-07
Incorrect variable warnings based on order of `let` declarations 


ratbeard:
I noticed this in my app.  It is reproducible on http://jshint.com/ with the esnext option checked:

```js
let a = 1;
let b = () => a;
```
is ok

---

```js
let b = () => a;
var a = 1;
```
is ok

---

```
let b = () => a;
let a = 1;
```
Has Error: 

```
One undefined variable
1	a
```

rwaldron:
Confirmed

lukeapage:
I'm working on this one.

latedef and the unused implications of latedef is broken for block scoped variables (its not implemented and the function scoped implementation is function scoped...)

lukeapage:
@rwaldron is this actually a bug?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

it seems that let and const do not get hoisted to the top of the block they are in, but generate a reference error in strict mode and a global without.

caitp:
@lukepage, it's not exactly that the variables do not get hoisted to the top of the block --- the variables are created but uninitialized, and it's an error to access them before they're initialized (the "temporal dead zone").

So the following is perfectly legal:

```js
let f = () => a;
let a = 1; // or just `let a;`
f();
```

The following is not legal:

```js
let f = () => a;
f();
let a = 1; // or `let a;`
```

caitp:
also note, there's a limit to how much TDZ handling we can do, because it's not always possible to know statically that a TDZ violation occurred. It's a runtime check (although currently in V8, we get rid of the runtime check if the reference occurs after the declaration, but I'm pretty sure that causes some bugs)

lukeapage:
Aha,  makes sense, thanks.

lukeapage:
@caitp any idea if this example from the jshint tests is valid? I've looked through the spec, some examples and tried babeljs.io out and it looks invalid, but I could be wrong..

```
export {
  varDefinedLater,
  letDefinedLater,
  constDefinedLater
};
var unusedGlobalVar = 41;
let unusedGlobalLet = 41;
const unusedGlobalConst = 41;
```

lukeapage:
aha seems you agree, found https://github.com/jshint/jshint/pull/1891#discussion_r18475719

Issue #2271, opened 2015-03-23
ES6 Constant Re-assignment


jclem:
This is valid according to JSHint with "esnext" on:

```javascript
const x = 0;
x++;
```

It should instead warn "Attempting to override 'x' which is a constant.", like the following does:

```javascript
const x = 0;
x += 1;
```

rwaldron:
Agreed!

Issue #2268, opened 2015-03-20
[ES6] const not treated as block-scoped


AgentME:
The following gives no warnings, as expected:

```javascript
// jshint esnext:true
{let a = 5;}
{let a = 5;}
```

But this does, even though const is also block-scoped:

```javascript
// jshint esnext:true
{const a = 5;}
{const a = 5;}
```

```
stdin: line 3, col 10, const 'a' has already been declared.
```

jshint v2.6.3

lukeapage:
I worked on this yesterday and hope to have a PR in the next 24 hours (to save anyone else making a similiar effort)

lukeapage:
duplicate of #1468

lukeapage:
@jugglinmike this can be closed now

jugglinmike:
Confirmed--resolved via gh-2326 / 1bb80f9a5551a639db2be0dea655c9edef7a9b8a . Expect this in the next release of JSHint (version > 2.7.0).

Issue #2090, opened 2015-01-09
Shadowing issue errorneuously reported as redefinition when using block scoping


fluidsonic:
```js
const Set = require('some-set-package');  // Redefinition of 'Set'
```

Since `let` and `const` are block scoped the global variable will be left untouched. 
This looks like a bug as this is shadowing, not redefinition.

I actually stumbled across this problem when overwriting `Set` (as node's implementation is very limited) and we can expect more such warnings in the future as the number of predefined global symbols increases.

(jshint 2.5.11)

lukeapage:
@jugglinmike this can probably be closed I think, unless @fluidsonic can provide a proper test case which doesn't work on master, but I think it should.

jugglinmike:
This is still an issue on `master`

`gh-2090.js`:

```js
// jshint esnext: true
{
  var Proxy = null; // expected error
  const Set = null;
  let Map = null;
}
(function() {
  var Proxy = null;
  const Set = null;
  let Map = null;
}());
```

    gh-2090.js: line 3, col 7, Redefinition of 'Proxy'.
    gh-2090.js: line 4, col 9, Redefinition of 'Set'.
    gh-2090.js: line 5, col 7, Redefinition of 'Map'.

PR #2326, opened 2015-04-20
[[Fix]] Make const have block scope

lukeapage:
I've re-combined the copy and pasted code for var/const/let and in the process made const block scoped as per the spec.

It was a bit of a nightmare and the code was a bit of a nightmare.

I'd like to see if there are any bits to tidy and add some more tests, but I wanted to start getting feedback first - no point polishing if you are unhappy with the general approach.

More things to do:

* I've removed the hacked-in way to allow unused variable tracking of consts and replaced it with a new system (again just for consts) - would it be nice if this was the same for all variables?
* I added a todo about the warnunsed method not being used - it was already like this, but I'll see if it makes sense to fix or take out
* more tests

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4887,6 +4887,40 @@ var JSHINT = (function() {
     }
   }
 
+  var warnUnused = function(name, tkn, type, unused_opt) {
+    var line = tkn.line;
+    var chr  = tkn.from;
+    var raw_name = tkn.raw_text || name;
+
+    if (unused_opt === undefined) {
+      unused_opt = state.option.unused;
+    }
+
+    if (unused_opt === true) {
+      unused_opt = "last-param";
+    }
+
+    var warnable_types = {
+      "vars": ["var"],
+      "last-param": ["var", "param"],
+      "strict": ["var", "param", "last-param"]
+    };
+
+    if (unused_opt) {
+      if (warnable_types[unused_opt] && warnable_types[unused_opt].indexOf(type) !== -1) {
+        if (!tkn.exported) {
+          warningAt("W098", line, chr, raw_name);
+        }
+      }
+    }
+
+    unuseds.push({
+      name: name,
+      line: line,
+      character: chr
+    });
+  };
+
   var blockScope = function() {
     var _current = {};
     var _variables = [_current];
@@ -4902,16 +4936,8 @@ var JSHINT = (function() {
             if (tkn.exported) {
               continue;
             }
-            var line = tkn.line;
-            var chr  = tkn.character;
-            warningAt("W098", line, chr, t);
 
-            // todo - use warnunused?
-            unuseds.push({
-              name: t,
-              line: line,
-              character: chr
-            });
+            warnUnused(t, tkn, "var");
           }
         }
       }
@@ -5258,40 +5284,6 @@ var JSHINT = (function() {
           delete implied[name];
         else
           implied[name] = newImplied;
-      };
-
-      var warnUnused = function(name, tkn, type, unused_opt) {
-        var line = tkn.line;
-        var chr  = tkn.from;
-        var raw_name = tkn.raw_text || name;
-
-        if (unused_opt === undefined) {
-          unused_opt = state.option.unused;
-        }
-
-        if (unused_opt === true) {
-          unused_opt = "last-param";
-        }
-
-        var warnable_types = {
-          "vars": ["var"],
-          "last-param": ["var", "param"],
-          "strict": ["var", "param", "last-param"]
-        };
-
-        if (unused_opt) {
-          if (warnable_types[unused_opt] && warnable_types[unused_opt].indexOf(type) !== -1) {
-            if (!tkn.exported) {
-              warningAt("W098", line, chr, raw_name);
-            }
-          }
-        }
-
-        unuseds.push({
-          name: name,
-          line: line,
-          character: chr
-        });
       };
 
       var checkUnused = function(func, key) {
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -463,7 +463,7 @@ var JSHINT = (function() {
       }
     }
 
-    // if the identifier is from a let, adds it only to the current blockscope
+    // if the identifier is blockscoped (a let or a const), add it only to the current blockscope
     if (isblockscoped) {
       funct["(blockscope)"].current.add(name, type, state.tokens.curr);
       if (funct["(blockscope)"].atTop() && exported[name]) {
@@ -1092,9 +1092,11 @@ var JSHINT = (function() {
             this.right.id !== "." && this.right.id !== "[") {
           warning("W017", this);
         }
+
+        // detect modification of a const
+        // in the case of a.b, right will be the "." punctuator
         if (this.right && this.right.identifier) {
-          var label = funct["(blockscope)"].getlabel(this.right.value);
-          if (label && label[this.right.value]["(type)"] === "const") {
+          if (funct["(blockscope)"].labeltype(this.right.value) === "const") {
             error("E013", this, this.right.value);
           }
         }
@@ -1328,8 +1330,7 @@ var JSHINT = (function() {
           warning("W021", left, left.value);
         }
 
-        var label = funct["(blockscope)"].getlabel(left.value);
-        if (label && label[left.value]["(type)"] === "const") {
+        if (funct["(blockscope)"].labeltype(left.value) === "const") {
           error("E013", left, left.value);
         }
 
@@ -1424,14 +1425,19 @@ var JSHINT = (function() {
     var x = symbol(s, 150);
 
     x.led = function(left) {
+      // this = suffix e.g. "++" punctuator
+      // left = symbol operated e.g. "a" identifier or "a.b" punctuator
+
       if (state.option.plusplus) {
         warning("W016", this, this.id);
       } else if ((!left.identifier || isReserved(left)) && left.id !== "." && left.id !== "[") {
         warning("W017", this);
       }
+
+      // detect modification of a const.
+      // in the case of a.b, left will be the "." punctuator
       if (left && left.identifier) {
-        var label = funct["(blockscope)"].getlabel(left.value);
-        if (label && label[left.value]["(type)"] === "const") {
+        if (funct["(blockscope)"].labeltype(left.value) === "const") {
           error("E013", this, left.value);
         }
       }
@@ -1962,6 +1968,8 @@ var JSHINT = (function() {
           this["function"] = true;
           break;
         case "const":
+          // consts can only exist inside block
+          // because they are never added to the scope, s
           block[v]["(unused)"] = false;
           break;
         case "function":
@@ -3409,16 +3417,16 @@ var JSHINT = (function() {
     });
   }
 
-  function variableStatement(type, statement, context) {
+  function blockVariableStatement(type, statement, context) {
+    // used for both let and const statements
+
     var prefix = context && context.prefix;
     var inexport = context && context.inexport;
     var isLet = type === "let";
     var isConst = type === "const";
-    var isVar = type === "var";
-    var isBlockScoped = isLet || isConst;
-    var tokens, lone, value, letblock, existingLabel;
+    var tokens, lone, value, letblock;
 
-    if (isBlockScoped && !state.option.inESNext()) {
+    if (!state.option.inESNext()) {
       warning("W104", state.tokens.curr, type);
     }
 
@@ -3429,8 +3437,8 @@ var JSHINT = (function() {
       advance("(");
       funct["(blockscope)"].stack();
       letblock = true;
-    } else if (isBlockScoped && funct["(noblockscopedvar)"]) {
-      error("E048", state.tokens.curr);
+    } else if (funct["(noblockscopedvar)"]) {
+      error("E048", state.tokens.curr, isConst ? "Const" : "Let");
     }
 
     statement.first = [];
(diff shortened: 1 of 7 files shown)


rwaldron:
Wow, nice work.

lukeapage:
@rwaldron I've been looking at fixing #2292 and uncovered a few problems around const and let, so I am thinking of waiting on simplifying the way in which things can be marked as unused until after those changes. 

So this PR is ready for review and (hopefully) merging.

lukeapage:
bugs I think will be fixed by this:
  #2271 #2268 #1884 
maybe #2090 but its lacking info

src/jshint.js
```diff
@@ -3511,21 +3450,27 @@ var JSHINT = (function() {
...
                 warning("W129", t.token, t.id);
               }
             }
           }
-          if (t.id) {
-            addlabel(t.id, { type: "unused", token: t.token });
+          if (t.id && (!isBlockScoped || !funct["(noblockscopedvar)"])) {
+            addlabel(t.id, { type: isConst ? "const" : "unused",
+              token: t.token,
+              isblockscoped: isBlockScoped }); //todo check in here for else block scoped const specofic
```
lukeapage: need to remove todo, its old

rwaldron:
As soon as @jugglinmike or @caitp review and give it the go ahead, it can land :) 

Thanks for putting in the work to address these issues!

src/jshint.js
```diff
@@ -1094,6 +1092,12 @@ var JSHINT = (function() {
             this.right.id !== "." && this.right.id !== "[") {
           warning("W017", this);
         }
+        if (this.right && this.right.identifier) {
+          var label = funct["(blockscope)"].getlabel(this.right.value);
+          if (label && label[this.right.value]["(type)"] === "const") {
+            error("E013", this, this.right.value);
```
caitp: Add a comment: "no incrementing or decrementing const values" or something

src/jshint.js
```diff
@@ -1424,6 +1429,12 @@ var JSHINT = (function() {
       } else if ((!left.identifier || isReserved(left)) && left.id !== "." && left.id !== "[") {
         warning("W017", this);
       }
+      if (left && left.identifier) {
```
caitp: This is great, but I think you want to look at what the `this` token is --- it's cool to reference properties of a const value, and even change them, but you can't  use a postfix increment/decrement (which I think is what you're trying to prevent here).

Also like above, add a comment explaining what the code is accomplishing, it helps a lot =)

src/jshint.js
```diff
@@ -1951,7 +1962,7 @@ var JSHINT = (function() {
           this["function"] = true;
           break;
         case "const":
-          setprop(funct, v, { unused: false });
+          block[v]["(unused)"] = false;
```
caitp: Document the contract here of why you know `block` is not null/undefined here (to make life easier later)

src/jshint.js
```diff
@@ -2015,7 +2026,7 @@ var JSHINT = (function() {
               funct[v] = s["(global)"] ? "global" : "outer";
               break;
             case "const":
-              setprop(s, v, { unused: false });
+              s[v]["(type)"] = false;
```
caitp: the code being replaced looks related to legacy const stuff, not sure. In general though, I wouldn't expect you to need to do anything to blockscope `const` where similar work wasn't already done for `let`. The two should be pretty consistent, just slightly different semantics for `const`.

src/jshint.js
```diff
@@ -3413,88 +3409,31 @@ var JSHINT = (function() {
     });
   }
 
-  var conststatement = stmt("const", function(context) {
+  function variableStatement(type, statement, context) {
```
caitp: Don't do this --- it's going to painfully bitrot some code of mine, and there are big differences between lexical declarations and var declarations. In particular, binding resolution for var statements is different from lexical declarations, so lets keep them separate.

caitp: sidenote, it also makes debugging a bit more painful (having to use conditional breakpoints is always annoying when it can be avoided)

tests/unit/options.js
```diff
@@ -1413,7 +1416,16 @@ exports.esnext = function (test) {
     'const foo = 9;',
     'var myConst = function (test) { };',
     'foo = "hello world";',
-    'var a = { get x() {} };'
+    'var a = { get x() {} };',
+    'if (a) {',
+    '  const bar = 3;',
+    '  bar++;',
+    '  bar+=1;',
```
caitp: Add some tests with different binary operations, eg

```js
const f = () => 1;
bar();
const a = [1, 2, 3];
a[0] = 2;
const o = { a: 1 };
o.a = 2;
```

things like that (all of these are legal)

tests/unit/parser.js
```diff
@@ -3795,20 +3795,45 @@ exports["no let not directly within a block"] = function (test) {
...
-    .addError(1, "Let declaration not directly within block.")
-    .addError(4, "Let declaration not directly within block.")
-    .addError(7, "Let declaration not directly within block.")
-    .addError(8, "Let declaration not directly within block.")
-    .addError(9, "Let declaration not directly within block.")
-    .addError(10, "Let declaration not directly within block.")
-    .addError(11, "Let declaration not directly within block.")
-    .addError(11, "Let declaration not directly within block.")
-    .addError(11, "Let declaration not directly within block.")
+    .addError(1, "Block scoped variable declaration not directly within block.")
```
caitp: Lets do this better --- Make the "let" or "const" mode a variable in the message, and pick the right one to display. Makes it more readable

caitp:
So, I'd really appreciate it if the variablestatement stuff was changed a bit --- keep the lexical declarations together, but I'd really appreciate it if `var` was kept separate for its distinct binding behaviour, and easier debugging.

caitp:
Otherwise LGTM --- deferring to @jugglinmike for the rest

src/jshint.js
```diff
@@ -464,7 +464,7 @@ var JSHINT = (function() {
     }
 
     // if the identifier is from a let, adds it only to the current blockscope
```
jugglinmike: Could you update this comment as well?

src/jshint.js
```diff
@@ -5071,12 +4993,16 @@ var JSHINT = (function() {
       },
 
       current: {
+        getlabel: function(t) {
```
jugglinmike: I like seeing more code organized into the "blockScope" object, but I think this method might be too powerful. As defined, the previously-internal "label" objects are exposed, and we'll have to be vigilant about how future patches handle those. For your purposes here, though, you don't need access to the whole object, and you don't need to modify it, either. A more special-purpose method would be safer and would also reduce the repetition in the places where you use it--something like `labelType(t)` that returns the string value (or `null` when not present).

src/jshint.js
lukeapage: thanks, I knew that but hadn't thought to test it. In this case `left` will not be an identifier but a punctuator. `this` will be the punctuator `++`.
Adding tests and comments...

src/jshint.js
lukeapage: I went for:
```
          // consts can only exist inside block
          // because they are never added to the scope, s
```

src/jshint.js
lukeapage: const is required to have a type set to const in order to check whether the label has been modified.. var/let use a type of "unused" which becomes "var" once used. function begins as "unction" and becomes "function" when used.

I could have started "const" as "onst" and then once used promoted to "const" but then I would also have to look for type "const" || "onst" in a few places which I didn't like.

Personally I'd prefer to migrate var,let and function to have types "var","let","function" and use a boolean - what do you think? I was going to do that as part of the PR I am working on after this.. but whatever you feel is more maintainable?

a) leave as is
b) use "onst" for unused "const"
c) move unused and unction to this way in this PR
d) ""                         ""                              in a later PR
e) something better you can think of?

src/jshint.js
lukeapage: I'll put it back- it was just in the past it was a bit of a nightmare comparing let and const and var in the case of a bug - I can see how it has some advantages but it also makes the code prone to start to differ in ways they shouldn't.

Can I use the same function for let/const and seperate var or you'd prefer all 3 seperate?

src/jshint.js
jugglinmike: I've been trying to resolve this myself. While the current approach isn't really pretty, fixing it will require a more wide-spanning change that should be done independently (and preferably after this patch lands).

My plans are slightly different than your option "d", though. I think we should use a single label type, "binding" for all of these, then introduce two new attributes: "bindingType" (one of "var", "function", "let", or "const") and "used" (true or false). But if @caitp agrees with the order of operations, we can discuss that more in a follow-up issue.

src/jshint.js
lukeapage: > but I think this method might be too powerful. As defined, the previously-internal "label" objects are exposed, and we'll have to be vigilant about how future patches handle those.

If you look above, you'll see that there is another `getlabel` function already that returns a label object and there is code in multiple places calling it and also changing the returned object.
The only difference with the new function is it only looks in the current scope, where as the existing function looks in all blockscopes 

(see this code)
```
block = funct["(blockscope)"].getlabel(v);
```

but you are right that a labeltype would reduce the repetition - I will change this to labeltype and add a labeltype in the non-`current` context and use that as well.

src/jshint.js
caitp: I'm fine with using the same function for let/const

caitp: If @jugglinmike is alright with your approach, doing it all in one function is fine too --- but in my experience that leads to more problems

jugglinmike: I'm used to using `debugger;` statements, so the annoyance of conditional breakpoints doesn't bother me too much. I just wish there was a better name than "variable", since it heavily suggests `var` and is somewhat contradictory with "constant". A function like `lexicalDeclaration` would avoid this, but I'm with @lukeapage w/r/t drift. Let's run with this approach for now.

src/jshint.js
jugglinmike: Looks like I don't know the codebase as well as I thought. Thanks for reminding me

lukeapage:
Thanks for the code review, it was good. I think I have addressed everything.

src/jshint.js
```diff
@@ -3483,13 +3508,27 @@ var JSHINT = (function() {
...
+  var conststatement = stmt("const", function(context) {
+    return blockVariableStatement("const", this, context);
   });
 
+  var letstatement = stmt("let", function(context) {
+    return blockVariableStatement("let", this, context);
+  });
+  letstatement.exps = true;
+
   conststatement.exps = true;
```
lukeapage: oops wrong place

src/jshint.js
lukeapage: > Let's run with this approach for now.

Which one? I've already applied @caitp suggestion..  So keep as lexical / variable or combine?

lukeapage:
0.10 travis build failed posting to coveralls - can you restart?

src/jshint.js
```diff
@@ -1419,12 +1425,23 @@ var JSHINT = (function() {
...
+      // left = symbol operated e.g. "a" identifier or "a.b" punctuator
+
       if (state.option.plusplus) {
         warning("W016", this, this.id);
       } else if ((!left.identifier || isReserved(left)) && left.id !== "." && left.id !== "[") {
         warning("W017", this);
       }
 
+      // detect modification of a const.
+      // in the case of a.b, left will be the "." punctuator
```
caitp: this comment is misleading. list the ops you want to protect against

src/jshint.js
```diff
@@ -3440,14 +3458,21 @@ var JSHINT = (function() {
...
+            if (predefined[t.id] === false) {
+              warning("W079", t.token, t.id);
+            }
           }
-          if (t.id) {
-            addlabel(t.id, { token: t.token, type: "const", unused: true });
+          if (t.id && !funct["(noblockscopedvar)"]) {
+            addlabel(t.id, { type: isConst ? "const" : "unused",
+              token: t.token,
+              isblockscoped: true });
```
caitp: this object literal looks weird, i think the jscs rules need a tune up

src/jshint.js
lukeapage: its in a function that is only called for ++ and -- so what should I say?
I've changed for ```// detect increment/decrement of a const``` ?

src/jshint.js
lukeapage: I moved `type:` onto the next line, assume thats what you mean.

src/jshint.js
caitp: It’s called for a number of prefixes, preinc, predec, unary minus, unary plus. The unary ops are totally fine because they’re not assignment ops.

lukeapage: > It’s called for a number of prefixes, preinc, predec, unary minus, unary plus. The unary ops are totally fine because they’re not assignment ops.

you put this comment on the postfix function, which is only called for ++ and --

The prefix function this is the case, but in that function my block is clearly in the `if(this.id === "--" || this.id === "++")` section. anyway, as I said I've changed the comment anyway in both places, is that what you want?

caitp: You deleted the comment on github (or it otherwise vanished for some other reason), so I wasn’t able to reply on github and can’t see what you’re talking about. Anyways, the point is pretty clear: always document the intentions of your changes. The pratt parser is a mess, and not documenting changes is a real problem for maintainability. Just make sure you’re always crystal clear about the inputs you’re expecting and why

tests/unit/core.js
```diff
@@ -742,6 +742,62 @@ exports.testES6ModulesNamedExportsAffectUnused = function (test) {
...
+    "a++;",
+    "b.a++;",
+    "--a;",
+    "--b.a;",
+    "a += 1;",
+    "a.b += 1;",
+    "const c = () => 1;",
+    "c();",
+    "const d = [1, 2, 3];",
+    "d[0] = 2;",
```
caitp: add some other operators here --- we should really cover them all.

```js
let x = -<const>;
x = +<const>;
x = <const> + 1;
x = <const> * 2;
x = <const> / 2;
x = <const> % 2;
x = <const> & 1;
x = <const> ^ 1;
```

etc etc

src/jshint.js
lukeapage: > You deleted the comment on github (or it otherwise vanished for some other reason), so I wasn’t able to reply on github and can’t see what you’re talking about.

when a line is updated in a pr, github hides the comment.

You can see the comments by finding the "caitp commented on an outdated diff" line and click "Show outdated diff".

tests/unit/core.js
lukeapage: > add some other operators here --- we should really cover them all.

I've added everything I can think of

src/jshint.js
caitp: This might be shocking, but I have been active on github for a number of years, and I'm aware of this. `$('.discussion-item-toggle-open').click()` didn't find your comment

lukeapage: > This might be shocking, but I have been active on github for a number of years

yep, me too

> I'm aware of this.

you never know! sorry, I didn't mean to offend.



> $('.discussion-item-toggle-open').click() didn't find your comment

good tip, bookmarklet added.



> Anyways, you've been asked to add some tests and comment your stuff clearly, I think that's the only thing blocking landing this

yep both those things are already done

jugglinmike:
Squashed and merged to `master` at commit 1bb80f9a5551a639db2be0dea655c9edef7a9b8a . This was a big one; thanks for your hard work, @lukeapage !

PR #2328, opened 2015-04-21
Lint everything

XhmikosR:
This is to be merged after #2087,  #2318 and #2321.

Note, that we could have JSCS check tests too; the only issue is `maxlen`.

https://github.com/jshint/jshint/pull/2328/files?w=1 without the whitespace changes.


lukeapage:
Marking needs discussions because I expect @jugglinmike is opposed for the same reasons as - https://github.com/jshint/jshint/pull/2087#issuecomment-99251935

It also, I think makes sense to do this when the list of PR's is more clear, so I would put the priority of this towards the bottom.

Issue #2324, opened 2015-04-20
Jshint with esnext fails on file


kharandziuk:
If I use jshint for file like http://pastebin.com/MQRH5nHb it newer completes

```
node /usr/local/bin/jshint -v
jshint v2.7.0
```
my .jshintrc:
```
{
    "esnext": "true"
}
```

jugglinmike:
Thanks for the report!

I'm inlining the input to keep things organized:

```js
var React = require('react');
var {DOM} = React;
 
 
class App extends React.Component {
    constructor(props) {
      super(props);
    }
    render() {
      return DOM.div(
          {className: 'container'},
          DOM.div({className: 'col-lg-offset-3'}, [
            DOM.form([
              DOM.div(
                {
                  className: 'form-inline',
                  children:  [
                  DOM.div({className: 'form-group'}, [
                    DOM.label('First Name'),
                    DOM.input({className: "form-contol"})
                  ]),
                ]
              ),
            ]),
            //DOM.table({className: 'table'})
          ])
      );
    }
}
 
React.render(
    React.createElement(App),
    document.body
);
```
```

**Comment by jugglinmike** [comment]
Minimal test case:

```js
class a { b
```

I've submitted a patch for this at gh-2338.

lselden:
If it helps, here's another example that can trigger the failure.

    class Foo {
      constructor (config = {}) {
        config =
      }
    }

I found that I was able to avoid problems with incomplete class definitions by adding a "(end)" token guard in the classbody function (around line 3761):

      if (!checkPunctuators(state.tokens.next, ["("])) {
        // error --- class properties must be methods
        error("E054", state.tokens.next, state.tokens.next.value);
        while (state.tokens.next.id !== "}" &&
               !checkPunctuators(state.tokens.next, ["("])
               && state.tokens.next.id !== '(end)') { // THIS IS THE LINE I ADDED
          advance();
        }
        if (state.tokens.next.value !== "(") {
          doFunction({ statement: c });
        }
      }

PR #2338, opened 2015-04-23
[[FIX]] Allow lexer to communicate completion

jugglinmike:
I'm not so gung-ho about modifying assertions, but I believe it's an acceptable change. Note that all test cases still pass (JSHint is still failing when it is expected to), but the "shape" of the reported errors is a little different. I would argue the new set of errors are more accurate in some cases, but that's pretty subjective. (See also [this conversation](https://github.com/jshint/jshint/pull/1989#issuecomment-63559108) for the general problem of hard-coding expected errors for invalid source code.) Resolves gh-2324

Commit message:

> Currently, JSHint's lexer continuously emits new "end-of-file" token
> once all input has been consumed. This behavior makes the internal
> `advance` function dangerous to use within iteration contexts. Such
> usages need to explicitly check for the "end-of-file" token in order to
> avoid entering non- terminant states.
>
> Some of JSHint's parsing logic (such as that for template literal
> parsing) takes this detail into account. Other areas (i.e. parsing class
> bodies and function parameter lists) does not. While it would be
> possible to extend each use with explicit safety checks, the possibility
> for error in future feature additions would remain.
>
> Update the lexer to consistently emit `null` after producing the
> "end-of-file" token, alleviating the requirement of explicit safety
> checks in `advance`-calling code. Update existing unit tests with the
> new parsing error information (which is more accurate in many cases).

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -30,7 +30,6 @@ exports.other = function (test) {
   TestRun(test)
     .addError(1, "Unexpected '\\'.")
     .addError(2, "Unexpected early end of program.")
-    .addError(2, "Expected an identifier and instead saw '(end)'.")
     .addError(2, "Unrecoverable syntax error. (100% scanned).")
     .test(code, {es3: true});
 
@@ -339,8 +338,7 @@ exports.numbers = function (test) {
     .addError(16, "Missing semicolon.")
     .addError(17, "Unexpected '1'.")
     .addError(17, "Unexpected early end of program.")
-    .addError(17, "Expected an identifier and instead saw '(end)'.")
-    .addError(17, "Missing semicolon.")
+    .addError(17, "Unrecoverable syntax error. (100% scanned).")
     .test(code, {es3: true});
 
   // Octals are prohibited in strict mode.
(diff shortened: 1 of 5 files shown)


caitp:
I mean, if we hit an unrecoverable error we should probably just report that and abort, am I wrong?

jugglinmike:
Yeah, I agree; [that's what `quit` does](https://github.com/jshint/jshint/blob/dd768c243af2bd3a8beb5e47768eef3ec7ab2e5e/src/jshint.js#L321-L334). The problem right now is that loops which don't explicitly check for this error don't terminate because they keep getting fed a new end-of-file token.

tests/unit/parser.js
```diff
@@ -4670,6 +4669,16 @@ exports.classExpressionThis = function (test) {
   test.done();
 };
 
+// Regression test for GH-2324
+exports.invalidClass = function (test) {
+  TestRun(test)
+    .addError(1, "Class properties must be methods. Expected '(' but instead saw ''.")
+    .addError(1, "Unrecoverable syntax error. (100% scanned).")
+    .test("class a { b", { esnext: true });
```
caitp: Lets add one with the text

```js
class Test {
  constructor: {
  }
}
```

too, (see #2339)

jugglinmike: Sounds good to me!

jugglinmike:
This also resolves the recently-reported gh-2362. I just added another regression test for that (to be squashed, of course)

rwaldron:
Ready for squashing?

jugglinmike:
Squashed!

PR #2341, opened 2015-04-25
[[FEAT]] support destructuring in ForIn/Of loops, lint bad ForIn/Of LHS

caitp:
Blocked on #2326 (test failures due to "const" not being block scoped)

Things to consider: maybe treat destructuring ForIn loops as "bad" and report an error? There isn't really any point to destructuring those

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3475,25 +3475,23 @@ var JSHINT = (function() {
           }
         }
       }
-      if (prefix) {
-        break;
-      }
 
       statement.first = statement.first.concat(names);
 
-      if (isConst && state.tokens.next.id !== "=") {
+      if (!prefix && isConst && state.tokens.next.id !== "=") {
         warning("E012", state.tokens.curr, state.tokens.curr.value);
       }
 
       if (state.tokens.next.id === "=") {
         advance("=");
-        if (state.tokens.next.id === "undefined") {
+        if (!prefix && state.tokens.next.id === "undefined") {
           warning("W080", state.tokens.prev, state.tokens.prev.value);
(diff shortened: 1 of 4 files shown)


src/jshint.js
```diff
@@ -4186,6 +4215,15 @@ var JSHINT = (function() {
         error("W104", nextop, "for of");
       }
 
+      var ok = !(initializer || comma);
+      if (initializer) {
```
jugglinmike: How about

```diff
-if (initializer) {
+if (initializer && state.option.inESNext()) {
```

caitp: technically the initializer is legal in es5.1, but I think we should lint it anyways, since it's meaningless in an enumeration context

jugglinmike: Yeah, it wouldn't hurt to be opinionated about this. I was worried this would be a breaking change, but I just checked and JSHint doesn't allow this now (just with a less helpful error, "Expected 'in' and instead saw '='."). The only concern I have is that users cannot ignore errors. I don't know if defining an analogous warning for this problem is really worthwhile, though, because I can't think of any case where someone would intentionally want this.

jayphelps:
This should be unblocked now :smile: via https://github.com/jshint/jshint/pull/2326#issuecomment-99132253

jayphelps:
Huge thanks for doing this btw. I'll finally get to turn jshint back on! :tada:

caitp:
rebased, and tests pass locally --- so this is probably good to land tonight. I'll check this in after travis is green, and we can revert if needed, but I'm pretty sure we're cool with it

PR #2344, opened 2015-04-26
Fix 2292

lukeapage:
the problem was that in the simple implied map

```
i = 1; // i goes into implied
let i; // this should error 
```

```
var a = () => { i = 1; /* i goes into implied */ };
let i; // this should not error
```

there was no way to differentiate between the two situations.

So I took a leaf out of the blockscope manager and created one for implied that manages the different scopes.

But then as I've been doing this I've become unsure about the approach of the (blockscope) manager and implied manager - that logic is already present for var's in order to detect var's out of scope

```
function() {
    for(var i = 0; i < 8; i++) {
        var j = 1;
    }
    for(j = 0; j < 8; j++) {
    }
}
```

because even though j is function scope, jshint detects that j was not declared in the outer scope.

and it seems to me the logic should be combined. whats more I prefer the explicitness of the implied manager and blockscope manager (I think the logic outside of it is difficult to follow, with little overall explanation).

so @caitp @jugglinmike .. what do you think.. do you think the approach here is okay? which way should it be going?

Would you be happy if I continued refactoring and put implieds, blockscope and scope all into one "manager" with better names to test and poke on different scenarios ?

### DIFF
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -860,6 +860,11 @@ exports.testConstModification = function (test) {
     "delete a[0];",
     "new a();",
     "new a;",
+    "function e() {",
+    "  f++;",
+    "}",
+    "const f = 1;",
+    "e();"
   ];
 
   TestRun(test)
@@ -871,6 +876,7 @@ exports.testConstModification = function (test) {
       .addError(8, "Attempting to override 'a' which is a constant.")
       .addError(8, "You might be leaking a variable (a) here.")
       .addError(53, "Missing '()' invoking a constructor.")
+      // .addError(55, "Attempting to override 'f' which is a constant.") todo gh-2368
       .test(src, {
         esnext: true
       });
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -458,6 +458,13 @@ var JSHINT = (function() {
       if (funct["(blockscope)"].atTop() && exported[name]) {
         state.tokens.curr.exported = true;
       }
+
+      if (implied.claimBlockImplied(name)) {
+        if (state.option.latedef) {
+          warning("W003", state.tokens.next, name);
+        }
+        funct["(blockscope)"].current.setUsed(name);
+      }
     } else {
       funct["(blockscope)"].shadow(name);
       funct[name] = type;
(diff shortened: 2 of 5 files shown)


tests/unit/core.js
```diff
@@ -872,6 +877,7 @@ exports.testConstModification = function (test) {
       .addError(8, "Attempting to override 'a' which is a constant.")
       .addError(8, "You might be leaking a variable (a) here.")
       .addError(53, "Missing '()' invoking a constructor.")
+      // .addError(17, "Attempting to override 'f' which is a constant.") TODO
```
lukeapage: do you put in failing tests or make a bug for them? not sure if I am going to work more on this and fix it yet...

caitp: If a test that should pass is failing, you can add a `//TODO(name of person who should look at it): short description`, and if you want you can also comment out some tests that are failing. File a bug too, you can link to the bug in the comment.

caitp: line 17 is `  let x = -a;` though, so I'm not too sure why this error is here or commented out

lukeapage:
@caitp @jugglinmike, now you've merged the other PR, would be good to get some discussion here. I think the best thing to do would be to discuss what should happen going forward so if I spend further time on this it will be accepted.

caitp:
I think it's fine to fix this within some limits --- particularly, I don't want to go too crazy trying to enforce TDZ semantics that can't be determined statically. But the examples in the PR description don't seem unreasonable

tests/unit/core.js
lukeapage: okay, If I don't fix it in this Pr I'll do that.
Line no is a mistake due to merging

lukeapage:
> I don't want to go too crazy trying to enforce TDZ semantics that can't be determined statically

I agree - determining whether or not a function is called before the declaration for instance would be going too far - right?

I'm most interested in your opinion on this:

> Would you be happy if I continued refactoring and put implieds, blockscope and scope all into one "manager" with better names to test and poke on different scenarios ?

or do you think that whats ended up here is where things should be at?

caitp:
>>Would you be happy if I continued refactoring and put implieds, blockscope and scope all into one "manager" with better names to test and poke on different scenarios ?

>or do you think that whats ended up here is where things should be at?

The churn has probably died down for a while, so if we're going to do that, this is a good time to do it

rwaldron:
@caitp based on your last comment, is this ready to go?

caitp:
Not sure if @lukepage wanted to work on it a bit more first or not.

rwaldron:
Got it!

lukeapage:
I am going to work on top of this, but I think its a big job and I don't know how much time I've got. I might have an update in a day or a week, or two. So, upto you.

lukeapage:
moved to #2435

PR #2355, opened 2015-04-30
The Fetch API adds the "fetch" global.

julienw:


caitp:
this is true, lgtm

tests/unit/fixtures/browser.js
```diff
@@ -25,3 +25,12 @@ var comment = new Comment("");
 var df = new DocumentFragment();
 var range = new Range();
 var text = new Text("");
+
+// Fetch example
```
julienw: maybe I should add that this comes from MDN https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch ?

silvenon:
:+1:

julienw:
I added the link to MDN.

tests/unit/fixtures/browser.js
julienw: done :)

rwaldron:
Thanks @julienw!

PR #2360, opened 2015-05-04
Options: toggle strict and globalstrict

keradus:
In docs (http://jshint.com/docs/options/)
`strict` should be enforcing option,
`globalstrict` should be relaxing option.


rwaldron:
Ok to land

Issue #2361, opened 2015-05-05
[es6] Support "u" and "y" flags for regular expressions


houkanshan:
ECMAScript 6 adds `u` & `y` as two new flags, according to
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-regexpinitialize

But JSHint doesn't support them:
https://github.com/jshint/jshint/blob/master/src%2Flex.js#L1382

Test:
```javascript
/* jshint esnext:true */
var uFlag = /𠮷/u;
var yFlag = /abc/y;
```

And the report:

```
regex.es6: line 2, col 17, Expected an assignment or function call and instead saw an expression.
regex.es6: line 3, col 18, Expected an assignment or function call and instead saw an expression.
regex.es6: line 2, col 17, 'u' is not defined.
regex.es6: line 3, col 18, 'y' is not defined.
```

jugglinmike:
Good one! I'll get right on this

jugglinmike:
My patch for this depends on the refactoring in gh-2306 -- I'll submit once we've landed that.

PR #2369, opened 2015-05-09
[[FIX]] Accept new RegExp flags introduced by ES6

jugglinmike:
This resolves gh-2361.

Commit message:

> ECMAScript 6 introduces two additional flags for regular expressions:
> "u" and "y". Extend the lexer to accept these when linting ES6 code.
>
> Existing code attempts to compile regular expression literals as a
> mechanism to ensure validity. Update this code to ignore regular
> expression flags in order to ensure consistent behavior when JSHint is
> executed in ES5 and ES6 environments.

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -40,6 +40,11 @@ var Context = {
   Template: 2
 };
 
+var validRegexFlags = {
+  es5: /[gim]/,
+  es6: /[gimuy]/
+};
+
 // Object that handles postponed lexing verifications that checks the parsed
 // environment state.
 
@@ -119,6 +124,7 @@ function Lexer(source) {
   this.inComment = false;
   this.context = [];
   this.templateStarts = [];
+  this.regexpFlags = validRegexFlags[state.inESNext() ? 'es6' : 'es5'];
 
   for (var i = 0; i < state.option.indent; i += 1) {
     state.tab += " ";
@@ -1379,7 +1385,7 @@ Lexer.prototype = {
 
     while (index < length) {
       char = this.peek(index);
-      if (!/[gim]/.test(char)) {
+      if (!this.regexpFlags.test(char)) {
         break;
       }
       flags.push(char);
@@ -1390,7 +1396,7 @@ Lexer.prototype = {
     // Check regular expression for correctness.
 
     try {
-      new RegExp(body, flags.join(""));
+      new RegExp(body);
     } catch (err) {
       malformed = true;
       this.trigger("error", {
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -510,6 +510,22 @@ exports.regexp = function (test) {
   TestRun(test).test("var a = 1; var b = a-- / 10;", {esnext: true});
   TestRun(test).test("var a = 1; var b = a-- / 10;", {moz: true});
 
+  TestRun(test)
+    .addError(1, "Missing semicolon.")
+    .addError(1, "Expected an assignment or function call and instead saw an expression.")
+    .test("var a = /.*/u;");
+
+  TestRun(test)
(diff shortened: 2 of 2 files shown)


caitp:
I feel like @mathiasbynens should maybe take a look

jugglinmike:
Sounds good to me!

I just pushed an additional commit to explicitly ensure that the flags aren't repeated (something we got "for free" when passing the flags to the `RegExp` constructor). Because we're no longer relying on the environment's regular expression parser to verify the flags, we'll need to be especially careful about false-negative regressions. Certainly these won't break any builds, but that's no reason to accept less strict enforcement.

src/lex.js
```diff
@@ -1390,7 +1407,7 @@ Lexer.prototype = {
     // Check regular expression for correctness.
 
     try {
-      new RegExp(body, flags.join(""));
+      new RegExp(body);
```
mathiasbynens: This fails for e.g. `/\u{1}/u` and `/[💩-💫]/u`.

See https://mathiasbynens.be/notes/es6-unicode-regex.

jugglinmike: Thanks, @mathiasbynens ! Cait was right to ask for your input :) I plan to incorporate that logic from Esprima into this patch, but I had an idea for a small improvement. I just opened https://github.com/jquery/esprima/pull/1180 to propose it.

jugglinmike:
@caitp I've worked with @mathiasbynens to improve esprima's RegExp parser (marginally) and back-ported the latest version to this patch. Mind giving it another look?

tests/unit/parser.js
```diff
@@ -510,10 +512,63 @@ exports.regexp = function (test) {
...
+
+  test.done();
+};
+
+exports.regexp.yFlag = function (test) {
+  // Flag validity
+  TestRun(test)
+    .addError(1, "Missing semicolon.")
+    .addError(1, "Expected an assignment or function call and instead saw an expression.")
+    .test("var a = /.*/y;");
```
nicolo-ribaudo: I think it would be better to throw a [W119](https://github.com/jshint/jshint/blob/master/src/messages.js#L197) warning.

jugglinmike: Good idea!

jugglinmike:
I've rebased this work (including @nicolo-ribaudo's suggested warning
improvement) and force-pushed the result. It should be good to go!

Note that we merged support for the `y` flag in gh-2999 (opened in the time
since this pull request was first submitted), so this patch now only concerns
the `u` flag.

jugglinmike:
(The original version of this branch is available for posterity here: https://github.com/jugglinmike/jshint/tree/2361-es6-regexp-flags-orig)

jugglinmike:
Mind taking a look at this, Cait?

rwaldron:
This looks good to me, are we still waiting for @caitp or should I go ahead and merge?

jugglinmike:
Hey @rwaldron,

I found this while working on support for ES2018 named capture groups.

Years later, I can see that it was woefully incomplete. While it correctly interprets the new Unicode escape sequences, there are a whole bunch of early errors that we need to re-implement. I've pushed up a `fixup!` commit which introduces those.

While working, I found a simplification that while not technically necessary, would make the code (and the review) easier. I've implemented that as a dedicated commit to help keep the changeset for the feature addition more focused.

Since the test suite has changed since I first submitted this patch (it now requires the expected column number of all errors in addition to the expected line number), I needed to rebase. Given the age of the pull request, I doubt this is very disruptive, but in the interest of transparency, [I've preserved the previous version of this branch (itself a rebase) as `2361-es6-regexp-flags-2`](https://github.com/jshint/jshint/compare/master...jugglinmike:2361-es6-regexp-flags-2).

PR #2373, opened 2015-05-11
[[FEAT]] config file extension: expand tilde

gurdiga:
This patch allows me to use "~/" in configuration's "extends" option.
For example:

{
  "extends": "~/.jshintrc",
  ...
}

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "htmlparser2":         "3.8.x",
     "minimatch":           "2.0.x",
     "shelljs":             "0.3.x",
+    "expand-home-dir":     "0.0.2",
     "strip-json-comments": "1.0.x",
     "lodash":              "3.7.x"
   },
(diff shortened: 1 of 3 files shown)


gurdiga:
Just in case, I will elaborate a bit on the reasoning behind this change.

The first thing that came to mind when I found about the `extends` option was:

> — Oh, that’s cool! I can have a base `~/.jshintrc`, and then add tweaks for every (part of) project! For example I have different needs for Node code and browser code.

And then I found that `jshint` reads `~/`literally.

So, what do you guys think?

jugglinmike:
Two thoughts:

1. The tilde does not behave this way on all platforms (I'm thinking of Windows specifically). Introducing the *nix semantics here may be surprising for some users
2. This use case quickly falls apart for distributed teams where it will be difficult to synchronize an externally-managed configuration file. I think this is potentially pernicious even for personal use. For instance, my own linting preferences have changed over time, but I do still return to old projects now and then. It would be frustrating to begin working on an old project only to find that it no longer passed its own linting checks because my global configuration changed.

gurdiga:
OK, makes sense. I didn’t think about it that deep, sorry. :)

jugglinmike:
No apology necessary. I appreciate the effort!

PR #2377, opened 2015-05-11
[[FIX]] Allow `__proto__` when using ES6

### LINKED ISSUE #2371: [es6] __proto__ isn't deprecated
For the [spec](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-__proto__-property-names-in-object-initializers), browsers must implement `__proto__`, so:

- If both `esnext` and `proto` are set to `false`, JSHint should warn that `'__proto__' is only available in ES6 (use esnext option).`
- If `esnext` is set to `true` but `browesr`, `node` and `rhino` are set to `false`, JSHint should warn that `The '__proto__' property may not be supported by non-browser environments.`.

nicolo-ribaudo:
In ECMAScript 6 `__proto__` is no longer deprecated (but non-browser environments may not
support it).

- If proto option is on, JSHint behaves as before.
- If esnext option is off , JSHint warns: `'__proto__' is available in ES6 (use esnext option).`. (W119)
- If esnext is is on but browser, node, and rhino options are off, JSHint warns: `The '__proto__ property may not be supported by non-browser environments.` (New W135 warning).

Fixes #2371

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -5,15 +5,18 @@ exports.register = function(linter) {
   // deprecated and then re-introduced for ES6.
 
   linter.on("Identifier", function style_scanProto(data) {
-    if (linter.getOption("proto")) {
+    var esnext = linter.getOption("esnext");
+
+    if (linter.getOption("proto") || (esnext && (linter.getOption("browser") ||
+        linter.getOption("node") || linter.getOption("rhino")))) {
       return;
     }
 
     if (data.name === "__proto__") {
-      linter.warn("W103", {
+      linter.warn(esnext ? "W135" : "W119", {
         line: data.line,
         char: data.char,
-        data: [ data.name ]
+        data: [ esnext ? "The '__proto__' property" : "__proto__" ]
       });
     }
   });
--- a/src/style.js
+++ b/src/style.js
@@ -7,8 +7,7 @@ exports.register = function(linter) {
   linter.on("Identifier", function style_scanProto(data) {
     var esnext = linter.getOption("esnext");
 
-    if (linter.getOption("proto") || (esnext && (linter.getOption("browser") ||
-        linter.getOption("node") || linter.getOption("rhino")))) {
+    if (linter.getOption("proto") || (esnext && linter.getOption("browser"))) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


src/style.js
```diff
@@ -5,15 +5,18 @@ exports.register = function(linter) {
   // deprecated and then re-introduced for ES6.
 
   linter.on("Identifier", function style_scanProto(data) {
-    if (linter.getOption("proto")) {
+    var esnext = linter.getOption("esnext");
+
+    if (linter.getOption("proto") || (esnext && (linter.getOption("browser") ||
+        linter.getOption("node") || linter.getOption("rhino")))) {
```
lukeapage: I spent a while looking at this line thinking it was wrong, till noticing the `(esnext && ...` - any chance we can reformat this to make it clearer?

e.g.
```
    if (linter.getOption("proto") ||
       (esnext && (linter.getOption("browser") || linter.getOption("node") || linter.getOption("rhino")))) {
```

So.. is `__proto__` annex b ? and node and rhino are classed as "browser environments" ?

If that's the case I wonder whether the error message should be more specific? if you didn't have the node option on, it wouldn't be obvious that the node option switches it off ?

nicolo-ribaudo: > I spent a while looking at this line thinking it was wrong, till noticing the (esnext && ... - any chance we can reformat this to make it clearer?

Would this be clearer?
```js
if (data.name !== "__proto__" || linter.getOption("proto")) {
  return;
}

var code;

if (linter.getOption("esnext")) {
  if (linter.getOption("browser") || linter.getOption("node") || linter.getOption("rhino")) {
    return;
  }
  code = "W135";
} else {
  code = "W119";
}

linter.warn(code, {
  line: data.line,
  char: data.char,
  data: [ "__proto__" ]
});
```

lukeapage: yep, I think so :+1:

nicolo-ribaudo: I edited my comment after your response...

lukeapage: Okay, good

lukeapage:
committed - https://github.com/jshint/jshint/commit/06b57649414e44a9df6a46907f3ac93513e44221
thanks.

PR #2378, opened 2015-05-11
[[CHORE]] Allow failures in coverage report

jugglinmike:
The Coveralls service occasionally fails to respond within the
continuous integration service's timeout interval. When invoked as part
of the test script, these timeouts trigger build failures.

Move coverage reporting to a task that occurs after build completion[1] so
that these timeouts do not effect build status.

This change also ensures that the project's full set of tests are run in
the continuous integration environment. Previously, only those tests
necessary for generating a code coverage report were executed (omitting
code style and linting checks).

[1] From the TravisCI service's documentation:

> The outcome of any of these commands (except `after_success`,
> `after_failure` or `after_script`) indicates whether or not this build
> has failed or passed.

http://docs.travis-ci.com/user/build-configuration/#Build-Lifecycle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-script: npm run coverage && cat ./coverage/lcov.info | coveralls
+after_success: npm run coverage && cat ./coverage/lcov.info | coveralls
 node_js:
   - "0.10"
   - "0.12"


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: node_js
-script: npm run coverage && cat ./coverage/lcov.info | coveralls
+after_script: npm run coverage && cat ./coverage/lcov.info | coveralls
```
caitp: this seems slightly wrong --- I think `npm run coverage` does the actual testing stuff. Probably just want to move the `cat ./coverage/lcov.info | coveralls` to after_script

caitp:
otherwise LGTM

.travis.yml
jugglinmike: It does some of the testing stuff, but not all. `script: npm test` would also run our JSCS and JSHint checks, but since the language is `node_js`, this is used implicitly... I think. We can wait for the build report to verify.

jugglinmike: [Verified](https://travis-ci.org/jshint/jshint/jobs/62116677):

```
The command "npm test" exited with 0.
after_script
15.61s
$ npm run coverage && cat ./coverage/lcov.info | coveralls
Done. Your build exited with 0.
```

jugglinmike:
@caitp One other thing to note: this is a little inefficient because a successful `npm test` executes all the unit tests, which the `after_script` repeats to generate coverage information. Looks like the unit tests take 7977ms, while coverage generation takes an additional 15.61s, but both are dwarfed by the average time we spend waiting for the job to run on TravisCI.org's free queue.

Also, would you prefer to see `script: npm test` declared in the `.travis.yml` file for the sake of explicitness?

caitp:
I think we can use `after_success` or something to only do the coverage stuff on successful builds (which is a tiny bit less wasteful).

We could also restructure the `script` task to avoid duplicating work

jugglinmike:
Coveralls is not responding to any requests today, blocking ongoing work. I've updated to use `after_success` instead of `after_script`.

PR #2379, opened 2015-05-11
[[FIX]] JSHint should warn about duplicate key '__proto__'.

### LINKED ISSUE #2372: JSHint doesn't warn about duplicated __proto__
```javascript
/*jshint proto: true */

var a = {
  __proto__: null,
  __proto__: {}
};
```

Excepted warning: `Duplicate key '__proto__'`

nicolo-ribaudo:
Spec (https://people.mozilla.org/~jorendorff/es6-draft.html#sec-B.3.1):
It is a *Syntax Error* if *PropertyNameList* of *PropertyDefinitionList* contains any duplicate entries for "\__proto\__" and at least two of those entries were obtained from productions of the form
*PropertyDefinition : PropertyName : AssignmentExpression* .

Fixes #2372

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4732,14 +4732,15 @@ var JSHINT = (function() {
       name = tkn.value;
     }
 
-    if (props[name] && _.has(props, name)) {
+    if ((props[name] && _.has(props, name)) || (name === "__proto__" &&
+        Object.getPrototypeOf(props) !== Object.prototype)) { // __proto__ has been modified
       warning("W075", state.tokens.next, msg, name);
-    } else {
-      props[name] = {};
     }
 
-    props[name].basic = true;
-    props[name].basictkn = tkn;
+    props[name] = {
+      basic: true,
+      basictkn: tkn
+    };
   }
 
   /**
--- a/tests/unit/envs.js
+++ b/tests/unit/envs.js
@@ -189,6 +189,13 @@ exports.es5 = function (test) {
   src = fs.readFileSync(__dirname + "/fixtures/es5.funcexpr.js", "utf8");
   TestRun(test).test(src, {  }); // es5
 
+  // JSHint should warn about duplicate key '__proto__'.
+  // (https://github.com/jshint/jshint/issues/2372)
+  src = "var a = { __proto__: null, __proto__: {} };";
+  TestRun(test)
+    .addError(1, "Duplicate key '__proto__'.")
+    .test(src, { proto: true });
+
   test.done();
 };
 


src/jshint.js
```diff
@@ -4732,14 +4732,15 @@ var JSHINT = (function() {
       name = tkn.value;
     }
 
-    if (props[name] && _.has(props, name)) {
+    if ((props[name] && _.has(props, name)) || (name === "__proto__" &&
+        Object.getPrototypeOf(props) !== Object.prototype)) { // __proto__ has been modified
```
caitp: Since this is annex B stuff, do you want to make this specific to the `browser` environment?

nicolo-ribaudo: I think JSHint should always warn about this, as it does with every other property.

lukeapage: It should always warn, I guess it should error if its `__proto__` but thats a minor point.

src/jshint.js
```diff
@@ -4732,14 +4732,15 @@ var JSHINT = (function() {
...
       warning("W075", state.tokens.next, msg, name);
-    } else {
-      props[name] = {};
     }
 
-    props[name].basic = true;
-    props[name].basictkn = tkn;
+    props[name] = {
```
lukeapage: was this change to fix anything? I guess its okay, since it is a new property we are now referring to.. have scoured the code to see if there is anything it could break and seems okay (accessors seem to be the only thing storing something else against the property and they are set elsewhere)

nicolo-ribaudo: I changed this just to remove the else block.

nicolo-ribaudo:
What do you think about this code?
```js
var __proto__ = 1;
var a = {
  __proto__: [],
  __proto__
};
```

This isn't a duplicated property, as `Object.getPrototypeOf(a)` is `[]`, but `a.__proto__` is `1`.

lukeapage:
ha, well IMO, I could be wrong, but I think that most likely the person writing
```
var a = {
  __proto__
};
```
probably made a mistake.. It seems to me like a dangerous thing to do.

I doubt anyone will complain if you still error..

src/jshint.js
lukeapage: I think it might be safer if you didn't change it.. but I don't feel massively strongly.. wdyt?

jugglinmike:
@nicolo-ribaudo Despite appearances, I have not forgotten about this patch. It's what prompted my patch at gh-2539 last week, and just now it helped me identify [a regression in `master`](https://github.com/jshint/jshint/issues/2419#issuecomment-121777097). It's been a deceptively involved review process for such an edge case, but I wanted to let you know that I'm still working with it.

nicolo-ribaudo:
Would this be a breaking change?

jugglinmike:
Superseded by gh-2566. The change there is preferable (now that gh-2559 has landed) because it covers additional identifier usages and because it works in more environments. Thanks for reporting the problem!

PR #2380, opened 2015-05-11
[[FIX]] Don't throw "Duplicate class method" with computed method names

### LINKED ISSUE #2350: [es6] "Duplicate class method" when two methods' names are got from an object
```javascript
/*jslint esnext: true */

const a = {
  b: 1,
  c: 2
};

class D {
  [a.b]() {}
  [a.c]() {}
}
```

I get this warning: `Duplicate class method '.'.`

nicolo-ribaudo:
Fixes #2350


caitp:
lgtm --- can you please add a test case which failed before? should be easy

caitp:
nevermind, somehow I didn't see the test case that was there already =) looks good

tests/unit/parser.js
```diff
@@ -5271,50 +5271,9 @@ exports.classes = function (test) {
 };
 
 exports["class and method naming"] = function (test) {
-  var code = [
```
caitp: I'd personally prefer that this was not moved into a fixture. It's easier to read in one file.

I don't see any reason to add the extra stuff to this same code though, you can write a new test function for that

jugglinmike:
Nice catch! Squashed and merged to `master` at commit ab12dfb558d45654a160cb2a3b74b38d355e61b6 . Thanks :)

PR #2382, opened 2015-05-12
[[FIX]] Ignore unused arrow-function parameters if unused: vars

### LINKED ISSUE #2345: Unused arrow-function parameters not ignored by "unused": "vars"
Setting `"unused": "vars"` disables all warnings concerning unused normal function parameters. However the new arrow functions don't seem to be supported by this option:

    on().then( thing => console.log( 'done' ) );

    'thing' is defined but never used.

I feel these should behave the same as normal function parameters when it comes to raising warnings.

nicolo-ribaudo:
Fixes #2345

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2880,7 +2880,7 @@ var JSHINT = (function() {
 
     if (loneArg && loneArg.identifier === true) {
       addlabel(loneArg.value, { type: "unused", token: loneArg });
-      return [loneArg];
+      return [loneArg.value];
     }
 
     next = state.tokens.next;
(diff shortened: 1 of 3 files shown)


rwaldron:
Makes sense to me, @caitp @jugglinmike ?

src/jshint.js
```diff
@@ -2880,7 +2880,7 @@ var JSHINT = (function() {
 
     if (loneArg && loneArg.identifier === true) {
```
caitp: Unrelated, but we should really, really make it clear that `loneArg` is a token --- maybe rename it to clarify that

caitp:
yeah, makes sense to me too

jugglinmike:
Agreed. Thanks, Nicolò!

PR #2384, opened 2015-05-13
[[FIX]] Don't warn when Array() is used without 'new'.

### LINKED ISSUE #1987: Allow calling Array as a function
The ECMAScript 5.1 specs [allow](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.1) calling `Array(…)` as a function, and the effect is the same as the `new Array(…)` constructor invocation. The same goes for other native constructors, e.g. `Function`. Currently, JSHint will issue a warning for this, though:

    Missing 'new' prefix when invoking a constructor. (W064)

It would be nice if there were some option to control the behavior here: enforce constructor syntax, enforce function syntax, or allow mixing both. It would also be nice if we could include the name of the constructor in the message, so that one can easily distinguish between things which may be used as functions and things which may not be.

Looking at the current source, I see the check is implemented as

    if ("Number String Boolean Date Object Error".indexOf(left.value) === -1)

This means that constructors which are *substrings* of the named native types will be subject to the same exceptions: a user-defined type called `Num`, `Str`, `Bool` or `Obj` would *not* trigger the error. I believe some proper set lookup would be preferable here. Or perhaps a regular expression matching all exceptions.

JasonCust:
Comply with [ECMAScript 5.1
spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.1) to allow `Array` to be called as a
function.

Fixes #1987

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2496,7 +2496,7 @@ var JSHINT = (function() {
     if (left) {
       if (left.type === "(identifier)") {
         if (left.value.match(/^[A-Z]([A-Z0-9_$]*[a-z][A-Za-z0-9_$]*)?$/)) {
-          if ("Number String Boolean Date Object Error Symbol".indexOf(left.value) === -1) {
+          if ("Array Number String Boolean Date Object Error Symbol".indexOf(left.value) === -1) {
             if (left.value === "Math") {
               warning("W063", left);
             } else if (state.option.newcap) {
--- a/tests/unit/fixtures/newcap.js
+++ b/tests/unit/fixtures/newcap.js
@@ -13,6 +13,7 @@ rat = iAnimal();
 bat = myAnimal();
 
 // Make sure we don't warn on Error, Number, etc.
+Array();
 Error();
 Number();
 String();


caitp:
Also section 15.4.1 in ES3 and section 22.1.1 in ES6 cover this behaviour as well. I'm guessing this was leftover from jslint.

LGTM

src/jshint.js
```diff
@@ -2496,7 +2496,7 @@ var JSHINT = (function() {
     if (left) {
       if (left.type === "(identifier)") {
         if (left.value.match(/^[A-Z]([A-Z0-9_$]*[a-z][A-Za-z0-9_$]*)?$/)) {
-          if ("Number String Boolean Date Object Error Symbol".indexOf(left.value) === -1) {
+          if ("Array Number String Boolean Date Object Error Symbol".indexOf(left.value) === -1) {
```
caitp: I think I do agree with the OP for the referenced bug though, doing this using `indexOf` is a bit iffy, because it means `BigNumber` or `LocalizedString` or `MyFancyError` are effected as well.

Probably a good idea to make a map in vars.js like

```js
var kCallableConstructors = Object.setPrototypeOf({
  Array: true,
  Number: true,
  String: true,
  Boolean: true,
  Date: true,
  Object: true,
  Symbol: true,
  Error: true,
  TypeError: true,
  SyntaxError: true,
  ReferenceError: true,
  // ETC for all the crazy DOM errors and such
}, null);
```

JasonCust: That seems more appropriate. I was basing this fix off of a previous commit for a similar issue: https://github.com/jshint/jshint/commit/2fecdea9fbc8c66e659b4be7035f03bab16b855e

rwaldron: > ` Symbol: true,`

Perhaps we should have a rule that warns when code includes `new` incorrectly? For example: `new Symbol()` is invalid.

rwaldron: Of course, not in this patch

jugglinmike: @caitp I don't think that's actually a problem:

```js
"Array Number String Boolean Date Object Error Symbol".indexOf("Number");
6
"Array Number String Boolean Date Object Error Symbol".indexOf("BigNumber");
-1
```

What could be a problem is truncated constructor names:

```js
"Array Number String Boolean Date Object Error Symbol".indexOf("umber");
7
```

...but this code path is only hit when the constructor begins with a capital letter. So while it's a bit of a hazard, [the option is deprecated](https://github.com/jshint/jshint/blob/7aa3e23e7e7bc53d1b19da25e669b30b6f75c5f5/src/options.js#L164-L183). I think this makes it safe to accept as-is. Can you confirm?

lukeapage: I agree.  The existing code would hit e.g. "Bool"..  So @caitp suggestion is better,  but this pr doesnt introduce that code and as you said its deprecated so this lgtm.

jugglinmike:
Merged to `master`: 5f88aa76eb1b4ec5acfba4a539e8ea33c965dc78 . Thanks, Jason!

Issue #2261, opened 2015-03-16
unused variable, but does not exist


mMerlin:
Environment: http://jshint.com/ as of 2015/03/15

```javascript
/*global */
var a - 10;
console.log(a);
```

jslint.com says there is an unused variable at line 1, but with no name.  It does not like it when the global tag ends up empty.

PR #2386, opened 2015-05-14
[[FIX]] Do not accept empty values for directives

jugglinmike:
Previously, when an directive was specified without a value, JSHint
would interpret this as the empty string and apply linting rules
accordingly. The empty string is never a valid value for any directive,
so the generated warnings were not appropriate.

If a directive is specified without a value, trigger an explicit error
describing the invalid configuration.

Resolves gh-2261

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -213,6 +213,47 @@ exports.options = function (test) {
   test.done();
 };
 
+exports.emptyDirectives = function (test) {
+  TestRun(test)
+    .addError(1, "Bad option value.")
+    .test('/* global */');
+
+  TestRun(test)
+    .addError(1, "Bad option value.")
+    .test('/* global : */');
+
+  TestRun(test)
+    .addError(1, "Bad option value.")
+    .test('/* global -: */');
+
+  TestRun(test)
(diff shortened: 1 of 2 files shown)


tests/unit/parser.js
```diff
@@ -213,6 +213,47 @@ exports.options = function (test) {
...
+  TestRun(test)
+    .addError(1, "Bad option value.")
+    .test('/* global : */');
+
+  TestRun(test)
+    .addError(1, "Bad option value.")
+    .test('/* global -: */');
+
+  TestRun(test)
+    .test('/* global foo, bar, baz, */');
```
lukeapage: I came across an instance where it was adding empty string as a global when the , is on the last item..
I agree on not erroring, but it would be good to skip adding it to `predef` or `exported`... ?

jugglinmike: Sure: I just pushed a commit to address this https://github.com/jshint/jshint/commit/8e4307d9571964fc65a3ee381c7b6d57f72b3dee . Is that what you had in mind?

lukeapage: Yes thanks.  It would be good to get this pulled before mine so i can remove the partial fix i put in.

PR #2389, opened 2015-05-15
[[FIX]] Reset generator flag for each method definition

### LINKED ISSUE #2388: Non-generator method after generator method is treated as generator method
If an object literal contains a generator method and subsequent non-generator method, that non-generator method is parsed as generator method, and it shows an error about missing ```yield``` statement.

Test case:

```
var obj = {
  *gen() {
    yield 1;
  },
  non_gen() {
  }
};
```

Expected result:
It passes.

Actual result:
It shows following error

```
test.js: line 6, col 3, A generator function shall contain a yield statement.
```

arai-a:
This patch fixes the handling of non-generator method after generator method.
For examle:

    var a = {
      *gen() {
        yield 1;
      },
      non_gen() {
      }
    };

Here, non_gen was treated as a generator function, since generator flag `g`
wasn't reset to false when `*` wasn't found before method name.

Fixes #2388

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3216,7 +3216,7 @@ var JSHINT = (function() {
 
   (function(x) {
     x.nud = function() {
-      var b, f, i, p, t, g, nextVal;
+      var b, f, i, p, t, isGeneratorMethod = false, nextVal;
       var props = {}; // All properties, including accessors
 
       b = state.tokens.curr.line !== startLine(state.tokens.next);
@@ -3281,7 +3281,9 @@ var JSHINT = (function() {
               warning("W104", state.tokens.next, "generator functions");
             }
             advance("*");
-            g = true;
+            isGeneratorMethod = true;
+          } else {
+            isGeneratorMethod = false;
           }
 
           if (state.tokens.next.id === "[") {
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3282,6 +3282,8 @@ var JSHINT = (function() {
             }
             advance("*");
             g = true;
+          } else {
+            g = false;
```
caitp: unrelated nit, I really think we should use a different name than `g` for this :)

tests/unit/parser.js
```diff
@@ -6584,3 +6584,21 @@ exports.functionKeyword = function (test) {
...
 };
+
+exports.nonGeneratorAfterGenerator = function (test) {
+  var run;
+  var code = [
+    'var obj = {',
+    '  *gen() {',
+    '    yield 1;',
+    '  },',
+    '  non_gen() {',
```
caitp: It's not super-clear what this is testing. Do we normally report an error if we think we have a generator without a yield expression? If so, probably worth marking that in a comment, just to help reviewer-sanity =)

caitp:
Otherwise, LGTM

src/jshint.js
arai-a: how about `isGenerator`?

caitp: yeah, something like that or `isGeneratorMethod` or something

src/jshint.js
```diff
@@ -3216,7 +3216,7 @@ var JSHINT = (function() {
 
   (function(x) {
     x.nud = function() {
-      var b, f, i, p, t, g, nextVal;
+      var b, f, i, p, t, isGeneratorMethod, nextVal;
```
rwaldron: Declare it with a value of `false`

caitp:
this is good to merge, imho =)

rwaldron:
Once it's full green, we can land

PR #2391, opened 2015-05-15
[[FIX]] Allow default values in destructuring.

### LINKED ISSUE #1941: Default values in deconstructed function parameter list
The following is valid ES6:

    function x({delay = 10} = {}) {...

This means, default the parameter to an empty object, then pull the delay property from it, but if it doesn't exist use 10 instead.

JSHint gives the message:

    Expected an identifier and instead saw '10'.

nicolo-ribaudo:
Fixes #1941

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3367,21 +3367,33 @@ var JSHINT = (function() {
     if (checkPunctuators(state.tokens.next, ["["])) {
       advance("[");
       var element_after_rest = false;
-      if (nextInnerDE() && checkPunctuators(state.tokens.next, [","]) &&
-          !element_after_rest) {
+      if (nextInnerDE() && checkPunctuators(state.tokens.next, [","])) {
         warning("W130", state.tokens.next);
         element_after_rest = true;
       }
       while (!checkPunctuators(state.tokens.next, ["]"])) {
-        advance(",");
-        if (checkPunctuators(state.tokens.next, ["]"])) {
-          // Trailing comma
-          break;
+        if (checkPunctuators(state.tokens.next, ["="])) {
+          if (checkPunctuators(state.tokens.prev, ["..."])) {
+            advance("]");
+          } else {
+            advance("=");
+          }
+          if (state.tokens.next.id === "undefined") {
+            warning("W080", state.tokens.prev, state.tokens.prev.value);
+          }
+          expression(10);
         }
-        if (nextInnerDE() && checkPunctuators(state.tokens.next, [","]) &&
-            !element_after_rest) {
-          warning("W130", state.tokens.next);
-          element_after_rest = true;
+        if (!checkPunctuators(state.tokens.next, ["]"])) {
+          advance(",");
+          if (checkPunctuators(state.tokens.next, ["]"])) {
+            // Trailing comma
+            break;
+          }
+          if (nextInnerDE() && checkPunctuators(state.tokens.next, [","]) &&
+              !element_after_rest) {
+            warning("W130", state.tokens.next);
+            element_after_rest = true;
+          }
         }
       }
       advance("]");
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3373,15 +3373,21 @@ var JSHINT = (function() {
         element_after_rest = true;
       }
       while (!checkPunctuators(state.tokens.next, ["]"])) {
-        advance(",");
-        if (checkPunctuators(state.tokens.next, ["]"])) {
-          // Trailing comma
-          break;
+        if (checkPunctuators(state.tokens.next, ["="])) {
+          advance("=");
+          expression(10);
```
caitp: Do we want to do any destructuring on the default expression? I feel like if we get an undefined literal we should emit a warning, just because it's redundant

caitp: s/destructuring/linting --- my brain hurts, too much looking at destructuring today =)

tests/unit/parser.js
```diff
@@ -2716,6 +2740,21 @@ exports["test destructuring function as legacy JS"] = function (test) {
   test.done();
 };
 
+exports["destructuring function default values"] = function (test) {
+  var code = [
+    "function a([ b = 2 ]) {}",
```
caitp: How about `function a([ first = 0, second = 1 ] = []) {}`?

caitp: we might already test that separately, i'll check

caitp: I am pretty sure we don't, but we do have some nonsense tests about default arguments in core :<

nicolo-ribaudo:
@catip Is this code valid?
```javascript
let {
  ...a = [ 1, 2 ]
} = {};
```

caitp:
>@caitp Is this code valid?
>
>```js
>let [
>  ...a = [ 1, 2 ]
>]=[];
```

No, AssignmentRestElement can't have an `Initializer`.

**Comment by nicolo-ribaudo** [comment]
What warning should JSHint throw?
Is `Rest assignment can't have a default value.` ok?

**Comment by jugglinmike** [comment]
Using `advance("]")` after parsing an AssignmentRestElement will cause your example to generate an error like "Expected ']' but instead saw '='." which ought to be descriptive enough.

**Comment by nicolo-ribaudo** [comment]
I updated the commit because JSHint generated too warnings for this code:
```js
/*jshint esnext: true, unused: false */

let [ ...a = 2 ] = [];
```
```
line 3, col 12, Expected ']' and instead saw '='.
line 3, col 14, Expected '=' and instead saw '2'.
line 3, col 16, Expected an identifier and instead saw ']'.
line 3, col 18, Expected an operator and instead saw '='.
line 3, col 20, Expected an operator and instead saw '['.
```

Now JSHint throws just `line 3, col 12, Expected ']' and instead saw '='.`.

jugglinmike:
Please try not to force push during a code review; it makes it difficult for reviewers to understand how your patch has changed.

nicolo-ribaudo:
I'm sorry, should I split this commit until the code review is done?

jugglinmike:
It's not a big deal, just a suggestion for future alterations

jugglinmike:
Merged to `master` at commit 04ace9a0e2094ce8e7dcfc3c370961ab8c8c1d52 . Thanks!

snig-b:
From which version is this available? I'm on version 2.9.1 and it doesn't work on that.

nicolo-ribaudo:
This should be available since JSHint 2.9.0. Can you open a new issue with an example to reproduce the bug?

snig-b:
I am getting this error on my Pycharm jshint library. This is not reproducible on your online editor, just my JetBrains library so not sure whether it requires a new issue.
The version used is 2.9.1
This is my statement : 

```constructor ({operationIdentifier='eq', comparisonValue} = {}) {```

I get the following errors in this statement:

1. ```, expected``` at **equals** operator at col 38.
2. ```statement expected``` at col 39 at **single quote**.
3. ```statement expected``` at **comma** operator at col 43.
4.  ```expression expected``` at col 62 at second **equals**.
5. ```unterminated statement``` & ```Expecting newline or semicolon``` at col 65 at ending **paranthesis**.

**Comment by nicolo-ribaudo** [comment]
> This is not reproducible on your online editor, just my JetBrains library so not sure whether it requires a new issue.

Then this is not an issue with JSHint.
Anyway, are you sure you are using JSHint 2.9.1? Does this code generate a warning? If it doesn't, you are using an older version.
```js
// jshint varstmt: true
for (var a in {});
```

PR #2404, opened 2015-05-19
[[FIX]] Prevent regression in `enforceall`

jugglinmike:
Ensure that a recent change to option organization does not effect the
behavior of the `enforceall` option. Update the documentation to more
accurately describe the behavior of `enforceall`.

This change is necessary since landing gh-2360 and should be merged prior to any patch release or minor release.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -284,12 +284,14 @@ exports.bool = {
     varstmt: false,
 
     /**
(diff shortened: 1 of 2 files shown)


caitp:
lgtm

jugglinmike:
Thanks!

PR #2405, opened 2015-05-19
[[FIX]] Remove module import declaration

### LINKED ISSUE #2395: jshint incorrectly supports the module x from 'y' format?
I noticed this is allowed...

```
module x from 'y';
```

Is that right? I haven't seen it in es6 or am I wrong? I'm not a spec expert.

should it be removed from jshint or put into mozilla javascript extensions, which has other non es6 stuff in like the strange let statement.

jamestaylr:
Fixes #2395 to remove the deprecated `module` import declaration from the parser and the associated test.

### DIFF
--- a/tests/unit/fixtures/module-keyword.js
+++ b/tests/unit/fixtures/module-keyword.js
@@ -1,9 +1,0 @@
-/* jshint undef: true */
-
-module ng from "angular";
-module em from "ember"
-
-ng.controller();
-
-var module = {};
-module.exports = "foo";
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1584,23 +1584,6 @@ var JSHINT = (function() {
       res = false;
     }
 
-    // detect a module import declaration
-    if (t.value === "module" && t.type === "(identifier)") {
-      if (peek().type === "(identifier)") {
-        if (!state.inESNext()) {
-          warning("W119", state.tokens.curr, "module");
-        }
-
-        advance("module");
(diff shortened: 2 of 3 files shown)


tests/unit/fixtures/module-keyword.js
```diff
@@ -1,9 +0,0 @@
-/* jshint undef: true */
-
-module ng from "angular";
```
caitp: Need to get rid of the test that loads this fixture too

caitp: Looks like tests/unit/core.js -> testModuleKeyword

jamestaylr: Okay I made the changes. Thank you!

caitp:
lgtm. @rwaldron this ok to you?

lukeapage:
@jugglinmike are we okay pulling this given (again) its technically a breaking change, though given the usage of the old module pattern it may be minimal. master or jshint 3?

jugglinmike:
Yeah, given that JSHint is a tool for linting JavaScript, and that this is official no longer JavaScript, we can call this change a bug fix.

jugglinmike:
@lukeapage Mind squashing and doing the honors?

lukeapage:
Thanks will do

lukeapage:
on master, thanks @jamestaylr 1749ac06d1aec9c195b3482171c5a6a94c3845a5

PR #2413, opened 2015-05-23
[[FIX]] Allow "\u{...}" escape sequence.

nicolo-ribaudo:
In ECMAScript 6 has been introduced the \u{...} syntax (valid both in identifiers and strings).
Astral code points in identifiers are now valid.

Examples:
```javascript
var \u{65}; // var \u0061;
var \u{102A7}; // astral code point
var 𐊧; // 𐊧 is U+102A7
var a = "\u{65}"; // var a = "\u0065";
var a = "\u{102A7}"; // var a = "\uD800\uDEA7";
```


data/non-ascii-identifier-part-only.js
```diff
@@ -1,5 +1,23 @@
...
+  astral: '1eyl,1fhc-9,1gjl-2,1gjp-1,1gjw-3,1gl4-2,1glb,1hq8-2,1hrs-e,1ht2-9,1hts-2,1hv4-a,1hww-9,1hxc-2,1hyf-d,1hyu-9,1i0w-2,1i2b-d,1i34-9,1j1n-c,1j28-9,20k1-19,20lr-3,2jxh-4,2jxp-5,2jy3-7,2jyd-6,2jze-3,2k3m-2,2l72-1d,jo5c-6n'
```
nicolo-ribaudo: Why I compressed this data:

|Size without astral (old size)|Size with astral          |Compressed size with astral   |
|------------------------------|--------------------------|------------------------------|
|`279kB` + `8kB` = `287kB`     |`635kB` + `12kB` = `647kB`|`3.37kB` + `1.93kB` = `2.30kB`|

src/lex.js
```diff
@@ -567,92 +590,108 @@ Lexer.prototype = {
...
-        if (asciiIdentifierPartTable[code] || isNonAsciiIdentifierPart(code)) {
-          index += 5;
-          return "\\u" + ch1 + ch2 + ch3 + ch4;
+      if (this.peek(index + 2) === "{") { // \u{...}
+        if (!state.inESNext(true)) {
+          this.trigger("warning", {
+            code: "W119",
+            line: this.line,
+            character: this.char + index,
+            data: [ "\\u{...}" ]
```
nicolo-ribaudo: What message should JSHint warn?
Is `'\u{...}' is only available in ES6 (use esnext option).` ok?

src/lex.js
```diff
@@ -567,92 +590,108 @@ Lexer.prototype = {
...
+        var cp = codePoint(ch2);
+        if (0xdc00 <= cp && cp <= 0xdfff) { // low surrogate
+          chr += ch2;
+          code = codePoint(chr);
+          if (!state.inESNext(true)) {
+            this.trigger("warning", {
+              code: "W119",
+              line: this.line,
+              character: this.char + index,
+              data: [ "Astral symbols in identifiers" ]
```
nicolo-ribaudo: How should the warning message be?
`'Astral symbols in identifiers' is only available in ES6 (use esnext option).` isn't obviously ok. (`symbols`...`is`)

src/lex.js
```diff
@@ -61,6 +59,31 @@ function asyncTrigger() {
   };
 }
 
+function fromCodePoint(codepoint) {
+  if (String.fromCodePoint) {
+    return String.fromCodePoint(codepoint);
```
nicolo-ribaudo: (Also line 77)

These lines are reported as not tested by coveralls (https://coveralls.io/builds/2635416/source?filename=src%2Flex.js#L64), because they never run on node (but `String.fromCodePoint` and `String.prototype.codePointAt` are supportd by iojs)
Is this a problem?

jugglinmike: You can avoid the coverage problem (and optimize the solution) by writing:

```js
/**
 * Implementation of ES6 `String.fromCodePoint` for legacy environments
 */
function fromCodePointShim(codepoint) {
  // (your function here, sans the initial `if` block)
}
var fromCodePoint = String.fromCodePoint || fromCodePointShim;
```

nicolo-ribaudo:
Rebased to include #2378 (because of travis-ci's failure)

src/lex.js
```diff
@@ -61,6 +59,36 @@ function asyncTrigger() {
...
+    var highSurrogate = (cp >> 10) + 0xd800;
+    var lowSurrogate = (cp % 1024) + 0xdc00;
+    return String.fromCharCode(highSurrogate, lowSurrogate);
+  }
+};
+
+/**
+ * Implementation of ES6 `String#codePointAt(0)` for legacy environments
+ */
+function codePoint(char) {
```
jugglinmike: This is a bit inconsistent. Why not define this function in the same way as `fromCodePoint` and avoid the Istanbul-specific workaround?

src/lex.js
```diff
@@ -660,19 +704,21 @@ Lexer.prototype = {
     }.bind(this);
 
     function removeEscapeSequences(id) {
-      return id.replace(/\\u([0-9a-fA-F]{4})/g, function(m0, codepoint) {
-        return String.fromCharCode(parseInt(codepoint, 16));
+      return id.replace(/\\u(?:([0-9a-fA-F]{4})|\{([0-9a-fA-F]{1,6})\})/g, function(m0, cp1, cp2) {
```
jugglinmike: This pattern is now probably too complicated to simply declare in-line. Mind moving it to `src/reg.js`?

src/lex.js
```diff
@@ -965,18 +1011,55 @@ Lexer.prototype = {
...
+      } else { // \uXXXX
+        hexCode = this.input.substr(1, 4);
+        escapeSequence += hexCode;
+        maxCode = 0xffff;
+        jump = 5;
+      }
+
       var code = parseInt(hexCode, 16);
-      if (isNaN(code)) {
+      if (!/^[0-9a-f]+$/i.test(hexCode) || code < 0 || code > maxCode) {
```
jugglinmike: This seems like a perfect place for your new `isHexNumber` function

jugglinmike:
This patch makes a faulty assumption: that unicode escape sequences using the bracket notation are limited to 6 hex digits. [The grammar for string literals is defined as `u{ HexDigits }`](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-literals-string-literals), and there is no restriction on the length of `HexDigits` (only the content). This means that the lexer should be fine with code like:

```js
// jshint esnext: true
var \u{0000000041};
```

Fixing this should also involve making the reported errors a little more helpful. For instance, the error on line 3 is misleading:

```js
var unclosedBrace = [
  "var a = '\\u{abc';",
  "var b = '\\u{abcs}';", // "s" isn't an hex digit
  "var c = '\\u{10abcde}';" // too long
];

var run = TestRun(test)
  .addError(1, "Expected '}' to match '{' from line 1 and instead saw '''.")
  .addError(2, "Expected '}' to match '{' from line 2 and instead saw 's'.")
  .addError(3, "Expected '}' to match '{' from line 3 and instead saw 'e'.");
run.test(unclosedBrace, { esnext: true });
```

Line 3 is definitely invalid, but only because [the MV of the HexDigits is greater than 1114111](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-string-literals-static-semantics-early-errors). When the restriction on pattern length is removed, that error should instead read, "Unexpected '\u{10abcde}'.".

Thanks for all your hard work so far!

nicolo-ribaudo:
Do you think it would be better a general `Invalid Unicode escape sequence: {a}` (where `{a}` can be `... is not a hex digit` or `too long`) error or to use two different errors?

nicolo-ribaudo:
Should JSHint throw `'{a}' is not a hexadecimal digit.` also while parsing the `\uXXXX` syntax?

lukeapage:
@jugglinmike commits to respond to your comments were done by @nicolo-ribaudo, in case you didn't notice ;)

src/reg.js
```diff
@@ -36,3 +36,6 @@ exports.fallsThrough = /^\s*\/\*\s*falls?\sthrough\s*\*\/\s*$/;
 // very conservative rule (eg: only one space between the start of the comment and the first character)
 // to relax the maxlen option
 exports.maxlenException = /^(?:(?:\/\/|\/\*|\*) ?)?[^ ]+$/;
+
+// Unicode escape sequences (\uXXXX and \u{XXXXXX})
+exports.unicodeEscapeSequence = /\\u(?:([0-9a-fA-F]{4})|\{([0-9a-fA-F]{1,6})\})/g;
```
jugglinmike: This RegExp has a hard-coded 6 from your initial solution. That gets in the way
of tracking references to variables; the following test fails:

    TestRun(test)
      .test("var \\u{00000078};\\u{78} = 9;", { esnext: true, unused: true });

Changing the `{1,6}` to a `+` should fix it, though!

jugglinmike:
I know the review for this one has been a little drawn out, but there are a lot
of details to consider and I'm working with limited availability. I spent some
time researching performance today, but it will be a little longer still before I
have anything constructive to say on that topic.

src/reg.js
nicolo-ribaudo: Fixed

nicolo-ribaudo:
@jugglinmike I've fixed the error you pointed out.

nicolo-ribaudo:
Rebased

Igor1201:
+1

mixed:
+1

Mithgol:
+1

Issue #1234, opened 2013-08-14
Please change jshint.js license, remove bad or evil


osallou:
Hi,
could you please modify the jshint.js license to be a real MIT license?

The "The Software shall be used for Good, not Evil." prevents library to be packaged in Debian and maybe other repo where code should really be free of use (yes, even evil). And honestly, if someone wants to use it for evil... well I don't think he will care about the license  :-)

Thanks

Olivier

piotr1212:
Thank you very much for working on this issue, I want to get this in Fedora. What exactly still needs to be done?

jugglinmike:
**The best way to confirm that we're still working on this** is to check [my `master-free` branch](https://github.com/jugglinmike/jshint/tree/master-free) -- I certainly wouldn't be keeping that in sync if I didn't intend to use it!

@piotr1212 Currently, we need to confirm the change with some of the previous contributors. It sounds like you're an RPM maintainer--is that correct?

piotr1212:
Thanks you for the update. I'll just wait paintently ;)
And yes, I am a package maintainer for Fedora/EPEL, There are many nodejs modules which depend on jshint and can't be packaged now...

paultag:
I'm with @piotr1212 - this has blocked some tools into Debian that I want to see. let's get this nonfree clause removed!

cburgmer:
I would like to promote this issue and call on all contributors to not forget our government employees inside the NSA or GCHQ that maintain parts of the pervasive surveillance system written in JavaScript. Clean code for everybody!

paultag:
![](http://2.bp.blogspot.com/-eO8o9_rI154/Tw4x1NQg0EI/AAAAAAAAAnk/QeUJsEsqgMg/s1600/gus-psyche-popcorn.gif)

foysavas:
I've signed the CLA and encourage others to do the same.

As a curiosity, I once had and took the opportunity to ask Howard Zinn to define evil. He, however, snapped back at me, claiming that if I didn't know what evil was by my age, then, for sure, I never would. After years of reflection on that statement, Zinn, of course, was right – and to be errantly good has remained the best I can hope for.

thejameskyle:
Is this license going to be updated as well? https://github.com/jshint/jshint/blob/master/src/jshint.js#L19

jugglinmike:
@thejameskyle That is the only license that needs changing. We'll be sure to update it as soon as we can

kreynen:
[...]

PR #2420, opened 2015-05-29
Update license in package.json

travishorn:
Following the latest spec at https://docs.npmjs.com/files/package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,12 +57,7 @@
     "unicode-6.3.0": "0.1.x"
   },
 
-  "licenses": [
-    {
-      "type": "MIT",
-      "url": "https://github.com/jshint/jshint/blob/master/LICENSE"
-    }
-  ],
+  "license": "MIT",
 
   "preferGlobal": true,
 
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the patch! It looks like the license currently documented in this project's `package.json` file is incorrect. The unfortunate reality is that [JSHint is licensed as MIT *and* JSON](https://github.com/jshint/jshint/blob/2444a0463e1a99d46e4afa50ed934c317265529d/src/jshint.js#L1-L29). We're still working to resolve this (see gh-1234), but in the mean time, we should ensure that the entry reflects reality.

Luckily, SPDX has [an identifier for the JSON license](http://spdx.org/licenses/JSON.html) and also supports a conjunction operator, so we can express this in a standards-compliant way:

    "license": "(MIT AND JSON)",

Would you mind updating the patch?

travishorn:
Sorry to see that so much time and discussion has been spent on licensing. Hopefully you are close to being able to switch completely to MIT.

jugglinmike:
You and me both.

Squashed and merged to `master` at commit e47168f5aa9576819830554f1b37dc0d49b76d31 . Thanks!

PR #2423, opened 2015-05-30
Fix infinite loop when parsing incomplete functions

grncdr:
See the unit test for the input that would previously cause an infinite loop. I

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2912,7 +2912,8 @@ var JSHINT = (function() {
           addlabel(ident, { type: "unused", token: state.tokens.curr });
         } else {
           // Skip invalid parameter.
-          while (!checkPunctuators(state.tokens.next, [",", ")"])) advance();
+          while (state.tokens.next.id !== "(end)" &&
+                 !checkPunctuators(state.tokens.next, [",", ")"])) advance();
         }
       }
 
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Infinite loop?
Using current master branch I haven't any infinite loop.

```
$ bin/jshint - <<< "function"
stdin: line 1, col 1, Missing name in function declaration.
stdin: line 1, col 1, Expected '(' and instead saw ''.
stdin: line 2, col undefined, Unrecoverable syntax error. (100% scanned).

3 errors
```

```
$ bin/jshint - <<< "function what ("
stdin: line 1, col 15, Expected an identifier and instead saw ''.
stdin: line 2, col undefined, Unrecoverable syntax error. (100% scanned).

2 errors
```

grncdr:
Nevermind, this was already fixed on master.

grncdr:
Cross-posted :smile: The issue is there on 2.7.0, seems like a common enough situation to warrant a point-release to npm in my mind.

caitp:
It has been a while since the last release, and there have been a number of fixes since then. We should really publish the next one soon

jugglinmike:
@caitp We're just blocked on gh-2404 . If you could give that a once-over, I'd be happy to release a new version.

Issue #1346, opened 2013-11-05
JSHint should complain (but not fail) when using deprecated options


pdehaan:
If I'm using a deprecated option in JSHint (ie: `nomen`, `onevar`, `passfail`, or `white`), JSHint should display a console warning or some other warning.

PR #2424, opened 2015-05-30
Warn when a deprecated option is used.

nicolo-ribaudo:


jugglinmike:
This is a nice idea, but it is a backwards-breaking change. We should revisit it for JSHint 3.0

lukeapage:
Fixes #1346

nicolo-ribaudo:
I think this pr can be close, as deprecated options will be removed in JSHint 3 (so this will be unuseful)

lukeapage:
@jugglinmike in v3 are we *just* giving deprecated warnings or actually removing those options completely?
@nicolo-ribaudo it has some value because it would be better to error/warn that the option doesn't exist and why than the generic invalid option..

jugglinmike:
> in v3 are we *just* giving deprecated warnings or actually removing those options completely?

There are many options that have been documented as "deprecated" for many months. We'll remove those completely in JSHint 3.0. Anything that we've only recently documented as deprecated (`predef` comes to mind, but there may be others) will trigger a notice but continue to function in the next major release.

PR #2433, opened 2015-06-01
[[FIX]] Don't report exported consts/lets as not defined.

### LINKED ISSUE #2430: export of const failing since 2.8.0
When exporting a variable that was declared as a ```const```, jshint fails:

```javascript
const FOO = 'foo';
export {FOO};
```

```
foo.js: line 2, col 9, 'FOO' is not defined.
```

This, however, works well:
```javascript
export const FOO = 'bar';
```

nicolo-ribaudo:
Fixes #2430

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4618,7 +4618,7 @@ var JSHINT = (function() {
         advance("(string)");
       } else if (ok) {
         exportedTokens.forEach(function(token) {
-          if (!funct[token.value]) {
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Fixed in #2435

lukeapage:
@nicolo-ribaudo I'll put your test changes into that PR, assuming you don't mind

PR #2434, opened 2015-06-01
Remove ArrayBufferView from globals

silverwind:
`ArrayBufferView` doesn't seem to exist in any environment, and it looks to be succeeded by `DataView`.


silverwind:
Updated to also remove it from tests.

silverwind:
Updated commit and PR title.

lukeapage:
Your correct it is not present in the browser.

https://html.spec.whatwg.org/multipage/infrastructure.html#arraybufferview
https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView

It is a type created for the spec, it does not exist as an interface or object in the browser.

lukeapage:
@jugglinmike I don't think this is breaking because it was never actually defined.. am I ok to rebase and push it?
Just checking because technically it is a breaking change.

jugglinmike:
It's definitely low-risk, but I also don't think there's any urgency here, so I'm happy to defer this for 3.0.

Issue #2429, opened 2015-05-31
for loops don't allow latedef


lukeapage:
The following warns that `i` is a global (line 1) and that `i` is unused from line 3.
```
for(i in {}) {
}
var i;
```

xak2000:
I recently updated JSHint (don't know which version was before) and it starts complain to this code:
```
var stompClient = {
      getClient: function () {return client;}
    },
    client;
```
`'client' was used before it was defined.`

This error is shown when `"latedef": "nofunc"` or `"latedef": true` and not shown when `"latedef": false`.

Before there was no error with the same options.
Don't know if it is the same case as `for` loop. If no I can open another issue.

nicolo-ribaudo:
http://jshint.com/docs/options/#latedef
> This option prohibits the use of a variable before it was defined. JavaScript has function scope only and, in addition to that, all variables are always moved—or hoisted— to the top of the function. This behavior can lead to some very nasty bugs and that's why it is safer to always use variable only after they have been explicitly defined.

> Setting this option to "nofunc" will allow function declarations to be ignored.

`client` is defined at line 4 but you are using it before, at line 2.

xak2000:
No, it is not used at line 2. It used in anonymous function defined at line 2. This anonymous function isn't called at line 2. Only defined.

If this were `getClient: client` or `getClient: (function () {return client;})()` then yes, `client` is used before defined.

Issue #2428, opened 2015-05-31
imports should be treated like const


lukeapage:
Imports are imported as const, but jshint adds them essentially as "vars" this means we do not detect re-assignments or mark as a warning if the import is used before imported.

I did have this change in my refactoring, but took it out to add as a seperate PR. will be easy to change.

example:
```
a = 1;
import {a} from 'a';
```

Issue #2426, opened 2015-05-30
no error when let or const use themselves during declaration


lukeapage:
I've inadvertantly fixed this already as part of work on other bugs.

example:
```
let a = a;
```

PR #2435, opened 2015-06-02
Refactor implied, unused, scope, global. Fix 2292, 2426, 2410, 2374 etc..

lukeapage:
Refactored scope and implied into a manager.
In the process, I've fixed alot of bugs and IMO made the code alot more readable
as well as making a chunk of code that could be taken out of jshint.js if
required.

bugs fixed / pr's obsolete (fixed by this)
https://github.com/jshint/jshint/issues/2439
https://github.com/jshint/jshint/issues/2426
https://github.com/jshint/jshint/issues/2410
https://github.com/jshint/jshint/issues/2409
https://github.com/jshint/jshint/issues/2374
https://github.com/jshint/jshint/issues/2430
https://github.com/jshint/jshint/issues/2368
https://github.com/jshint/jshint/issues/2342
https://github.com/jshint/jshint/issues/2363
https://github.com/jshint/jshint/issues/2292
https://github.com/jshint/jshint/issues/2282
https://github.com/jshint/jshint/pull/2199
https://github.com/jshint/jshint/issues/2191
https://github.com/jshint/jshint/issues/2090
https://github.com/jshint/jshint/issues/1996
https://github.com/jshint/jshint/issues/1775
https://github.com/jshint/jshint/pull/1682
https://github.com/jshint/jshint/issues/1462
https://github.com/jshint/jshint/issues/1061

further work - related issues not fixed
- doCatch could not use funct - use scope to see if we are in a catch
- members = [] is assigned but never used
- imports should be const - https://github.com/jshint/jshint/issues/2428
- incorrect for..in warning - https://github.com/jshint/jshint/issues/2429
- functions redeclared in the global scope - https://github.com/jshint/jshint/issues/2400
- unused when shadow declared twice - https://github.com/jshint/jshint/issues/2381
- I have part of a fix for https://github.com/jshint/jshint/issues/2261, see https://github.com/jshint/jshint/pull/2435#discussion_r31503940
- https://github.com/jshint/jshint/issues/1894 - can be fixed quite easily in new code base. kept existing behaviour

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -74,7 +74,6 @@ var JSHINT = (function() {
     },
 
     declared, // Globals that were declared using /*global ... */ syntax.
-    exported, // Variables that are used outside of the current file.
 
     functionicity = [
       "closure", "exception", "global", "label",
@@ -83,8 +82,6 @@ var JSHINT = (function() {
 
     functions, // All of the functions
 
-    global, // The global scope
-    implied, // Implied globals
     inblock,
     indent,
     lookahead,
@@ -93,7 +90,6 @@ var JSHINT = (function() {
     membersOnly,
     predefined,    // Global variables defined by option
 
-    scope,  // The current scope
     stack,
     unuseds,
     urls,
@@ -321,12 +317,6 @@ var JSHINT = (function() {
     };
   }
 
-  function isundef(scope, code, token, a) {
-    if (!state.ignored[code] && state.option.undef !== false) {
-      JSHINT.undefs.push([scope, code, token, a]);
-    }
-  }
-
   function removeIgnoredMessages() {
     var ignored = state.ignoredLines;
 
@@ -411,82 +401,10 @@ var JSHINT = (function() {
     return i;
   }
 
-  // adds an indentifier to the relevant current scope and creates warnings/errors as necessary
-  // name: string
-  // opts: { type: string, token: token, isblockscoped: bool }
-  function addlabel(name, opts) {
-
-    var type  = opts.type;
-    var token = opts.token;
-    var isblockscoped = opts.isblockscoped;
-
-    // Define label in the current function in the current scope.
-    if (type === "exception") {
-      if (_.has(state.funct["(context)"], name)) {
-        if (state.funct[name] !== true && !state.option.node) {
-          warning("W002", state.tokens.next, name);
-        }
-      }
-    }
-
-    if (_.has(state.funct, name) && !state.funct["(global)"]) {
-      if (state.funct[name] === true) {
-        if (state.option.latedef) {
-          if ((state.option.latedef === true && _.contains([state.funct[name], type], "unction")) ||
-              !_.contains([state.funct[name], type], "unction")) {
-            warning("W003", state.tokens.next, name);
-          }
-        }
-      } else {
-        if ((!state.option.shadow || _.contains([ "inner", "outer" ], state.option.shadow)) &&
-            type !== "exception" || state.funct["(blockscope)"].getlabel(name)) {
-          warning("W004", state.tokens.next, name);
-        }
-      }
-    }
-
-    if (state.funct["(context)"] && _.has(state.funct["(context)"], name) && type !== "function") {
-      if (state.option.shadow === "outer") {
-        warning("W123", state.tokens.next, name);
-      }
-    }
-
-    // if the identifier is blockscoped (a let or a const), add it only to the current blockscope
-    if (isblockscoped) {
-      state.funct["(blockscope)"].current.add(name, type, state.tokens.curr);
-      if (state.funct["(blockscope)"].atTop() && exported[name]) {
-        state.tokens.curr.exported = true;
-      }
-    } else {
-      state.funct["(blockscope)"].shadow(name);
-      state.funct[name] = type;
-
-      if (token) {
-        state.funct["(tokens)"][name] = token;
-      }
-
-      if (state.funct["(global)"]) {
-        global[name] = state.funct;
-        if (_.has(implied, name)) {
-          if (state.option.latedef) {
-            if ((state.option.latedef === true &&
-                _.contains([state.funct[name], type], "unction")) ||
-                !_.contains([state.funct[name], type], "unction")) {
-              warning("W003", state.tokens.next, name);
-            }
-          }
-
-          delete implied[name];
-        }
-      } else {
-        scope[name] = state.funct;
-      }
-    }
-  }
-
   function doOption() {
     var nt = state.tokens.next;
     var body = nt.body.split(",").map(function(s) { return s.trim(); });
+
     var predef = {};
     if (nt.type === "globals") {
       body.forEach(function(g, idx) {
@@ -534,7 +452,7 @@ var JSHINT = (function() {
           return;
         }
 
-        exported[e] = true;
+        state.funct["(scope)"].addExported(e);
       });
     }
 
@@ -873,7 +791,7 @@ var JSHINT = (function() {
       }
       isLetExpr = true;
       // create a new block scope we use only for the current expression
-      state.funct["(blockscope)"].stack();
+      state.funct["(scope)"].stack();
       advance("let");
       advance("(");
       state.tokens.prev.fud();
@@ -950,7 +868,7 @@ var JSHINT = (function() {
       }
     }
     if (isLetExpr) {
-      state.funct["(blockscope)"].unstack();
+      state.funct["(scope)"].unstack();
     }
 
     state.nameStack.pop();
@@ -1109,9 +1027,7 @@ var JSHINT = (function() {
         // detect increment/decrement of a const
         // in the case of a.b, right will be the "." punctuator
         if (this.right && this.right.identifier) {
-          if (state.funct["(blockscope)"].labeltype(this.right.value) === "const") {
-            error("E013", this, this.right.value);
-          }
+          state.funct["(scope)"].block.modify(this.right.value, this);
         }
       }
 
@@ -1336,15 +1252,11 @@ var JSHINT = (function() {
             warning("W121", left, nativeObject);
         }
 
-        if (predefined[left.value] === false &&
-            scope[left.value]["(global)"] === true) {
-          warning("W020", left);
-        } else if (left["function"]) {
-          warning("W021", left, left.value);
-        }
-
-        if (state.funct["(blockscope)"].labeltype(left.value) === "const") {
-          error("E013", left, left.value);
+        if (left.identifier) {
+          // reassign also calls modify
+          // but we are specific in order to catch function re-assignment
+          // and globals re-assignment
+          state.funct["(scope)"].block.reassign(left.value, left);
         }
 
         if (left.id === ".") {
@@ -1360,8 +1272,8 @@ var JSHINT = (function() {
         } else if (left.id === "[") {
           if (state.tokens.curr.left.first) {
             state.tokens.curr.left.first.forEach(function(t) {
-              if (t && state.funct[t.value] === "const") {
-                error("E013", t, t.value);
+              if (t && t.identifier) {
+                state.funct["(scope)"].block.modify(t.value, t);
               }
             });
           } else if (!left.left) {
@@ -1375,7 +1287,7 @@ var JSHINT = (function() {
           that.right = expression(10);
           return that;
         } else if (left.identifier && !isReserved(left)) {
-          if (state.funct[left.value] === "exception") {
+          if (state.funct["(scope)"].labeltype(left.value) === "exception") {
             warning("W022", left);
           }
           state.nameStack.set(left);
@@ -1448,9 +1360,7 @@ var JSHINT = (function() {
       // detect increment/decrement of a const
       // in the case of a.b, left will be the "." punctuator
       if (left && left.identifier) {
-        if (state.funct["(blockscope)"].labeltype(left.value) === "const") {
-          error("E013", this, left.value);
-        }
+        state.funct["(scope)"].block.modify(left.value, left);
       }
 
       this.left = left;
@@ -1583,7 +1493,7 @@ var JSHINT = (function() {
   }
 
   function statement() {
-    var i = indent, r, s = scope, t = state.tokens.next;
+    var i = indent, r, t = state.tokens.next, hasOwnScope = false;
 
     if (t.id === ";") {
       advance(";");
@@ -1611,7 +1521,9 @@ var JSHINT = (function() {
 
         advance("module");
         var name = identifier();
-        addlabel(name, { type: "unused", token: state.tokens.curr });
+        state.funct["(scope)"].addlabel(name, {
+          type: "var",
+          token: state.tokens.curr });
         advance("from");
         advance("(string)");
         parseFinalSemicolon();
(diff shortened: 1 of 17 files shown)


src/jshint.js
```diff
@@ -412,81 +402,13 @@ var JSHINT = (function() {
...
-      } else {
-        scope[name] = funct;
-      }
-    }
-  }
-
   function doOption() {
     var nt = state.tokens.next;
-    var body = nt.body.split(",").map(function(s) { return s.trim(); });
+    var body = nt.body
```
lukeapage: collides with PR #2386, hence my comment there. if that comment is fixed and PR pulled, I can remove these changes.

tests/unit/parser.js
```diff
@@ -3673,22 +3674,22 @@ exports["for of as esnext"] = function (test) {
     "for (const x = 1, y = 2 of [1,2,3,4]) print(x + y);"
   ];
   TestRun(test)
-    .addError(6, "Invalid for-of loop left-hand-side: initializer is forbidden.")
-    .addError(7, "Invalid for-of loop left-hand-side: more than one ForBinding.")
-    .addError(8, "Invalid for-of loop left-hand-side: initializer is forbidden.")
+    .addError(7, "Invalid for-of loop left-hand-side: initializer is forbidden.")
```
lukeapage: this is confusing as I haven't removed any errors, just moved the line number, because I added the `var xg, yg;` line.

jugglinmike: Since the errors are independent, you could declare the tests in a new test case

lukeapage: @jugglinmike not sure what you mean? Sorry! ;- I changed this test to avoid it outputting new errors that are unrelated to for-of, because the test cases all re-used the same variable names.

Are you suggesting I split the test up into multiple tests? I could, but its alot of grunt work..

jugglinmike: Nah, it's alright. That would be *more* noise, so it's fine as-is.

src/jshint.js
```diff
@@ -3518,6 +3303,30 @@ var JSHINT = (function() {
         }
       }
 
+      for (var t in tokens) {
+        if (tokens.hasOwnProperty(t)) {
+          t = tokens[t];
+          if (funct["(global)"]) {
+            if (predefined[t.id] === false) {
+              warning("W079", t.token, t.id);
```
nicolo-ribaudo: Needs tests.

https://coveralls.io/builds/2708395/source?filename=src%2Fjshint.js#L3311

src/jshint.js
```diff
@@ -4909,14 +4704,14 @@ var JSHINT = (function() {
...
-            isundef(funct, "W117", state.tokens.curr, v);
+            funct["(scope)"].current.use(v, state.tokens.curr);
             return true;
           // When we are in "filter" state,
           } else if (_current && _current.mode === "filter") {
             // we check whether current variable has been declared
             if (use(v)) {
               // if not we warn about it
-              isundef(funct, "W117", state.tokens.curr, v);
+              funct["(scope)"].current.use(v, state.tokens.curr);
```
nicolo-ribaudo: Needs tests.

https://coveralls.io/builds/2708395/source?filename=src%2Fjshint.js#L4714

src/jshint.js
```diff
@@ -5045,102 +4844,629 @@ var JSHINT = (function() {
...
+
+          _current["(usages)"][labelName]["(reassigned)"].push(token);
+        },
+
+        modify: function(labelName, token) {
+
+          _current["(usages)"][labelName]["(modified)"].push(token);
+        },
+
+        labeltype: function(labelName) {
```
nicolo-ribaudo: Needs tests.

https://coveralls.io/builds/2708395/source?filename=src%2Fjshint.js#L5354

lukeapage: this ones easy, the code is unused - now removed.

src/jshint.js
lukeapage: thanks fixed. I didn't actually add this code, just moved it within the function. While testing I found #2438 which is present before and after this PR is applied.

src/jshint.js
lukeapage: Added tests and discovered 2 more existing bugs (1 fixed, one not) - #2439 #2440

lukeapage:
note to self:
coverage tests to add
---------------------
https://coveralls.io/builds/2718305/source?filename=src%2Fjshint.js#L3723
 - no catch param
https://coveralls.io/builds/2718305/source?filename=src%2Fjshint.js#L3777
 - no catch block
https://coveralls.io/builds/2718305/source?filename=src%2Fjshint.js#L4154
 - continue X;
  - where X doesn't exist and does exist
https://coveralls.io/builds/2718305/source?filename=src%2Fjshint.js#L4711
 - array comprehension when in filter state, undef
https://coveralls.io/builds/2718305/source?filename=src%2Fjshint.js#L4961
 - can never be reached, remove return false;

lukeapage:
@jugglinmike @rwaldron @caitp this is ready for review, hope its to your liking, thanks.

jugglinmike:
Thanks, Luke! I'm hoping to give this a proper review this weekend.

src/jshint.js
```diff
@@ -2959,7 +2762,7 @@ var JSHINT = (function() {
       "(metrics)"   : null,
       "(statement)" : null,
       "(context)"   : null,
-      "(blockscope)": null,
+      "(scope)": null,
```
jugglinmike: I'm no fan of the object literal formatting style, either, but consistency trumps personal preference. Would you mind observing the whitespace convention here?

src/jshint.js
```diff
@@ -5032,115 +4831,625 @@ var JSHINT = (function() {
...
+        },
+
+        // adds an indentifier to the relevant current scope and creates warnings/errors as necessary
+        // name: string
+        // opts: { type: string, token: token, isblockscoped: bool }
+        addlabel: function(labelName, opts) {
+
+          var type  = opts.type;
+          var token = opts.token;
+          var isblockscoped = opts.isblockscoped;
```
jugglinmike: I think you can replace this with

```js
var isblockscoped = type === "let" || type === "const";
```

and then remove the `isblockscoped` option altogether.

src/jshint.js
```diff
@@ -5011,11 +4804,7 @@ var JSHINT = (function() {
     }
   }
 
-  var warnUnused = function(name, tkn, type, unused_opt) {
-    var line = tkn.line;
-    var chr  = tkn.from;
-    var raw_name = tkn.raw_text || name;
-
+  var getUnusedOption = function(unused_opt) {
```
jugglinmike: I like this refactoring, but it introduces yet another function that closes around the global `state` variable. A `state.getUnused` method in `state.js` would make the function's dependency on the `state` object much more obvious, and it would be a more discoverable location for the logic. What do you think?

src/messages.js
```diff
@@ -168,7 +169,6 @@ var warnings = {
   W089: "The body of a for in should be wrapped in an if statement to filter " +
     "unwanted properties from the prototype.",
   W090: "'{a}' is not a statement label.",
-  W091: "'{a}' is out of scope.",
```
jugglinmike: It looks like the convention here is to maintain properties for obsolete messages, but set them to `null`. I can only guess why (maybe to make it more obvious that the "skipped" message ID is intentional, maybe to make it easier to `git blame` the removal), but it's probably a good idea to stick with the convention for now.

src/jshint.js
```diff
@@ -5032,115 +4831,625 @@ var JSHINT = (function() {
...
-          if (item && item["(token)"]) {
-            item["(token)"].exported = true;
+      functionScope: {
+        /**
+         * Returns the label type given certain options
+         * @param labelName
+         * @param {Object=} options
+         * @param {Boolean=} options.onlyBlockscoped - only include block scoped labels
+         * @param {Boolean=} options.excludeParams - exclude the param scope
+         * @param {Boolean=} options.excludeCurrent - exclude the current scope
```
jugglinmike: `@returns {string}` ?

src/jshint.js
```diff
@@ -74,7 +74,6 @@ var JSHINT = (function() {
     },
 
     declared, // Globals that were declared using /*global ... */ syntax.
-    exported, // Variables that are used outside of the current file.
```
jugglinmike: I love the global abatement!

src/jshint.js
```diff
@@ -5032,115 +4831,625 @@ var JSHINT = (function() {
...
-          return null;
+
+          if (token && (state.ignored.W117 || state.option.undef === false)) {
+            token.ignoreUndef = true;
+          }
+
+          _addUsage(labelName, token);
+        },
+
+        reassigned: function(labelName, token) {
```
jugglinmike: `reassigned` is a good property name, but the method should be named `reassign`

src/jshint.js
```diff
@@ -5032,115 +4831,625 @@ var JSHINT = (function() {
...
+        add: function(labelName, type, tok, unused) {
+          _current["(labels)"][labelName] = {
+            "(type)" : type,
+            "(token)": tok,
+            "(blockscoped)": false,
+            "(unused)": unused };
         }
       },
 
       current: {
```
jugglinmike: The name `current` makes sense--`funct["scope"].current` accurately describes the object as the current scope. The `functionScope` property muddies this a little bit--given `funct["scope"].current`, it's not so clear that `funct["scope"].functionScope` is the *current* function scope... or that `current` is a block scope specifically.

What do you think about

- `funct["scope"].current` -> `funct["scope"].currentBlock`
- `funct["scope"].functionScope` -> `funct["scope"].currentFunct`

or (since that's a lot of typing):

- `funct["scope"].current` -> `funct["scope"].block`
- `funct["scope"].functionScope` -> `funct["scope"].funct`

?

jugglinmike:
This looks really nice, @lukeapage! Many times I caught myself writing a comment about some inconsistency, only to find that you were just preserving the existing pattern. Obviously, it's important to stay focused during a big refactoring like this, and I just wanted to say that I appreciate your restraint :)

There's one change I would like to request, but I'd really like @caitp's and @rwaldron's support before asking you to go through with it:

`jshint.js` is a giant file. I feel very strongly that the more code we can write in other files, the more maintainable and approachable the source becomes. This is why I moved the options configuration to `options.js` and created the `name-stack.js` file when implementing function name inference. Since you are contributing a good chunk of new work, I would very much like to see it contained in an independent source file. The transformation would require about 5 steps; the first three are trivial:

1. Cut-and-paste the `scopeManager` factory function definition into a new file
2. Include lodash in that file
3. Update the factory function to accept a reference to the `state` object (and update the invocations to match)

Next, you'd have to use an event emitter to communicate warnings and errors. This is consistent with how the lexer (the other major external stateful object) works (see [these lines in `jshint.js` for example](https://github.com/jshint/jshint/blob/7aa3e23e7e7bc53d1b19da25e669b30b6f75c5f5/src/jshint.js#L5324-L5330)), and I expect it would be pretty straightforward.

The last change is to redefine the `funct` variable in `jshint.js` to instead exist as an attribute of `state`. Even outside of this refactoring, that seems like the better location for `funct`. I'm describing it last, but it could (should?) come first; I would be happy to do this in a separate pull request if you would be game for resolving the conflicts it produced.

But before we go down that road, I'd like to see what @caitp and @rwaldron have to say. I've made a big deal about it because of that last step, but I don't really think it's too dangerous. Would you two agree?

src/jshint.js
lukeapage: yes I can now. when we were adding break labels to the same object, they got true too. I think the only other one is the catch param, but that gets added through stackParams, so will do this.

src/jshint.js
lukeapage: yep - both `getUnusedOption` and `warnUnused` are now only used within scope manager, so they can both be moved in there.

src/messages.js
lukeapage: yes, I took my lead from W092 which is missing (https://github.com/jshint/jshint/commit/9b8b363b04112c6cb38ad200124826d1ff0fc37d) but your right, later removals are null. Have changed to null.

src/jshint.js
lukeapage: well spotted, fits with modify too

src/jshint.js
lukeapage: yes, current made more sense when it was just a block scope manager. Will take your suggestion.
One thing that I wondered was whether funct["scope"] should be just scope? Its the same object all the way through - but this is how blockScope was set up - is it to reduce the number of variables available throughout the jshint file ?

lukeapage:
> I appreciate your restraint

Thanks, given the size of it already I was worried I'd done too much as it was.

> I feel very strongly that the more code we can write in other files, the more maintainable and approachable the source becomes

I agree, this was my end intention. I figured it might be best in another PR - but I could do it here if you like. What do you mean by conflicts produced from it? Presumably if it was on top of this one, there wouldn't be any conflicts?

I've fixed all the other comments now.

jugglinmike:
> I figured it might be best in another PR - but I could do it here if you
> like. What do you mean by conflicts produced from it? Presumably if it was on
> top of this one, there wouldn't be any conflicts?

I agree that it would be good to do that in another PR, but I would like to see
that done in advance of landing this. The order of operations matters to me
because I would like to avoid landing your code in the [tainted](https://github.com/jshint/jshint/blob/7aa3e23e7e7bc53d1b19da25e669b30b6f75c5f5/src/jshint.js#L4-L5) `jshint.js`.

I'm not sure how important that is to you, so I wanted to see just how
difficult my request would be. I've opened gh-2464 which moves `funct` to
`state` in today's `master`. If we land that refactoring, you can rebase your
work and automatically resolve conflicts by preferring your changeset in all
cases:

    $ git rebase -s recursive -X theirs master

This will leave you with a patch that uses `funct` for lines that you
added/modified, and `state.funct` for lines that you did not touch. You can
then use the same simple transformation that I used to update your patch:

    $ sed -i 's/\([^\.]\)funct\b/\1state.funct/g' src/jshint.js

That operation is a little naive, but the JSHint and JSCS tasks that run during
`npm test` will show you the locations that need to be manually fixed (not
many--basically `functor`, `hasParsedCode`, and a few additional occurences).

Finally, you can force-push your rebased work to this branch to update the pull
request. All in all, this should take less than 10 minutes. Would you be down
for that?

If you'd like to try it before making a decision, you can:

    $ git checkout -b tmp fix-unused
    $ git fetch jugglinmike

...and then use `jugglinmike/move-funct` in place of `master` in the above `git
rebase` command.

lukeapage:
> I agree that it would be good to do that in another PR, but I would like to see
> that done in advance of landing this.

I may as well do it in this PR then? - if that is the case I'll rebase-squash the commits to far into 1 (or would you prefer some history left?) or do you mean you'd prefer *just* #2464 done in advance of this?

Either way I'm happy for you to go ahead and pull #2464, I don't see any reason why your suggestion wouldn't work.

I'll rebase squash everything so far and create a good commit message, then I'll do what you suggest, then I'll work on a 2nd commit message (to go into the master history) that pulls everything out to a seperate file.. ok?

jugglinmike:
I'd prefer just gh-2464 in advance of this. Reason being: that change makes sense as a standalone commit as a "pure" refactoring that does not modify functionality. If instead, we merged it into your patch here, this pull request would end up more lines changed, but none of those new changes would actually mean anything for the new functionality itself. Basically, I'd like to take this opportunity to avoid making your patch larger since the `funct`->`state.funct` refactoring makes sense on its own.

Coveralls has been unresponsive all day, so I merged gh-2378 myself (which I typically hate to do). I've also merged gh-2386 since you reviewed it for me and it's useful for this patch. Note that this has repercussions for the `-s recursive -X theirs` advice I gave above. You'll need to rebase this branch onto `master` *before* landing gh-2464, since the current conflicts require human intervention.

After that, we can land gh-2464. This will introduce conflicts for this patch once again, but then you can use the steps I described above to simplify the process of resolution.

I know all this is a little onerous, but this is an important patch, so I want to get it right!

lukeapage:
> You'll need to rebase this branch onto master before landing gh-2464

done.

lukeapage:
@jugglinmike I have now rebased this on top of master (and therefore #2464)

PR #2441, opened 2015-06-03
Add `mocha` to `mocha` environment

BYK:
Mocha exposes the global name `mocha` too for configuration etc. This adds that to the `mocha` environment.


jugglinmike:
Thanks!

PR #2445, opened 2015-06-04
[[FIX]] Add `HTMLCollection` to browser environment.

nicolo-ribaudo:


TheSavior:
:+1:

jugglinmike:
@nicolo-ribaudo I'd like to move the property higher in the object initializer to maintain the (semi) alphabetical ordering. I'm happy to do this in the process of landing your commit, but I wanted to tell you here so you would have chance to do it yourself (or object) before I do. Let me know!

nicolo-ribaudo:
Travis-ci failed because of coveralls (#2378), can you restart it, please?

lukeapage:
I restarted but coveralls is still down. if you rebase on master, that has a commit in that allows travis to pass even when coveralls fails..

jugglinmike:
Thanks!

Issue #1914, opened 2014-10-21
Docs bug: globals and predef are both missing at http://www.jshint.com/docs/options/


paulvi:
`globals` and `predef` keywords are both missing at http://www.jshint.com/docs/options/ .

While http://www.jshint.com/docs/ shows `predef` in example.

see also http://stackoverflow.com/questions/22551402/difference-between-globals-and-predef-in-jshintrc

jugglinmike:
These variables have since been included in http://www.jshint.com/docs/options/ . Thanks for the report (and sorry about the delay)!

PR #2451, opened 2015-06-06
[[DOCS]] Document `predef` option

jugglinmike:
Explain the behavior and usage of the option and denote its deprecated
status.

Resolves gh-1914

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -757,7 +757,18 @@ exports.val = {
    */
   maxerr       : false,
 
-  predef       : false, // predef is deprecated and being replaced by globals
+  /**
+   * This option allows you to control which variables JSHint considers to be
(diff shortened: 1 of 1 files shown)


lukeapage:
given what you learnt in #2453, does this need adjusting? Do you still want to deprecate the option? I guess there is no problem deprecating it if we mention that we are deprecating where it can be replaced with globals, or if #2453 is merged first?

jugglinmike:
I suppose that realistically, we can't deprecate this option if the alternative is not viable today. I've removed the deprecation note

lukeapage:
committed: https://github.com/jshint/jshint/commit/47bd5ca197bb103cfeb4a7cdaf54136fb379c937

PR #2453, opened 2015-06-08
Remove globals

duncanbeevers:
The `predef` option allowed us to blacklist globals otherwise considered valid. Since this option is being deprecated, this PR restores this functionality by allowing the specification of blacklisted as well as whitelisted globals.

Whitelisted globals continue to act as normal while blacklisted globals are indicated as they were in the `predef` array, with a leading hyphen sigil.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -171,6 +171,11 @@ var JSHINT = (function() {
   function combine(dest, src) {
     Object.keys(src).forEach(function(name) {
       if (_.has(JSHINT.blacklist, name)) return;
+      var match = name.match(/^\-(.*)/);
+      if (match) {
+        JSHINT.blacklist[match[1]] = src[name];
+        return;
+      }
       dest[name] = src[name];
     });
   }
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -50,6 +50,26 @@ exports.testUnusedDefinedGlobals = function (test) {
   test.done();
 };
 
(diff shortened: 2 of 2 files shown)


jugglinmike:
Ah, so `predef` cannot be entirely replaced by `globals`. Thanks for pointing this out!

Since we're talking about a breaking change, I would like to consider an alternative that is more obtrusive for migration but (in my opinion) better in the long-term. What if we changed the meaning of option values for globals:

- `true` - pre-defined and writable (no change)
- `"readonly"` - pre-defined and non-writable (i.e. the effect of the `false` value today)
- `false` - black listed (i.e. the effect of `predef` with a `"-"` prefix today)

To demonstrate, here's a configuration for JSHint v2.x:

```json
{
  "predef": ["-performance", "require"],
  "globals": {
    "jQuery": false,
    "myApp": true
  }
```

...and here's how that would have to be re-written according to my proposal:

```json
{
  "globals": {
    "jQuery": "readonly",
    "myApp": true,
    "performance": false,
    "require": true
  }
}
```

I think this is far more readable and intuitive. We just have to weigh that against the migration hassle for existing users--`false` will have to be changed to `"readonly"` in all cases. What do you think?

duncanbeevers:
This sounds like a good way forward.

jugglinmike:
Alright, great! I've opened gh-2477 to describe our preferred solution.

PR #2463, opened 2015-06-12
[[FIX]] Add `window.performance` to browser vars

nicolo-ribaudo:
Spec: http://www.w3.org/TR/navigation-timing/#sec-window.performance-attribute


jugglinmike:
Thanks!

PR #2464, opened 2015-06-13
[[FIX]] Declare `func` as a property of `state`

jugglinmike:
The current `funct` object stores stateful information, so it should be
defined as a property of the global variable reserved for that purpose.
This allows other code to more simply operate based on the current
state because it requires one less reference to be managed across
contexts.


lukeapage:
LGTM

jugglinmike:
Thanks, @lukeapage!

PR #2465, opened 2015-06-13
Update package.json

paladox:
Updating per https://david-dm.org/jshint/jshint

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,14 +34,14 @@
   "main": "./src/jshint.js",
 
   "dependencies": {
-    "cli":                 "0.6.x",
+    "cli":                 "0.8.x",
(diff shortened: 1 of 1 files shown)


lukeapage:
The update to jscs breaks the build, so this cannot be merged.

paladox:
Ok i have changed it. Please reopen

lukeapage:
I'm still not sure about this - it seems a bit strange just to update our dependencies - to update dependencies you need to have a working knowledge of the code base and then read the update notes for each module to see if there is anything changed, before making the update. If I wanted to "review" this, I'd have to do all that work (and squash 3 commits) rather than the simpler method of just updating them myself..

Also, looks like you haven't pushed.

Issue #2409, opened 2015-05-22
out of scope error not given if used in sub function


lukeapage:
e.g.
```
function test() {
  if (true) {
    var a;
    var b;
  }
  a = 1;
  function test2() {
    b = 1;
  }
}
```

`a` errors - use, out of scope
`b` does not error

I'm working on this as part of the variable work - unless someone thinks the behaviour is correct.

jugglinmike:
This issue was automatically closed when gh-2468 landed in `master` because of a reference in one of the commit messages. [A later commit in that same changeset reverted the bug fix](https://github.com/lukeapage/jshint/commit/b376de7c71a3c43a5a0e622dff4ca8f43858479f) because it is technically a backwards-breaking change. As such, I'm manually re-opening this issue and labeling it for the next major release of JSHint.

lukeapage:
added to v3 milestone since the work has already been done for it.

jugglinmike:
Right on, I realized I forgot this one after I signed off last night.

PR #2468, opened 2015-06-16
Scope manager - extract to external file

lukeapage:
@jugglinmike 

This PR is based on #2435 and pulls scope manager into a seperate file. Since #2435 is rebased to a single commit, this could be pulled as-is instead of that PR if you wish.


jugglinmike:
Awesome, @lukeapage! I'm really looking forward to reviewing this version. Forgive me, though: it may be a few days before I have the opportunity.

tests/regression/thirdparty.js
```diff
@@ -156,21 +159,22 @@ exports.lodash_0_6_1 = function (test) {
...
     .addError(1122, "'isArr' used out of scope.")
     .addError(1127, "'className' used out of scope.")
     .addError(1129, "Use '===' to compare with 'true'.")
     .addError(1153, "'isArr' used out of scope.")
     .addError(1159, "'isArr' used out of scope.")
     .addError(1490, "Use '===' to compare with '0'.")
+    .addError(1504, "'noNodeClass' used out of scope.")
     .addError(1670, "Missing semicolon.")
-    .addError(2731, "'array' is already defined.")
-    .addError(2732, "'array' is a statement label.")
```
jugglinmike: This is awesome

tests/unit/core.js
```diff
@@ -793,8 +827,9 @@ exports.testConstRedeclaration = function (test) {
   ];
 
   TestRun(test)
-      .addError(2, "const 'a' has already been declared.")
-      .addError(9, "const 'a' has already been declared.")
+      .addError(2, "'a' has already been declared.")
+      .addError(9, "'a' has already been declared.")
+      .addError(13, "'b' has already been declared.")
```
jugglinmike: Very nice

tests/unit/parser.js
```diff
@@ -2253,7 +2252,7 @@ exports["let statement in for loop as legacy JS"] = function (test) {
     "for (let i = 0; i<15; ++i) {",
     "  print(i);",
     "}",
-    "for (let i=i ; i < 10 ; i++ ) {",
+    "for (let i=0 ; i < 10 ; i++ ) {",
```
jugglinmike: Bizarre

jugglinmike:
Hey @lukeapage, this time I focused on the tests. I have two pieces of feedback
based on what I learned there.

First, treating unreferenced `globals` as unused is a breaking change. While
this could be a great way for users to avoid accumulating unnecessary
configuration, I'm not so sure about including outside of a major release.

The second is a lot trickier. JSHint's default behavior is to treat `var`
statements as block scoped... kind of. Based on modifications to existing tests
(the so-called "regression" tests in particular), it looks like your patch
makes this behavior more consistent. Whenever I see new errors added to
existing tests, I get a little nervous because it means the change is
backwards-incompatable. In most cases, this is fine because your patch is
correctly identifying invalid JavaScript--I'm prepared to defend the decision
to issue new warnings for code that was never valid to begin with. But this is
a little trickier for two reasons: it's an opinionated rule, and it is enabled
by default. This means that many users may not even be aware that it is being
enforced, so I have more sympathy if their code starts to fail lint.

To me, the only way to reconcile that second point is to defer landing this
until we're ready for 3.0 . Do you agree? If so, I'd be happy to create a
`v3.0.0` feature branch (I think it's high time we got started on that anyway)
and make this the first patch to land there. But maybe I'm being too
conservative (or even misinterpreting the changes I'm seeing), so don't
hesitate to tell me that, either!

lukeapage:
> First, treating unreferenced globals as unused is a breaking change. While
> this could be a great way for users to avoid accumulating unnecessary
> configuration, I'm not so sure about including outside of a major release.

I've been looking through my test changes and I don't see any change to unreferenced globals being unused.
At the moment this causes 2 warnings
```
/* jshint unused: true */
/* globals a */
var b;
```
and it still produces 2 warnings after my change.

lukeapage:
> The second is a lot trickier. JSHint's default behavior is to treat var
> statements as block scoped... kind of. Based on modifications to existing tests
> (the so-called "regression" tests in particular), it looks like your patch
> makes this behavior more consistent.

I think its this bugfix in particular: https://github.com/jshint/jshint/issues/2409

There were lots of
```
try {
   var foo = window.foo.bar;
}
catch(e) {
   foo = polyfill;
}
function() {
  foo();
}
```

> Whenever I see new errors added to
> existing tests, I get a little nervous because it means the change is
> backwards-incompatable. In most cases, this is fine because your patch is
> correctly identifying invalid JavaScript--I'm prepared to defend the decision
> to issue new warnings for code that was never valid to begin with. But this is
> a little trickier for two reasons: it's an opinionated rule, and it is enabled
> by default. This means that many users may not even be aware that it is being
> enforced, so I have more sympathy if their code starts to fail lint.

I can undo this particular bug fix.. I'll look at how hard that would be, I'm hoping it's quite easy. then I could label the bugfix as for jshint 3?

There are alot of changes here and given the current pace of development, I can't see jshint 3 coming particularly soon..

lukeapage:
> I can undo this particular bug fix..

I've added code so that I don't add warnings. I recommend we keep this in a separate commit, branch a jshint 3 branch and then revert that commit.

If you are happy, I'd like to see this land asap so it can be worked on top of

lukeapage:
Hold on, I made a mistake, will comment again once I've corrected the commit

lukeapage:
Okay corrected problem with 8cf4234cec743e261015b59e041449e98f880424 - thats the one that needs undoing on a 3 branch.

I also found a bug in my work which is fixed in the next commit  - 53b24706644092305a51e2e7166c8fadbdea3496

lukeapage:
rebased again on jshint/master as I merged in a couple of things that conflicted

jugglinmike:
@lukeapage Thank you for all your hard work and patience! This is a really positive step forward for the project :D

lukeapage:
Thanks for reviewing.. I know it wasn't an easy one

PR #2078, opened 2015-01-04
bugfix: exported generator functions

sebv:
Prevents 'unused' warning to be thrown on exported generator function.

PR #2473, opened 2015-06-18
[[FIX]] Correctly parse exported generators

### LINKED ISSUE #2472: Export generator function from ES6 module
ES6 modules syntax allows to have named exports. In case of export generator function jshint throws `is defined but never used` error, but it should not.

### Steps to reproduce

Use `jshint@2.8.0` with default configuration + per file config.
Function export just works.

```js
// jshint unused:true, esnext:true
export function add(a, b) {
  return a + b;
}
```

Generator export throws an error.

```js
// jshint unused:true, esnext:true
export function* add(a, b) {
  yield a + b;
}
```

```
$ jshint -v
jshint v2.8.0
$ jshint export-function.js 
$ jshint export-generator.js 
export-generator.js: line 2, col 18, 'add' is defined but never used.
1 error
```

nicolo-ribaudo:
Before of this patch, JSHint considered the name of expored
generators `*`, because it was the first "thing" after `function`.

Fix #2472

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3817,7 +3817,8 @@ var JSHINT = (function() {
     checkProperties(props);
   }
 
-  blockstmt("function", function() {
+  blockstmt("function", function(context) {
+    var inexport = context && context.inexport;
     var generator = false;
     if (state.tokens.next.value === "*") {
       advance("*");
@@ -3835,6 +3836,8 @@ var JSHINT = (function() {
 
     if (i === undefined) {
       warning("W025");
+    } else if (inexport) {
+      exported[i] = true;
     }
 
     // check if a identifier with the same name is already defined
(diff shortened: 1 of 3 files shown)


src/jshint.js
```diff
@@ -4662,9 +4665,8 @@ var JSHINT = (function() {
       // ExportDeclaration :: export Declaration
       this.block = true;
       advance("function");
-      exported[state.tokens.next.value] = ok;
       state.tokens.next.exported = true;
-      state.syntax["function"].fud();
+      state.syntax["function"].fud({ inexport:true });
```
lukeapage: I've been comparing with #2078 and this seems a better way of doing it, since the export doesn't have to know about the structure of the function, good.

lukeapage:
can generators be used with export default ? It looks like they can, since..
```
HoistableDeclaration[Yield, Default] :
    FunctionDeclaration[?Yield,?Default] 
    GeneratorDeclaration[?Yield, ?Default]
```

That's handled in the export code above your change (it seems to allow any expression, which is possibly another bug)

Would you mind looking at that too?

The rest of the change looks good to me.

lukeapage:
> (it seems to allow any expression, which is possibly another bug)

the comment confusingly misses
```
export default [lookahead  { function, class }] AssignmentExpression[In] ;
```

lukeapage:
I've added the export default comment to another PR and also checked that generator functions already work with export default and I've already covered the confusing aspects of the export default code, so I don't think there is anything wrong with this and it LGTM.

lukeapage:
committed - https://github.com/jshint/jshint/commit/06048169e73647c704c7c8d75362d5e65d904db9

PR #2474, opened 2015-06-19
Support `--prereq` CLI options

### LINKED ISSUE #1685: ERROR: Unknown option --prereq

lukeapage:
cherry picked original PR - https://github.com/jshint/jshint/pull/1909
Resolved comments 
 - from @mikesherov - whitespace formatting - it now seems this is the default so it matches the file
 - from @mikesherov - using null instead of "" - change made
 - from @jugglinmike - I don't think its a valid comment, I've replied on the original PR
I've reviewed but I'm not familiar with the cli code so will wait for someone else to say okay before merging.

Since [the docs][1] mention it, and it's already supported via [config][2],
I thought I'd just pop the CLI support in.

This PR make `jshint` allow specifying the `--prereq` option on the command
line, accepting a comma-separated list of paths.

Test-related notes:

- Fixes the existing `testPrereq` test to check the right exit code. In its
  current incarnation, it could never fail (even when `lint` didn't pass)
- Added 2 tests for specifying the `--prereq` options from the command line.

Implementation notes:

I did the simplest thing possible (just merging the CLI `prereq` options in
with the config ones in the `run` method in `cli.js`). It's a little bit
awkward because there are 2 paths to `lint`:

1. `opts.useStdin`
2. `!opts.useStdin`

And each loads the config in a slightly different way. So the `mergeCLIPrereq`
function has to be in 2 code paths. So, slightly awkward, but not too terrible.

Fixes #1685.

[1]: http://jshint.com/docs/cli
[2]: https://github.com/jshint/jshint/blob/master/src/cli.js#L451-458

Conflicts:

	src/cli.js
	tests/cli.js

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -15,6 +15,13 @@ var defReporter       = require("./reporters/default").reporter;
 var OPTIONS = {
   "config": ["c", "Custom configuration file", "string", false ],
   "reporter": ["reporter", "Custom reporter (<PATH>|jslint|checkstyle|unix)", "string", undefined ],
+  "prereq": [
+    "prereq",
+    "Comma-separate list of prerequisite (paths). E.g. files which include" +
+    "definitions of global variabls used throughout your project",
+    "string",
+    null
+  ],
   "exclude": ["exclude",
     "Exclude files matching the given filename pattern (same as .jshintignore)", "string", null],
   "exclude-path": ["exclude-path", "Pass in a custom jshintignore file path", "string", null],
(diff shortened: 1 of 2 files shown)


jugglinmike:
Looks good to me! @satchmorun would you mind signing the JSHint CLA? It should only take a minute:

https://www.clahub.com/agreements/jshint/jshint

lukeapage:
@satchmorun please could you sign the CLA ? We cannot merge this without and it is very simple

satchmorun:
@lukeapage @jugglinmike CLA signed.

jugglinmike:
Applied to `master` at commit 81fe3fa780a1d8797f9f1bb1fac07ad5fedadd75 . Thanks to you both!

PR #2476, opened 2015-06-20
Bump lodash version

roderickhsiao:


lukeapage:
@jugglinmike how do you handle dependency updates? I've read through the notes and don't see anything scary. At the same time there are several packages now out of date, not just lodash and lodash 3.10 has just been released.

roderickhsiao:
lodash 4 is out, we probably want to move to that,

PR #2478, opened 2015-06-21
Add test fixture saved with leading Unicode BOM

Krinkle:
See pull #2285. For issue #2284.


jugglinmike:
This test does not demonstrate the underlying issue because the "BOM-accepting" behavior is limited to JSHint's CLI. I'm proposing we remove that inconsistency by *not* filtering out the BOM in https://github.com/jshint/jshint/issues/2486 .

PR #2487, opened 2015-06-26
Fix labeled break and continue semantics

tstarling:
A labeled break may refer to any statement, it is not required to be inside a loop or switch. Whereas a labeled continue must refer to a loop, you cannot continue a switch.

In ECMAScript terminology, the identifier following "continue" must appear in the label set of an enclosing *IterationStatement* (do, while, for). The identifier following "break" must appear in the label set of an enclosing *Statement* (which includes a simple *Block*). http://www.ecma-international.org/ecma-262/5.1/#sec-12.7

Added tests.

I am using labeled breaks in code automatically generated by PEG.js, in my pending PEG.js pull request: https://github.com/pegjs/pegjs/pull/348 . Its automated test suite is failing due to jshint. If I had goto, I could have generated even faster code, but you make do with what you've got, right? ;)

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4111,20 +4111,19 @@ var JSHINT = (function() {
   stmt("break", function() {
     var v = state.tokens.next.value;
 
-    if (state.funct["(breakage)"] === 0)
-      warning("W052", state.tokens.next, this.value);
-
     if (!state.option.asi)
       nolinebreak(this);
 
-    if (state.tokens.next.id !== ";" && !state.tokens.next.reach) {
-      if (state.tokens.curr.line === startLine(state.tokens.next)) {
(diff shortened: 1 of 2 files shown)


lukeapage:
Nice,  looks good so far,  thanks. 

You reference es5 - any idea if things changed between 3 and 5? Just wondering if the es3 option has any relevance...

tstarling:
It is the same in ES3, it has the exact same wording of the relevant sections. According to the Mozilla JavaScript reference, the labeled break/continue feature was introduced in ES3.

lukeapage:
committed - https://github.com/jshint/jshint/commit/da66f70217fc7bcea0b0211bcc45476a33be7022

Thanks! Will be in the next release (after current 2.8.0)

Issue #2488, opened 2015-06-27
Parse error on JSON file with strings that contain semicolons


iwehrman:
```
$ cat test.json
{
    "a": ";"
}

$ jshint test.json 
test.json: line 2, col 5, Expected an assignment or function call and instead saw an expression.
test.json: line 2, col 8, Missing semicolon.
test.json: line 2, col 8, Expected '}' to match '{' from line 1 and instead saw ':'.
test.json: line 2, col 10, Expected an assignment or function call and instead saw an expression.
test.json: line 2, col 13, Missing semicolon.
test.json: line 2, col 1, Unrecoverable syntax error. (50% scanned).

6 errors

$ jshint --version
jshint v2.8.0
```

caitp:
This will be a big where we need to use checkPunctuators rather than just testing the value of the token,i believe. Should be an easy first big if anyone's interested

jugglinmike:
Yup, I'm halfway there :)

caitp:
I said *first* bug :p

nicolo-ribaudo:
Now this can be closed

lukeapage:
This will be released in the next version of jshint, after 2.8.0

PR #2489, opened 2015-06-27
[[FIX]] Allow semicolon as string value in JSON

jugglinmike:
Resolves gh-2488

Commit message:

> Previously, any token with the value ";" would trigger JSHint to lint
> the input as though it were JavaScript code. Because the token value may
> also describe a string, this caused JSHint to incorrectly interpret JSON
> input as JavaScript in some cases.
>
> Ensure that only semicolon tokens parsed as punctuators trigger
> JavaScript parsing.
>
> This change necessitated modifying an existing test. That test was
> authored to ensure that JSHint would halt on invalid input and report
> the input as invalid. The modified version of the test continues to
> assert that, but the specific errors have changed because JSHint now
> interprets the input as invalid JSON (not invalid JavaScript).

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4511,7 +4511,7 @@ var JSHINT = (function() {
           break;
         }
       }
-      if (pn.value === ";") {
+      if (checkPunctuators(pn, [";"])) {
         ret.isBlock = true;
         ret.notJson = true;
       }
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -877,6 +877,17 @@ exports.json.errors = function (test) {
   test.done();
 }
 
+// Regression test for gh-2488
+exports.json.semicolon = function (test) {
+  TestRun(test)
+    .test("{ \"attr\": \";\" }");
+
+  TestRun(test)
+    .test("[\";\"]");
+
+  test.done();
+};
+
 exports.comma = function (test) {
   var src = fs.readFileSync(__dirname + "/fixtures/comma.js", "utf8");
 
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -643,7 +643,8 @@ exports.ownProperty = function (test) {
   test.done();
 };
 
-exports.jsonMode = function (test) {
+exports.json = {};
+exports.json.dflt = function (test) {
   var code = [
     '{',
     '  a: 2,',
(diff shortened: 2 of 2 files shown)


caitp:
LGTM, but we do have a utility function for that

jugglinmike:
Yeah, but it really irks me to create a new array literal every time. Maybe consistency trumps my unsupported insistence on "efficiency" in JavaScript.

jugglinmike:
@caitp I took your advice, and thought you might be interested: in my original patch

```diff
@@ -4511,7 +4511,7 @@ var JSHINT = (function() {
           break;
         }
       }
-      if (pn.value === ";") {
+      if (pn.type === "punctuator" && pn.value === ";") {
         ret.isBlock = true;
         ret.notJson = true;
       }
```

`pn.type` will never be "punctuator". ["(punctuator)", on the other hand...](https://github.com/jshint/jshint/blob/a4f4bf4319186480778f7b75ef74382a88428e99/src/jshint.js#L4614)

caitp:
Yeah I don't have that stuff committed to memory. I think 'id' might be the one we want, I'd check the helper function to be sure

jugglinmike:
Oh, I was just saying that I used the wrong value, and that the error would have been avoided if I had just used the utility function in the first place. It should be good now

nicolo-ribaudo:
> Yeah, but it really irks me to create a new array literal every time

[You can use a string](https://github.com/lodash/lodash/blob/master/lodash.js#L6547):
```js
checkPunctuators(pn, ";");
```

lukeapage:
It would be inconsistent and somone might use it with e.g. `... ` where it would appear to work. Better to modify the function to take multiple arguments? You might be able to use indexof and skip the first item in arguments to make it elegant iniside the check function too..

caitp:
if it's a string, it should really be just a `return type === '(punctuator)' && id === theParticularPunctuatorWeCareAbout` which is a pretty trivial change to make in the helper

lukeapage:
will merge and make a new pr for @caitp's suggestion

lukeapage:
committed https://github.com/jshint/jshint/commit/0bda42297e435d16b9a244be72b69b1462077d5d and https://github.com/jshint/jshint/commit/ab73e010677ad6c34e21b8854fd5a43f7f1ca766

PR #2491, opened 2015-06-28
Change Imports to be const and change multiple warnings on block scope redeclaration to one warning

### LINKED ISSUE #2428: imports should be treated like const

lukeapage:
As per the es6 spec, treat the variables declared via import as
constants, meaning they cannot be re-assigned and cannot be declared
twice.
Fixes #2428 

Also fixed the following which I noticed. included in the same PR because there would be a merge conflict in the tests..

[[FIX]] Dont warn twice in block var redeclaration
When a block scoped variable is declared twice in the same scope it is a
syntax error. If the variable is used between those declarations jshint
should not warn about use before declared and error on re-declaration, so
remove the use before declared warning in this case.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4300,7 +4300,7 @@ var JSHINT = (function() {
       // ImportClause :: ImportedDefaultBinding
       this.name = identifier();
       state.funct["(scope)"].addlabel(this.name, {
-        type: "var",
+        type: "const",
         token: state.tokens.curr });
 
       if (state.tokens.next.value === ",") {
@@ -4325,7 +4325,7 @@ var JSHINT = (function() {
       if (state.tokens.next.identifier) {
         this.name = identifier();
         state.funct["(scope)"].addlabel(this.name, {
-          type: "var",
+          type: "const",
           token: state.tokens.curr });
       }
     } else {
@@ -4349,7 +4349,7 @@ var JSHINT = (function() {
         }
 
         state.funct["(scope)"].addlabel(importName, {
-          type: "var",
+          type: "const",
           token: state.tokens.curr });
 
         if (state.tokens.next.value === ",") {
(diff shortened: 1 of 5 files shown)


lukeapage:
[[Edit]] refactored function to be clearer (all is block scope under one block).
Also noticed 2.8.0 does not warn about use before declaration with function scoped variable, so added test case and fixed that too.

jugglinmike:
@lukeapage: This looks good! I wasn't aware that imported bindings were treated as constants; that's a nice behavior. Would you mind if I amended 89ba0ee142989741b36a1c528df812ae0d5fad93 to add a reference to the spec to the calls to `state.funct`?

Reference: http://www.ecma-international.org/ecma-262/6.0/#sec-createimportbinding:

> **8.1.1.5.5 CreateImportBinding (N, M, N2)**
>
> The concrete Environment Record method CreateImportBinding for module
> Environment Records creates a new initialized immutable indirect binding for
> the name N. A binding must not already exist in this Environment Record for
> N. M is a Module Record (see 15.2.1.15), and N2 is the name of a binding that
> exists in M’s module Environment Record. Accesses to the value of the new
> binding will indirectly access the bound value of value of the target
> binding.
>
>     Let envRec be the module Environment Record for which the method was invoked.
>     Assert: envRec does not already have a binding for N.
>     Assert: M is a Module Record.
>     Assert: When M.[[Environment]] is instantiated it will have a direct binding for N2.
>     Create an immutable indirect binding in envRec for N that references M and N2 as its target binding and record that the binding is initialized.
>     Return NormalCompletion(empty).

lukeapage:
> Would you mind if I amended 89ba0ee to add a reference to the spec to the calls to state.funct?

That sounds good, please go ahead

jugglinmike:
Merged to `master` at commits 94a677902af35efffb4f2388c6039968ed3adb2a and e32e17b97289e4bb5a53116ffc3a979356792088. Thanks, Luke!

Issue #2455, opened 2015-06-10
typeof is not safe in TDZ


lukeapage:
I believe this does not error after my PR which does TDZ detection.

http://es-discourse.com/t/why-typeof-is-no-longer-safe/15

```
let b = typeof a;
let a;
```

lukeapage:
verified, this is already fixed on master

PR #2499, opened 2015-06-29
[[TEST]] Test typeof in TDZ

lukeapage:
Test that you cannot use typeof on a block scoped variable in the TDZ
Verifies #2455


jugglinmike:
Awesome--thanks for following through on this. Merged to `master` at commit d5885da6daa82fab2e99eb5ac2eb5c800844507d

PR #2500, opened 2015-06-29
[[FIX]] Classes are not hoisted

### LINKED ISSUE #1934: ES6 class is not hoisted – should error
This should error, because `Foo` is not hoisted. JSHint currently does not error.

``` js
new Foo();

class Foo { constructor() {} }
```

lukeapage:
Classes are not hoisted and have a TDZ like let and const.
In addition logic around when a class/function name is exposed has been
tidied up so that function/class expressions do not have their name made
available outside the expression.
Fixes #1934


src/jshint.js
```diff
@@ -4417,14 +4424,10 @@ var JSHINT = (function() {
 
       expression(10);
 
-      if (state.tokens.next.id === "class") {
```
lukeapage: to explain this change, this was a typo - `state.tokens.next` is at this point the thing after an export. This made the following code fail
```
/* jshint undef: true */
export default class A{}
class B{}
A();
```

talltyler:
Awesome, please merge!

lukeapage:
rebased on master

src/jshint.js
```diff
@@ -2894,6 +2894,19 @@ var JSHINT = (function() {
 
     functions.push(state.funct);
 
+    // So that the function is available to itself and referencing itself is not
+    // seen as a closure, add the function name to a new scope, but do not
+    // test for unused (unused: false)
+    // it is a new block scope so that params can override it, it can be block scoped
+    // but declarations inside the function don't cause already declared error
+    var internallyAccessibleName = name || classExprBinding;
+    var internallyAccessibleNameType = name ? "function" : "class";
```
jugglinmike: It looks like we don't need `internallyAccessibleNameType`. Normally I prefer naming values with variables because they serve as documentation, but this one would tend to leak in strange ways--it would be defined as `"class"` even in cases where this was no name binding.

src/messages.js
```diff
@@ -95,7 +95,7 @@ var warnings = {
   W018: "Confusing use of '{a}'.",
   W019: "Use the isNaN function to compare with NaN.",
   W020: "Read only.",
-  W021: "'{a}' is a function.",
+  W021: "'{a}' is a {b}.", // is a class or is a function
```
jugglinmike: I know you're just extending the warning, but while we're updating this message, why don't we make it something slightly more user-friendly. Saying "C is a class" doesn't really explain why it is a warning or how to fix it. What about

> Reassignment of '{a}', which is is a '{b}'. Use "var" or "let" to declare bindings that may change.

jugglinmike: Formatting fixes:

> Reassignment of '{a}', which is is a {b}. Use 'var' or 'let' to declare bindings that may change.

jugglinmike: I just realized that a change like that would require updating a bunch of unrelated unit tests and add a lot of noise to your patch. But it seems as though we have zero tests for reassignment of function bindings. I've opened gh-2564 to address this

src/jshint.js
lukeapage: okay its gone and I switched round the expression.

src/messages.js
lukeapage: Thanks, it is present in the regressions of two libraries, but yes no unit tests.

I've changed the message (your suggestion is alot better, I noticed it wasn't very informative but wasn't sure it was a good idea to change it). I've done it in a seperate [[CHORE]] commit - upto you if you want to squash it or not.

jugglinmike:
Excellent! Squashed and merged as two separate commits: 

- 87378cca3c98e1ff4d5be11f6c16a9f24a212e3e
- 3e55c3542f40669f858ddc3257492d8cdafed160

Thanks, Luke!

PR #2501, opened 2015-06-29
[[FIX]] block scope vars dont redefine in blocks

### LINKED ISSUE #2438: incorrect redefinition of global warning with block scoped code when in block
The following warns
```
{
  let Map = true;
}
```
But I guess it shouldn't. The code checks `if (funct["(global)"]) {` but should perhaps be located into scope manager and call `_currentFunct["(global)"] && _current === _currentFunct` for block scoped variables.

lukeapage:
Do give a redefinition error when a block scope declaration is inside a
block.
Fixes #2438

### DIFF
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -645,7 +645,7 @@ var scopeManager = function(state, predefined, exported, declared) {
        * @returns Boolean
        */
       isGlobal: function() {
-        return _currentFunct["(global)"] && _current === _currentFunct;
+        return _current["(global)"];
       },
 
       use: function(labelName, token) {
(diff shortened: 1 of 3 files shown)


src/scope-manager.js
```diff
@@ -643,6 +639,15 @@ var scopeManager = function(state, predefined, exported, declared) {
     },
 
     block: {
+
+      /**
+       * is the current block global?
+       * @returns Boolean
+       */
+      isGlobal: function() {
+        return _currentFunct["(global)"] && _current === _currentFunct;
```
jugglinmike: How about just `return _current["(global)"];`?

lukeapage: much better, that completely escaped me.

lukeapage:
updated with code review comments

jugglinmike:
Squashed and merged to `master` at commit 9e7402581f95d1605d5dd54f312cd61d8ba0880f. Nice going, Luke!

Issue #1920, opened 2014-10-23
2.5.6 breaks for in loops with switch


laurens:
Version 2.5.6 of jshint breaks with `for…in` loops that contain a `switch`:

`````
for (var key in objects) {
  if (!objects.hasOwnProperty(key)) {
    switch (key) {
    case 'foo':
      if (true) {
        hey();
      }
      break;
    case 'bar':
      if (true) {
        hey();
      }
    }
  }
}

````


This will result in the following error:

`````
TypeError: Cannot read property 'type' of undefined
    at Object.oldScope [as fud] (/Users/pnch/github/jshint/src/jshint.js:3734:38)
    at expression (/Users/pnch/github/jshint/src/jshint.js:1045:32)
    at statement (/Users/pnch/github/jshint/src/jshint.js:1730:9)
    at statements (/Users/pnch/github/jshint/src/jshint.js:1773:16)
    at block (/Users/pnch/github/jshint/src/jshint.js:1899:13)
    at Object.blockstmt.labelled [as fud] (/Users/pnch/github/jshint/src/jshint.js:4075:11)
    at expression (/Users/pnch/github/jshint/src/jshint.js:1045:32)
    at statement (/Users/pnch/github/jshint/src/jshint.js:1730:9)
    at statements (/Users/pnch/github/jshint/src/jshint.js:1773:16)
    at itself (/Users/pnch/github/jshint/src/jshint.js:5001:9)
````

laurens:
Thx, @lukeapage

PR #2503, opened 2015-06-29
[[FIX]] Do not crash if the forin check is block

lukeapage:
Stops the crash reported in gh-1920.

I am not 100% sure on the outcome of the conversation since code is now missing, so instead of removing checks, I have just stopped the crash which was the main concern.


src/jshint.js
```diff
@@ -3744,7 +3744,7 @@ var JSHINT = (function() {
     // When the if is within a for-in loop and the condition has a negative form,
     // check if the body contains nothing but a continue statement
     if (forinifcheck && forinifcheck.type === "(negative)") {
-      if (s && s.length === 1 && s[0].type === "(identifier)" && s[0].value === "continue") {
+      if (s && s[0] && s[0].type === "(identifier)" && s[0].value === "continue") {
```
jugglinmike: I'm all for defensive programming, but the real problem here is that `statement` is returning `undefined` for `switch` statements. Would you mind address that, too?

This also means you can simplify the test case--any `switch` statement will do.

lukeapage: thanks, done both suggestions

jugglinmike:
Squashed and merged to `master` at commit d1cbe84cbfe9ac49d7a684f1c8ccde3cf89efaed. Rock on

PR #2509, opened 2015-07-01
[[FIX]] Do not parse arrays which contain `for` as array comprehensions.

nicolo-ribaudo:


src/jshint.js
```diff
@@ -4540,13 +4540,14 @@ var JSHINT = (function() {
...
-    var pn, pn1;
+    var pn, pn1, prev;
     var i = -1;
     var bracketStack = 0;
     var ret = {};
     if (checkPunctuators(state.tokens.curr, ["[", "{"]))
       bracketStack += 1;
     do {
+      prev = pn || state.tokens.curr;
       pn = (i === -1) ? state.tokens.next : peek(i);
```
lukeapage: how about ?
```
if (i === -1) {
    prev = state.tokens.curr;
    pn = state.tokens.next;
} else {
   prev = pn;
   pn = peek(i);
}
```
or making both the inline conditionals the same?
```
prev = i === -1 ? state.tokens.curr : pn;
```
at the moment they use a different method to determine is first and they are reversed conditionals!

nicolo-ribaudo: I prefer the inline conditionals

lukeapage:
squashed and committed - https://github.com/jshint/jshint/commit/d70876c6cbdd560930f03f2f20fefd8d2fa12d62

PR #2511, opened 2015-07-02
[[FIX]] Allow object destructuring assignment

### LINKED ISSUE #2269: Object destructoring variable reuse
```js
let {start, stop} = node;
// ...
({start, stop}) = othernode;
```

Throws 2 errors:
  - bad assignment
  - Expected an assignment not an expression

Sorry if its a dupe, only glanced through issues

------

It's valid according to the folks at babel

[Repl example](https://babeljs.io/repl/#?experimental=true&playground=true&evaluate=true&loose=false&spec=false&code=let%20%7Bstart%2C%20stop%7D%20%3D%20node%3B%0A%2F%2F%20...%0A(%7Bstart%2C%20stop%7D)%20%3D%20othernode%3B)

lukeapage:
Rather than parse an array/object literal and then pull out the
identifiers, both the array and object literal parsing now detect a
destructuring assignment and use the destructuring expression to make the
parsing consistent.
Fixes #2269 

This requires a fix to ignores for all the unit tests to pass, but I will include that in a separate PR that must be pulled before this!

Alternatively, temporarily disable the one ignore test that is failing.


lukeapage:
Depends on https://github.com/jshint/jshint/pull/2512

lukeapage:
rebased on master. this now passes tests.

src/jshint.js
```diff
@@ -1282,20 +1282,22 @@ var JSHINT = (function() {
...
-        } else if (left.id === "[") {
-          if (state.tokens.curr.left.first) {
-            state.tokens.curr.left.first.forEach(function(t) {
-              if (t && t.identifier) {
-                state.funct["(scope)"].block.modify(t.value, t);
-              }
+        } else if (left.id === "{" || left.id === "[") {
+          if (state.tokens.curr.left.destructAssign) {
+            state.tokens.curr.left.destructAssign.forEach(function(t) {
+              state.funct["(scope)"].block.modify(t.id, t.token);
```
lukeapage: t.id can be null in the case of `,,,`

lukeapage:
rebased on master

jugglinmike:
I think it's important that we match the spec language here. Let's make the following changes:

- warning message from "destructuring expression" to "destructuring binding"
- internal function name from `destructuringExpression` to `destructuringPattern` (since it is used for both destructuring binding and destructuring assignment)

Otherwise, this is great! Much simpler than I would have expected.

jugglinmike:
Ideally, this new code path would be used in `ForIn/Of` statements, since that is a destructuring assignment, not a destructuring binding. The current reality is [a whole lot uglier](https://github.com/jshint/jshint/blob/561d9785cdca88fc0a3178795274207520d05596/src/jshint.js#L4121-L4123). My initial thought was that approach would break for

```js
let x;
for ([x] of [[1]]) {}
```

for the same reason `let x; var x;` fails. Somehow, it works though. So at the moment, the only practical difference I can see here is that the warning message generated in pre-ES6 envronments will inaccurately refer to that production as "destructuring binding". Even if there are more severe implications, that problem is independent from your change. I'll file a bug for that when this lands--just thought you might be interested.

lukeapage:
> Let's make the following changes:
> 
> warning message from "destructuring expression" to "destructuring binding"
> internal function name from destructuringExpression to destructuringPattern (since it is used for both destructuring binding and destructuring assignment)

I've made both changes as seperate commits.

Note that the message text wasn't mine - it is what is currently shown, so it generated alot of test changes. I'm fine with that though if you are.

lukeapage:
> Ideally, this new code path would be used in ForIn/Of statements, since that is a destructuring assignment, not a destructuring binding. The current reality is a whole lot uglier.

Wow yes, I missed that. I tested it worked but didn't track down why/how it worked

> Somehow, it works though.

The `implied` goes all the way through to the `varstatement`. I thought this might fail

```
var a;
for(a,b in a) {
}
```

but it doesn't - then I realised thats because it [prescans for commas](https://github.com/jshint/jshint/blob/561d9785cdca88fc0a3178795274207520d05596/src/jshint.js#L4082), ignoring destructuring then [warns before it gets to the varstatement](https://github.com/jshint/jshint/blob/561d9785cdca88fc0a3178795274207520d05596/src/jshint.js#L4107)

I'm surprised there isn't a bug in there - it would be worth refactoring at some point.

jugglinmike:
> Note that the message text wasn't mine - it is what is currently shown, so it generated alot of test changes. I'm fine with that though if you are.

Definitely more noise, but now seems like a good time to correct that. Thanks for using separate commits; that made verification much easier!

Squashed and merged to `master` at ae4896670a46247ef59df9d3ea9cf3814c864559. Nice work :)

Issue #2215, opened 2015-02-27
jshint ignore:start/end not working as of 2.6.1


shelbys:
We have the above wrapped around a very long line, and we're now getting a `jshint.W101` `Line is too long` warning, whereas we weren't in 2.6.0

rwaldron:
Thanks for the report. Please provide the exact code that's causing this issue

shelbys:
We have `"maxlen": 640` in `.jshintrc`

```
        /* jshint ignore:start */
[...]

Issue #1792, opened 2014-08-08
Options being ignored if it's the first line of a function.


cxmcc:
a.js:
```
(function () {
var a = 1;
/* jshint ignore:start */
var b = 2
/* jshint ignore:end */
})();
```
"jshint a.js" works fine (No error.)

b.js:
```
(function () {
/* jshint ignore:start */
var b = 2
/* jshint ignore:end */
})();
```
"jshint b.js":
```
jshint b.js
b.js: line 3, col 10, Missing semicolon.

1 error
```

jugglinmike:
Fixed via gh-2427

Issue #1465, opened 2014-01-08
ignore doesn't work in some cases


clee704:
The following code gives you a warning.

    (function () {
      function foo() {
        /* jshint ignore:start */
        @@workload
        /* jshint ignore:end */
      }
      foo();
    })();

Strangely, the warning disappears when there are odd number of newlines between the top level function and function `foo`. So for example, with the following code:

    (function () {

      function foo() {
        /* jshint ignore:start */
        @@workload
        /* jshint ignore:end */
      }
      foo();
    })();

You see no errors.

This is confirmed with v2.4.1.

lukeapage:
re-tested with master after latest improvements (https://github.com/jshint/jshint/commit/09716080b4e49c6ab6b96c29351a9f816c3e6419) to ignore and it is still an issue - though some of these cases might be fixed.

jugglinmike:
This was a tricky one! The test cases provided in the comments on this issue were resolved via gh-2427 as @lukeapage mentioned, but it wasn't until we landed gh-2512 that the originally-reported example was fixed.

Anyway, we're all better now! Expect the fix in the next release of JSHint (whatever follows 2.8.0). Thanks for the help, everyone!

PR #2512, opened 2015-07-02
Fix ignore

lukeapage:
Continuation of #1967

commits need squashing, left in to see what I did

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -601,15 +601,12 @@ var JSHINT = (function() {
 
         if (key === "ignore") {
           switch (val) {
-          case "start":
-            state.ignoreLinterErrors = true;
-            break;
-          case "end":
-            state.ignoreLinterErrors = false;
-            break;
           case "line":
             state.ignoredLines[nt.line] = true;
             removeIgnoredMessages();
+            break;
+          case "start":
+          case "end":
             break;
           default:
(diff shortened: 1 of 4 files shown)


jugglinmike:
This is really nice. What do you think about this simplification? https://github.com/jugglinmike/jshint/commit/bbcdf38bdcb496e60caeebeef28a99b5cea6b1e3

lukeapage:
yes please go ahead and squash that in when you merge

lukeapage:
just tested and all of the examples in #1465 are fixed by this PR

lukeapage:
and #2215 is also fixed by this

lukeapage:
and #1792 is also fixed by this

jugglinmike:
> yes please go ahead and squash that in when you merge

In this case, we have multiple authors collaborating on a single feature. Each author's changeset incrementally defines the fix, so it would be inappropriate to simply cherry pick the commits. At the same time, squashing hides the history in a way that is disingenuous. The individual commits may not be useful on their own, but it's not quite correct to say that you or Cait or I were the sole author. So although we generally, discourage merge commits for landing patches, I think this is a perfect circumstance!

Merged to `master` at commit 39709b1e4c611973394ae7c3f3e1f6f989d15301

PR #2514, opened 2015-07-02
[[CHORE]] Tidy up checkPunctuators to not require array literal

lukeapage:
As discussed on #2489 remove the requirement for an array literal

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1446,9 +1446,9 @@ var JSHINT = (function() {
       }
       advance();
 
-      if (checkPunctuators(state.tokens.next, ["..."])) {
+      if (checkPunctuator(state.tokens.next, "...")) {
         warning("E024", state.tokens.next, "...");
-        while (checkPunctuators(state.tokens.next, ["..."])) {
+        while (checkPunctuator(state.tokens.next, "...")) {
           advance();
         }
       }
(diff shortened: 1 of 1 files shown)


jugglinmike:
I like the thought, but the name `checkPunctuators` doesn't really make sense when you are checking a single token against a single punctuator value. What do you think about defining a separate `checkPunctuator` function?

lukeapage:
Change made. I rebased because it involved changing every line.

lukeapage:
rebased on master

jugglinmike:
Hey Luke: this no longer applies cleanly, but you've already had to rebase twice (two times too many for a patch like this), so I've rebased this time. Given the straightforward nature of this patch, the resolution was trivial, but I still feel weird modifying someone else's patch without their knowledge. If you can verify that https://github.com/jugglinmike/jshint/commit/850c19317c6b45b1723c7ef7ffd8e7ac2a0c9fae looks good to you, I'll land it right away.

lukeapage:
yep, looks fine.

jugglinmike:
Great! 850c193 is now in `master`. Thanks for following up on my gripe :) One day, I'd like to have a `Token` class (and corresponding source file) with methods like this attached to it.

PR #2517, opened 2015-07-02
[[FIX]] Warn when shadowing in the global scope

### LINKED ISSUE #1019: function redefinition does not trigger warning
In doing code review, I just ran across a place where we defined a function twice in a row.  It turns out that jshint 1.1.0 doesn't notice this, and I couldn't find an option to suggest that it do so. 

Having that functionality, perhaps even on by default, would help catch a new class of bugs.

### LINKED ISSUE #1304: Warn when variable and functions are defined more than once.
E.g.

``` javascript

var a = 1;
var a = 2; // Warn here

function foo() {}
function foo() {} // Warn here
```

Should be opt-in of course.

lukeapage:
Previously when shadow was set to inner or outer, no warnings were given
about redeclaring in the global scope. This adds this warning, which in
many cases will be a clear mistake.
Fixes #1019, Fixes #1304 

You may want to class this as a breaking change and put into v3? I think its a clear bug that people should be happy to hear about, but it would match the decision on the other scope related bug.


lukeapage:
rebased and fixed commented out code.

cpirich:
+1 would love to see this get merged

joel-airspring:
+1

I just encountered a duplicate variable at the global scope and since there was no warning, I wasted a bit of time trying to to fix my .jshintrc.  (Surely duplicate variables would be caught!)  But detecting this case isn't supported yet.

Could we get this feature merged in?

Issue #1939, opened 2014-10-30
Experimental Traceur features/ES7


OliverJAsh:
One experimental/ES7 feature supported through Traceur that I want to start using today is [async/await](jakearchibald.com/2014/es7-async-functions/).

What is JSHint’s stance on supporting experimental Traceur features i.e. ES7, e.g. `async/await`?

IanVS:
Hi, I read through a few pull requests and issues regarding support of ES7 async/await, but it seems the conversation has died off.  Has there been any other thinking or progress towards support in some shape or form?

abrkn:
:+1:

jugglinmike:
Nothing's been discussed off-line--the conversation on GitHub is the extent of it

IanVS:
Do I understand correctly then that async/await won't be handled (supported or ignored) until a plugin system is developed?

jugglinmike:
Please remember that the async/await is a proposal for a far-future version of the language and not a specification. Beyond the initial development effort, extending JSHint core with support for this syntax would be a burden on both the maintainers and the consumers. Changes to the proposal will need to be reflected in the user's platform (whether a runtime or a transpiler), the user's code, and also in JSHint itself. It would be hard enough for JSHint to synchronize updates with platform changes, but the problem becomes intractable when you recognize that the userbase will upgrade incrementally. (Proposal changes might seem unlikely, but we're even today struggling with last-minute modifications to ES6's module syntax--see gh-2019 and gh-2036 for a couple examples.)

This is why prior to standardization, the healthiest way for JSHint to support this syntax would be through a plugin system. Once this feature approaches standardization, I expect JSHint's core will be extended to support it natively (just as is happening with ES6 features today).

@rwaldron @caitp @mikesherov If this reasoning reflects your thoughts, I think it would be best to close this issue.

sebv:
How long for the plugin system to emerge?

CrabDude:
I certainly appreciate how a request to support async/await may feel like a burden on the JSHint contributors, but I just wanted to provide my perspective and nothing more to encourage support in spite of @jugglinmike's stated concerns:

With the rise in popularity of transpilers (with support for async/await) like traceur and especially babel, the flood of increasingly standardized ESNext features (further encouraging the use of transpilers), the proliferation of promises, and the present support for generators in node.js and io.js (which makes async/await sugar for function* instead of a hack), async/await will see a dramatic increase in usage in the near future.

[...]

Issue #1148, opened 2013-06-20
Change EcmaScript levels to a single flag "ecma"


nschonni:
Separating out the discussion from https://github.com/jshint/jshint/issues/1137#issuecomment-19694125 although I've changed 'eslevel' to 'ecma'

This would allow the current flags to be folded into one:
- es3 -> ecma: 3
- es5 -> ecma: 5
- esnext -> ecma: 6? or 'next'
- moz -> ecma: 'moz'

lukeapage:
merging to #2124

PR #2519, opened 2015-07-02
`esversion` option

nicolo-ribaudo:


src/jshint.js
```diff
@@ -4247,12 +4305,12 @@ var JSHINT = (function() {
     x.lbp = 25;
   }(prefix("yield", function() {
     var prev = state.tokens.prev;
-    if (state.inESNext(true) && !state.funct["(generator)"]) {
+    if (state.inES6(true) && !state.funct["(generator)"]) {
```
lukeapage: generator functions aren't in 6.. but I guess the plan for this is for v2.x.. meaning it cannot break?
Still might be good to differentiate *now* e.g. esnext means esversion = 7 and this requires es7?

src/messages.js
```diff
@@ -181,7 +181,7 @@ var warnings = {
   W101: "Line is too long.",
   W102: null,
   W103: "The '{a}' property is deprecated.",
-  W104: "'{a}' is available in ES6 (use esnext option) or Mozilla JS extensions (use moz).",
+  W104: "'{a}' is available in ES6 (use 'esversion: 6') or Mozilla JS extensions (use moz).",
```
lukeapage: and this obviously also needs adjusting with the right version?

src/jshint.js
```diff
@@ -4648,13 +4706,13 @@ var JSHINT = (function() {
...
   function destructuringAssignOrJsonValue() {
-    // lookup for the assignment (esnext only)
+    // lookup for the assignment (ECMAScript 6 only)
     // if it has semicolons, it is a block, so go parse it as a block
     // or it's not a block, but there are assignments, check for undeclared variables
 
     var block = lookupBlockType();
     if (block.notJson) {
-      if (!state.inESNext() && block.isDestAssign) {
+      if (!state.inES6() && block.isDestAssign) {
```
lukeapage: wdyt about inES(version) as opposed to a set of functions?

src/messages.js
lukeapage: We also need to decide what to do about the moz function.. in v3 will it be there? will it be removed? if it will be removed, then this PR is doing the right thing

tests/unit/options.js
```diff
@@ -2930,3 +2930,71 @@ exports.module.declarationRestrictions = function( test ) {
...
+
+
+  TestRun(test, "precedence over `es3`") // TODO: Remove in JSHint 3
+    .test(es6code, { esversion: 6, es3: true });
+
+  TestRun(test, "precedence over `es5`") // TODO: Remove in JSHint 3
+    .addError(0, "ES5 option is now set per default")
+    .test(es6code, { esversion: 6, es5: true });
+
+  TestRun(test, "precedence over `esnext`") // TODO: Remove in JSHint 3
```
lukeapage: not sure what the precedence is for TODO's, a core team member probably has an opionion

lukeapage:
Not all the comments are for you, some are just things that probably need "deciding" by someone (so probably @jugglinmike). I wouldn't do anything based on them yet till he responds and its too risky a changeset for me to consider merging, so I'll be leaving it to him.
Thanks though - its a step forward.

src/jshint.js
```diff
@@ -2589,11 +2647,11 @@ var JSHINT = (function() {
   prefix("[", function() {
     var blocktype = lookupBlockType();
     if (blocktype.isCompArray) {
-      if (!state.inESNext()) {
+      if (!state.inES6()) {
         warning("W119", state.tokens.curr, "array comprehension");
```
lukeapage: same here, not ES6

src/jshint.js
nicolo-ribaudo: > generator functions aren't in 6

Do you mean *generator comprehensions*?

> Still might be good to differentiate now e.g. esnext means esversion = 7 and this requires es7?

If you mean the *esnext option*, it would be a breaking change...

jugglinmike:
> We also need to decide what to do about the moz function.. in v3 will it be there? will it be removed? if it will be removed, then this PR is doing the right thing

We're going to keep `moz`, but as proposed in gh-1148, we should deprecate the dedicated option and include a "moz" value for `esversion`. Today, it can be taken to mean "ES6 plus Mozilla extensions". @nicolo-ribaudo would you be up for that in this patch?

(Moving forward, we will remove list comprehensions from ES6 but persist them in "moz".)

src/jshint.js
lukeapage: Okay,  im doubting myself now,  but anyway the gyst is... 
Any es7 features should only be on in esversion = 7 and esnext should map to esversion 7. That way esversion is "correct" and it is not a breaking change..

lukeapage: Sorry for the confusion, generators did make es6, this thread only applies to arrray comprehensions.

src/jshint.js
nicolo-ribaudo: Done

src/jshint.js
nicolo-ribaudo: Done, but I've not yet documented `esversion: 7` becasue of https://github.com/jshint/jshint/issues/1939#issuecomment-62282120

nicolo-ribaudo:
@jugglinmike See https://github.com/jshint/jshint/issues/1148#issuecomment-19786395.

Since "moz" isn't standard ECMAScript, I think it shouldn't be included in `esversion`, but JSHint should complain when moz and esversion are used at the same time.

jugglinmike:
> Since "moz" isn't standard ECMAScript, I think it shouldn't be included in esversion, but JSHint should complain when moz and esversion are used at the same time.

I'm fine with this, but by the same token, `7` shouldn't be an accepted value of `esversion`.

> Another question: can esversion be changed after JSHint has parsed some code?

No. You can re-use [the logic I introduced for the `module` option](https://github.com/jugglinmike/jshint/commit/290280c29d47baad06d8958ef4a90807f4a67e40) to achieve this.

lukeapage:
<opinion>
I dont like the moz option.  We should remove let blocks since afaik they are a dead proposal,  replaced by let proper and not have moz.  Array comprehensions should go under 7 or esversion expeimental since afaik they are on the cards for 7/es2016. we could say it is not a breaking change for experimental to evolve.. 
If chrome implements comprehensions then we have a badly named option. 
Imo if something like comprehensions is still on the standards track there is no point removing it unless its hurting or causing issues or we are planning a big refactor
</opinion>

nicolo-ribaudo:
> I dont like the moz option. We should remove let blocks since afaik they are a dead proposal, replaced by let proper and not have moz.

Removing `moz` would be a very very breaking change, I don't think it would ever done

> we could say it is not a breaking change for experimental to evolve..

It would be a bug not evolving following proposals, but implementing an older version of them :wink:

---

> I'm fine with this, but by the same token, `7` shouldn't be an accepted value of `esversion`.

I think `esversion: 7` would be more intuitive than `esversion: moz`

jugglinmike:
> I dont like the moz option.

I'm not a fan of it either, but it's a commitment that was made for us, and we should honor it so long as it doesn't impede continued development. I have no idea what the active usage is like, but [a simple GitHub code search](https://github.com/search?l=json&q="moz"%3A+true&type=Code&utf8=✓) suggests that removing it would put a lot of people off.

> I think esversion: 7 would be more intuitive than esversion: moz

It's not about intuitiveness, it's about correctness. ES7 doesn't exist yet, and when it does, it will be a moving target. It should be considered experimental and exposed in a way that makes this clear. Definitely out-of-scope for this patch, though.

nicolo-ribaudo:
> It's not about intuitiveness, it's about correctness. ES7 doesn't exist yet,

I've removed `esversion: 7`, and `esversion: 6` still doesn't enable array comprehensions.

> Definitely out-of-scope for this patch, though.

I agree with you

nicolo-ribaudo:
Before last commit, linting this code:
```js
// jshint esversion: 3
// jshint moz: true
var a = {
  get x() {}
};
```
JSHint warned that getters are es5 features, even if moz is > than ES5

lukeapage:
> JSHint warned that getters are es5 features, even if moz is > than ES5

I remember seeing plenty of bugs about people not realising they had to have moz and es5/esnext on, but I think they were mostly in reference to a previous breaking change.

nicolo-ribaudo:
Should tests use `esversion` instead of `es3/5/next`?

lukeapage:
> Should tests use esversion instead of es3/5/next?

I think so.. it needs to be done at some point anyway. It seems like alot of hassle so you might want to wait for @jugglinmike's opinion.

jugglinmike:
I'd rather not since it adds a lot of noise to the non-trivial change happening here. I think that change would fit better in a single "Remove version-specific language options" commit, somewhere down the line

lukeapage:
@nicolo-ribaudo you might be interested to see Mike's blog post on jshint's way forward w.r.t. esversion - http://jshint.com/blog/new-lang-features/
It looks to me like this PR matches what was said.

nicolo-ribaudo:
I think I can revert 47587445016dca32ffa0b0e7a02a84a72e99494e

---

> @nicolo-ribaudo you might be interested to see Mike's blog post on jshint's way forward w.r.t. esversion - http://jshint.com/blog/new-lang-features/

Thanks for the link

nicolo-ribaudo:
Rebased on `master` to resolve merge conflicts.

jugglinmike:
@nicolo-ribaudo This patch no longer merges cleanly, but it doesn't seem fair to ask you to rebase again. I've squashed the commits together and added a single merge commit to resolve the conflicts; you can find that on the `esversion-option` branch of my fork: https://github.com/jugglinmike/jshint/tree/esversion-option

If that looks good to you, I can either squash those commits together or land them as-is. Or, if you prefer, you can of course resolve the conflicts yourself. Please let me know how you'd like to proceed.

I'm sorry I let this go for so long!

nicolo-ribaudo:
@jugglinmike I won't be able to rebase this until next weekend.
If this commit is needed sooner, feel free to merge your branch as-is.

nicolo-ribaudo:
@jugglinmike Rebased

jugglinmike:
> Rebased, to not introduce noise (the merge commit) in commits history.

A man after my own heart.

Squashed and merged to `master` at commit cf5a699b0f6fbe2be61f4972ea0a2bbef51a0bd2. Thanks, Nicolò!

PR #2521, opened 2015-07-04
[[FIX]] Param destructuring should not effect max params

### LINKED ISSUE #2183: Destructing operator hits parameter limit
With param limit set to 3:

let getResturants = ({page, pageSize, domain, metroId} = {}) => {
                              ^ This function has too many parameters. (4)

lukeapage:
Fixes destructuring so that multiple destructured params do not count
towards the maxparams limit. This is because destructuring is the solution
to having alot of numbered params, since you would replace a hard to
figure out list of params with an object, then use destructuring to pull
out from that object.
Fixes #2183

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2973,10 +2973,10 @@ var JSHINT = (function() {
 
     if (paramsInfo) {
       state.funct["(params)"] = paramsInfo.params;
-      state.funct["(metrics)"].verifyMaxParametersPerFunction(paramsInfo.arity);
-      state.funct["(arity)"] = paramsInfo.arity;
+      state.funct["(metrics)"].arity = paramsInfo.arity;
+      state.funct["(metrics)"].verifyMaxParametersPerFunction();
     } else {
-      state.funct["(arity)"] = 0;
+      state.funct["(metrics)"].arity = 0;
     }
 
     if (isArrow) {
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
I think you should also fix `JSHINT.data().functins[x].metrics.parameters `

lukeapage:
> I think you should also fix JSHINT.data().functins[x].metrics.parameters

done, good spot.

lukeapage:
rebased on master

jugglinmike:
The name "count" is a little generic and potentially misleading. Initially, I was wondering why you would track that separately than the `parameters` array itself because I was thinking it was redundant. In reality (and as I'm sure you know), `parameters` describes all bindings from the parameter list, so the number is definitely distinct. What do you think about re-naming `count` to `arity` to more clearly communicate the concept it describes?

lukeapage:
> What do you think about re-naming count to arity to more clearly communicate the concept it describes?

yes, arity is a more descriptive name, I'll change

lukeapage:
Fixed. Also added another commit - https://github.com/jshint/jshint/commit/f75b2d36f46fa9db913efcf736cbaec3df2b0708
When renaming it occurred to me this was a bit cleaner than polluting the function directly.

jugglinmike:
Excellent! Merged to `master` at commit 9d021eede3e5ebffe2c98fce12293703586b491b. Thanks, Luke!

Issue #1604, opened 2014-04-03
`undefined` incorrectly labeled as unused


kangax:
Here is a small repro:

``` js
(function() {
  var undefined;

  function foo() {
    return arguments[0] || undefined;
  }
  return foo;
}());
```

PR #2523, opened 2015-07-04
[[FIX]] Allow definition (but not assignment) of variables named `undefined`

nicolo-ribaudo:


lukeapage:
related - #1604

nicolo-ribaudo:
`scope-manager.js`, lines 133-156:
```js
function _checkParams() {
  var params = _current["(params)"];
  var param = params.pop();
  var unused_opt;

  while (param) {
    var label = _current["(labels)"][param];

    unused_opt = _getUnusedOption(state.funct["(unusedOption)"]);

    // 'undefined' is a special case for (function(window, undefined) { ... })();
    // patterns.
    if (param === "undefined")
      return;

    if (label["(unused)"]) { // <-- THIS LINE
      _warnUnused(param, label["(token)"], "param", state.funct["(unusedOption)"]);
    } else if (unused_opt === "last-param") {
      return;
    }

    param = params.pop();
  }
}
```

Why shouldn't JSHint warn that `undefined` is unused?

lukeapage:
> Why shouldn't JSHint warn if undefined is unused?

https://github.com/jshint/jshint/commit/e07b48a53f5eb157f18f02c1e39f3b9afa7ff0cc

 > There is also a special key: a variable named 'undefined'. JSHint
assumes that such variables are part of the popular defensive pattern
and doesn't consider such variables as unused. Otherwise I'd had to
treat 'undefined' as a variable everywhere and that's too much work.

But also, as its use in a defensive pattern, it kind-of makes sense to allow it to be unused but still have the same consistent pattern everywhere. It would be nice if there was better checking about that defensive pattern though.

tests/unit/core.js
```diff
@@ -183,24 +183,109 @@ exports.testNewNonNativeObject = function (test) {
...
-  var code1 = 'var undefined = 1;';
+  TestRun(test)
+    .test("var undefined;");
 
-  TestRun(test).test(code);
+  TestRun(test)
+    .test("let undefined;", { esnext: true });
+
+  TestRun(test)
+    .addError(1, "const 'undefined' is initialized to 'undefined'.")
```
lukeapage: I wonder if we should remove this error for the `undefined` label. it puts you in a spot, because if you were doing this in es6, you would really want to use `const` and not `let`.. but then with const you get an error assigning a value and an error not assigning a value.. ? WDYT?

tests/unit/core.js
```diff
@@ -183,24 +183,109 @@ exports.testNewNonNativeObject = function (test) {
...
+
+  TestRun(test)
+    .test("var { undefined } = {};", { esnext: true });
+
+  TestRun(test)
+    .test("var [ undefined ] = [];", { esnext: true });
+
+  var functions = [
+    "function a(undefined) {}",
+    "function b({ undefined }) {}",
```
lukeapage: I am not sure I like allowing undefined to be in destructuring expressions.
It *is* technically valid, you are right, but from a jshint perspective its confusing - we cannot guarentee that people will not pass `{ undefined: true }` and end up with what we were trying to protect against.
I think the patterns of having it as an argument and having it as a un-assigned variable is enough freedom ?

tests/unit/core.js
nicolo-ribaudo: I think JSHint should always warn *something* if you are naming a const `undefined`, as it isn't possible:

```js
(function() {"use strict";

const undefined; // Chrome throws "Unexpected token ;"

}());

(function() {"use strict";

const undefined = undefined; // Chrome throws 'undefined is not defined'

}());
```

lukeapage: I'm on chrome 43 and I get no error
```
(function() { "use strict"; const undefined = 1; return undefined; }())
```
Could be a chrome/V8 bug?

lukeapage: I cannot see it not disallowed in the grammar, so it might be a chrome bug?

nicolo-ribaudo: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-let-and-const-declarations-static-semantics-early-errors
```
It is a Syntax Error if Initializer is not present and IsConstantDeclaration of the LexicalDeclaration containing this production is true.
```

tests/unit/core.js
nicolo-ribaudo: Should JSHint warn `Expected an identifier and instead saw 'undefined' (a reserved word).` (as it does now)

tests/unit/core.js
lukeapage: yep you're right

tests/unit/core.js
lukeapage: what about the `"Assigning a value to 'undefined'` ?

nicolo-ribaudo:
>> There is also a special key: a variable named 'undefined'. JSHint
assumes that such variables are part of the popular defensive pattern
and doesn't consider such variables as unused.

Should JSHint warn that `undefined` is unused when it is declared using `let` or `var`?

lukeapage:
> Should JSHint warn that undefined is unused when it is declared using let or var?

I have no idea, I can see arguments either way.

Probably to be consistent with the argument pattern, it shouldn't ? unlike other variables a unused `var undefined;` is not a possible indicator of bugs, its just un-necessary code.

nicolo-ribaudo:
Rebased on master to resolve merge conflicts

src/jshint.js
```diff
@@ -2726,6 +2721,9 @@ var JSHINT = (function() {
       } else {
         if (checkPunctuators(state.tokens.next, ["..."])) pastRest = true;
         ident = identifier(true);
+        if (pastRest && state.tokens.curr.value === "undefined") {
```
lukeapage: could these use `checkUndefined` too ? or do you think that is over the top?
```
if (pastRest) {
    checkUndefined();
}
```

nicolo-ribaudo:
Rebased

nicolo-ribaudo:
Is this a breaking change?

This code passes in JSHint 2.8.0, but not in this branch
```js
// jshint -W024
var undefined = 2;
```

nicolo-ribaudo:
Rebased

shellscape:
I really hope there's movement on this!

PR #2527, opened 2015-07-05
[[FIX]] Allow destructuring in catch blocks parameter

### LINKED ISSUE #2526: Destructuring in catch's parameter
Code:
```js
// jshint esnext: true

try {} catch ({ message }) {}
```

Errors reported by JSHint:
```
$ bin/jshint --verbose test.js
test.js: line 3, col 15, Expected an identifier and instead saw '{'. (E030)
test.js: line 3, col 15, Expected ')' and instead saw '{'. (W116)
test.js: line 3, col 17, Expected '{' and instead saw 'message'. (E021)
test.js: line 3, col 17, Expected an assignment or function call and instead saw an expression. (W030)
test.js: line 3, col 25, Unrecoverable syntax error. (100% scanned). (E041)

5 errors
```

Expected:
no errors

Spec: ([13.15](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-try-statement))
> *Catch*<sub>[Yield, Return]</sub> :
> &nbsp; **catch (** *CatchParameter*<sub>[?Yield]</sub> **)** *Block*<sub>[?Yield, ?Return]</sub>
>
> *CatchParameter*<sub>[Yield]</sub> :
> &nbsp; *BindingIdentifier*<sub>[?Yield]</sub>
> &nbsp; *BindingPattern*<sub>[?Yield]</sub>

(*BindingPattern* is destructuring)

nicolo-ribaudo:
Example:
```js
try {
  throw new Error("foo");
} catch ({ message }) {
  console.log(message); // foo
}
```

Fixes #2526

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3761,13 +3761,11 @@ var JSHINT = (function() {
 
       if (checkPunctuators(state.tokens.next, ["[", "{"])) {
         var tokens = destructuringExpression();
-        var t;
-        for (t in tokens) {
-          t = tokens[t];
-          if (t.id) {
-            params.push({ id: t.id, token: t });
+        _.each(tokens, function(token) {
+          if (token.id) {
+            params.push({ id: token.id, token: token });
           }
-        }
+        });
       } else if (state.tokens.next.type !== "(identifier)") {
         warning("E030", state.tokens.next, state.tokens.next.value);
       } else {
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3754,18 +3754,25 @@ var JSHINT = (function() {
     var b;
 
     function doCatch() {
-      var e;
+      var params = [];
 
       advance("catch");
       advance("(");
 
-      e = state.tokens.next;
-      if (e.type !== "(identifier)") {
-        warning("E030", state.tokens.next, e.value);
-        e = null;
+      if (checkPunctuators(state.tokens.next, ["[", "{"])) {
+        var tokens = destructuringExpression();
+        var t;
+        for (t in tokens) {
+          t = tokens[t];
+          if (t.id) {
+            params.push({ id: t.id, token: t });
+          }
+        }
+      } else if (state.tokens.next.type !== "(identifier)") {
+        warning("E030", state.tokens.next, state.tokens.next.value);
       } else {
         // only advance if we have an identifier so we can continue parsing in the most common error - that no param is given.
-        advance();
+        params.push({ id: identifier(), token: state.tokens.curr, type: "exception" });
       }
 
       state.funct = functor("(catch)", state.tokens.next, {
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3754,18 +3754,25 @@ var JSHINT = (function() {
...
+          if (t.id) {
+            params.push({ id: t.id, token: t });
+          }
+        }
+      } else if (state.tokens.next.type !== "(identifier)") {
+        warning("E030", state.tokens.next, state.tokens.next.value);
       } else {
         // only advance if we have an identifier so we can continue parsing in the most common error - that no param is given.
-        advance();
+        params.push({ id: identifier(), token: state.tokens.curr, type: "exception" });
```
lukeapage: using identifier does allow the rest operator, which doesn't look like its allowed in exception catch.
This is already a bug elsewhere, but I am not sure at the moment whether identifier allowing spread is causing the allowance of spread everywhere or if its down to allowing "..." as an infix.
It might be nice to be able to pass something to identifier to tell it not to allow spread? not sure.

src/jshint.js
```diff
@@ -3754,18 +3754,25 @@ var JSHINT = (function() {
...
       advance("(");
 
-      e = state.tokens.next;
-      if (e.type !== "(identifier)") {
-        warning("E030", state.tokens.next, e.value);
-        e = null;
+      if (checkPunctuators(state.tokens.next, ["[", "{"])) {
+        var tokens = destructuringExpression();
+        var t;
+        for (t in tokens) {
```
lukeapage: I don't like the unguarded for-in, but that's the way its already done elsewhere. Maybe I'm just old fashioned. Who extends the array prototype now-a-days?

src/jshint.js
nicolo-ribaudo: > using identifier does allow the rest operator

Rest operator allowed because of lines 3771 and 3772.

---


```js
// jshint esnext: true
try {} catch (...error) {}
```

```
$ bin/jshint --verbose test.js
test.js: line 2, col 15, Expected an identifier and instead saw '...'. (E030)
test.js: line 2, col 15, Expected ')' and instead saw '...'. (W116)
test.js: line 2, col 18, Expected '{' and instead saw 'error'. (E021)
test.js: line 2, col 18, Expected an assignment or function call and instead saw an expression. (W030)
test.js: line 2, col 23, Missing semicolon. (W033)
test.js: line 2, col 23, Expected an identifier and instead saw ')'. (E030)
test.js: line 2, col 23, Expected an assignment or function call and instead saw an expression. (W030)
test.js: line 2, col 24, Missing semicolon. (W033)
test.js: line 2, col 18, 'error' is not defined. (W117)

9 errors
```

src/jshint.js
nicolo-ribaudo: I've copied these lines from lines 2718-2725. Should I also fix the loop there?

src/jshint.js
lukeapage: Sorry I missed you were specifically guarding against it before going into the function

lukeapage:
Squashed and committed, thanks - https://github.com/jshint/jshint/commit/759644c80f7d9d954dff9e5fe2f7a2d141238367

PR #2531, opened 2015-07-06
[[FIX]] Report an error when a necessary semicolon is missing

### LINKED ISSUE #1327: fails to detect invalid code - doing asi in a space, not a newline
If you have -W033 disabled (semicolon warning) the following lints just fine.
```js
var one, two, three
one = two two = three
```
When there's clearly a syntax error. And if you don't have the warning off, it still only produces a warning, not an error.

nicolo-ribaudo:
The `asi` option also disabled warning about missing necessary semicolons (e.g. semicolons between
two tokens on the same line). Now, they are reported as errors and can't be disabled.

Fixes #1327

### DIFF
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -635,7 +635,36 @@ exports.safeasi = function (test) {
     .addError(10, "Bad line breaking before '/'.")
     .addError(10, "Expected an identifier and instead saw '.'.")
     .addError(10, "Expected an assignment or function call and instead saw an expression.")
+    .addError(10, "Missing semicolon.")
     .test(src, { asi: true });
+
+  test.done();
+};
+
+exports["missing semicolons not influenced by asi"] = function (test) {
+  // These tests are taken from
+  // http://www.ecma-international.org/ecma-262/6.0/index.html#sec-11.9.2
+
+  var code = [
+    "void 0;", // Not JSON
+    "{ 1 2 } 3"
+  ];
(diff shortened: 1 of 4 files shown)


tests/unit/options.js
```diff
@@ -635,11 +635,40 @@ exports.safeasi = function (test) {
...
+  // http://www.ecma-international.org/ecma-262/6.0/index.html#sec-11.9.2
+
+  var code = [
+    "void 0;", // Not JSON
+    "{ 1 2 } 3"
+  ];
+
+  JSHINT(code, { expr: true, asi: true });
+
+  var error = JSHINT.errors[0]
```
lukeapage: did you do the test this way to specify the character ? if so, you can do
```
.addError(line, msg, { character: 4})
```

nicolo-ribaudo: Thank you, I didn't know that.

lukeapage:
This looks good to me, I've thought about it, played about etc. @jugglinmike do you want to take a look at this first (thought I should ask, since it adds an Error and isn't 100% straight forward) or is it good to land ?

jugglinmike:
Squashed, rebased (had to increment the error ID in light of 2fbf621d71225e5ce4381c65383225d60349a310), and merged to `master` at commit 45d8e3e45f064e07ea9edfbcbf6ec588c30554b2. Thanks!

PR #2537, opened 2015-07-08
[[FIX]] Support `new.target`

nicolo-ribaudo:
Spec: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-meta-properties

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1037,9 +1037,11 @@ var JSHINT = (function() {
           warning("W017", this);
         }
 
+        if (this.right && this.right.isMetaProperty) {
+          error("E031", this);
         // detect increment/decrement of a const
         // in the case of a.b, right will be the "." punctuator
-        if (this.right && this.right.identifier && !this.right.isMetaProperty) {
+        } else if (this.right && this.right.identifier) {
           state.funct["(scope)"].block.modify(this.right.value, this);
         }
       }
@@ -1298,6 +1300,10 @@ var JSHINT = (function() {
 
           that.right = expression(10);
           return that;
+        } else if (left.isMetaProperty) {
+          error("E031", that);
+          that.right = expression(10);
+          return that;
         } else if (left.identifier && !isReserved(left)) {
           if (state.funct["(scope)"].labeltype(left.value) === "exception") {
             warning("W022", left);
@@ -1343,6 +1349,11 @@ var JSHINT = (function() {
       }
 
       if (left) {
+        if (left.isMetaProperty) {
+          error("E031", that);
+          that.right = expression(10);
+          return that;
+        }
         if (left.id === "." || left.id === "[" ||
             (left.identifier && !isReserved(left))) {
           expression(10);
@@ -1369,9 +1380,11 @@ var JSHINT = (function() {
         warning("W017", this);
       }
 
+      if (left.isMetaProperty) {
+        error("E031", this);
       // detect increment/decrement of a const
       // in the case of a.b, left will be the "." punctuator
-      if (left && left.identifier && !left.isMetaProperty) {
+      } else if (left && left.identifier) {
         state.funct["(scope)"].block.modify(left.value, left);
       }
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2234,8 +2234,14 @@ var JSHINT = (function() {
       if (!state.inESNext(true)) {
         warning("W119", state.tokens.prev, "new.target");
       }
-      if (!state.inClassBody) {
-        warning("W136", state.tokens.prev, "new.target", "class body");
+      var inFunction, c = state.funct;
+      while (c) {
+        inFunction = !c["(global)"];
+        if (!c["(arrow)"]) { break; }
+        c = c["(context)"];
+      }
+      if (!inFunction) {
+        warning("W136", state.tokens.prev, "new.target");
       }
     });
     if (mp) { return mp; }
@@ -2801,6 +2807,7 @@ var JSHINT = (function() {
       "(scope)"     : null,
       "(comparray)" : null,
       "(generator)" : null,
+      "(arrow)"     : null,
       "(params)"    : null
     };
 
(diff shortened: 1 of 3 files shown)


src/jshint.js
```diff
@@ -2218,6 +2217,19 @@ var JSHINT = (function() {
     return this;
   }));
   prefix("new", function() {
+    var mp = metaProperty("target");
```
lukeapage: you already specify the name once, I wonder if it would be cleaner if metaProperty only returns if the value === "target" then you won't need to check it again below ?

src/jshint.js
```diff
@@ -2218,6 +2217,19 @@ var JSHINT = (function() {
...
+    if (mp) {
+      if (mp.value === "target") {
+        if (!state.inESNext(true)) {
+          warning("W119", state.tokens.prev, "new.target");
+        }
+        if (!state.inClassBody) {
```
lukeapage: reading up on this (thanks, I hadn't heard of this before!), I thought it was available in all functions, but not arrow functions ?

src/jshint.js
```diff
@@ -1265,7 +1264,7 @@ var JSHINT = (function() {
             warning("W121", left, nativeObject);
         }
 
-        if (left.identifier) {
+        if (left.identifier && !left.isMetaProperty) {
```
lukeapage: can we throw an error if it is assigned to?

lukeapage: I guess the same for any modification like ++, should be an error

nicolo-ribaudo: Are you sure it isn't possible to assign to new.target?

The spec:
```
AssignmentExpression[In, Yield] :
  ...
  LeftHandSideExpression[?Yield] = AssignmentExpression[?In, ?Yield]
  LeftHandSideExpression[?Yield] AssignmentOperator AssignmentExpression[?In, ?Yield]
```
```
LeftHandSideExpression[Yield] :
  NewExpression[?Yield]
  ...
```
```
NewExpression[Yield] :
  MemberExpression[?Yield]
  ...
```
```
MemberExpression[Yield] :
  ...
  MetaProperty
  ...
```

src/jshint.js
nicolo-ribaudo: if metaProperty only returns if value === "target", JSHint would throw a ton of errors when it encounters invalid meta properties because it would treat them as construction.

src/jshint.js
nicolo-ribaudo: If `metaProperty` only returns if value === "target", JSHint would throw a ton of errors when linting invalid meta properties because it would treat `new` as a construction.

nicolo-ribaudo: What do you think about using the `metaPrpoerty` function like this?
```js
var mp = metaProperty("target", function() {
  if (!state.inESNext(true)) {
    warning("W119", state.tokens.prev, "new.target");
  }
  if (!state.inClassBody) {
    warning("W136", state.tokens.prev, "new.target", "class body");
  }
});
if (mp) { return mp; }
```

lukeapage: Are there likely to be other meta properties ? I'm all for abstraction when it makes the code clearer or reduces duplication, but I wonder whether or not to just inline it? Because if other meta properties are added I can see having to pass an array to metaProperty (but we don't need to jump the gun on this)
as for the pattern above, I have no objections, though it isn't a pattern I've seen used before - I do prefer it to the current code.
I'll leave it to you - use the above or inline.

src/jshint.js
lukeapage: I saw your previous comment that you deleted about am I sure - the answer is no, I haven't found the exact place in the spec that says it can't, I got the information from kangax's compatibility table - https://kangax.github.io/compat-table/es6/#new.target_can%27t_be_assigned_to so if you are in doubt too, please check!

nicolo-ribaudo: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-static-semantics-static-semantics-isvalidsimpleassignmenttarget
```
Static Semantics: IsValidSimpleAssignmentTarget

...
NewTarget :
  new . target

1. Return false.
```

http://www.ecma-international.org/ecma-262/6.0/index.html#sec-assignment-operators-static-semantics-early-errors
```
AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression

* It is an early Reference Error if IsValidSimpleAssignmentTarget of LeftHandSideExpression is false.
```

src/jshint.js
nicolo-ribaudo: > Are there likely to be other meta properties?

There is a stage 2 proposal - [`function.sent`](https://github.com/allenwb/ESideas/blob/master/Generator%20metaproperty.md)

src/jshint.js
lukeapage: Good to know!

src/jshint.js
nicolo-ribaudo: How can I check if it is in an arrow function?

lukeapage: the arrow function is simply ignored. So, you need to know, am I in a non-arrow function (somewhere on the stack). I guess you might need a new state property? but also this is exactly the same behaviour as `arguments` - can't remember how that is done though, it might be done by adding a variable to the scope, which is clearly not an option.

jugglinmike:
Re: use within arrow functions--there's no explicit restriction there. It's just that [there needs to be a `this` binding](http://www.ecma-international.org/ecma-262/6.0/#sec-getthisenvironment). That means this is valid:

```js
(function() {
  () => { new.target; };
});
```

...but this is not:

```js
(() => { new.target; });
```

lukeapage:
> there needs to be a this binding

a much more succinct description than my clarification

> in a non-arrow function (somewhere on the stack)

lukeapage:
Great work! LGTM.

lukeapage:
committee - https://github.com/jshint/jshint/commit/2fbf621d71225e5ce4381c65383225d60349a310

PR #2538, opened 2015-07-08
[[Fix]] Speed issues with linting very large files

### LINKED ISSUE #2532: Processing of large scripts takes very long (starting with release 2.6.1)
I am working with Javascripts for single-page apps which have concatenated sizes of 400kb-800kb.

Using releases until 2.5.11, processing of these scripts takes approx. 2 seconds on an average PC.

In later releases, jshint needs up approx. 30 seconds to process the same scripts, which makes it unusable in frequently executed development builds.

lukeapage:
Fixes #2532 

linting the example file goes from 60 seconds to 2 seconds with these changes!

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -684,7 +684,11 @@ var JSHINT = (function() {
   //     for ( var i = ...
 
   function peek(p) {
-    var i = p || 0, j = 0, t;
+    var i = p || 0, j = lookahead.length, t;
+
+    if (i < j) {
+      return lookahead[i];
+    }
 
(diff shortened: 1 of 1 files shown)


jugglinmike:
Merged to `master` at commit b33e2d7a9a457733fc76c7af30da3dd8f1aa79a9

Sorry for the delay here. I'm interested in a more formal method for enforcing expectations on performance, but that work is taking longer than expected. I don't want to hold this up any longer, especially because the changeset is so minimal.

Thanks!

PR #2539, opened 2015-07-10
Revert "[[FIX]] Allow `__proto__` when using ES6"

jugglinmike:
This reverts commit 06b57649414e44a9df6a46907f3ac93513e44221. That
changeset changed which warnings JSHint produces for existing code,
making it backwards-incompatable with the last release.

PR #2540, opened 2015-07-11
[[FIX]] Allow `__proto__` when using ES6

lukeapage:
Implement the breaking change of a modified warning for proto.
This reverts commit 2c0dc5e7bdb865ca31b903f6eb487bbb45f49e55.

originally implemented by @nicolo-ribaudo in https://github.com/jshint/jshint/pull/2377
reverted by @jugglinmike in https://github.com/jshint/jshint/pull/2539 due to breaking change


lukeapage:
Discussion point: whether this should be a breaking change and in the v3 release or the fix should be changed to not be breaking and put in a v2 release.

PR #2542, opened 2015-07-12
[[FIX]] Do not allow spread operators where not allowed

### LINKED ISSUE #2525: We allow spread operators anywhere
E.g.
```
var a = (...(a) => ());
var b = ...a;
[{a:...a}] = [];
```
I believe they are only valid
a) in array destructuring
b) in the parameter list
c) in arrays

I need to investigate the spec to be sure

lukeapage:
Fixes #2525 

After some consideration I mark the token as to whether it can be a spread operator. It seems the closest to what is done elsewhere, the other viable alternative would be making `prefix("..."` into `spreadExpression` and calling that in the 2 places, which optionally allows a spread before the expression (this is close to what identifier does, but in the issue @jugglinmike wasn't so keen).

I looked at using an expression classifier, but I think that approach involves a substantial change to the way jshint works - as I understand it, all expressions would track their type as they went along and then at the end of the expression the caller would work out whether that type was valid. This is the same approach discussed as an option for #2511 - at the point of the assignop, work out if its a destructure, in all other places if a destructure occurs, after parsing that bit, then show an error.. I think thats a large change because jshint a) it starts taking as down a road that makes the parser *harder* to understand (when the approach is used by v8 for performance reasons) b) jshint tracks state and does asserts as it goes along, it doesn't parse and then afterwards traverse and error - all errors and state changes would have to be put into the expression classifier as it went along.


lukeapage:
rebased on master

src/jshint.js
```diff
@@ -2201,7 +2202,10 @@ var JSHINT = (function() {
     // }
     // // false
     //
-    if (!state.tokens.next.identifier &&
+    if (!state.tokens.curr.allowSpread) {
+      error("E059", state.tokens.curr);
+    }
+    else if (!state.tokens.next.identifier &&
```
nicolo-ribaudo: Just for style consistency, `else` should be on the same line of `}`

lukeapage: Fixed, thanks

jugglinmike:
One thing that I did not consider about maintaining this as a "prefix" operator is the need for a dedicated warning message. If you took "option A" as proposed in gh-2525, then the parser would automatically generate errors whenever it encountered `...` in an invalid location.

In some ways an explicit error is nicer because it is more likely to make sense to the end user. In other ways, it reflects a potentially untenable attitude towards parsing; we can't hope to interpret the intention behind all syntax errors, and just trying to do so would quickly result in a huge list of rarely-emitted ultra-specific error strings.

@lukeapage I know this is a bit inconsistent of me, but I wanted to share this thought because I'm honestly on the fence about this. I'd be happy to land it either way; I just thought I should give you an opportunity to refactor in case you shared any of these concerns. Let me know!

lukeapage:
> I just thought I should give you an opportunity to refactor in case you shared any of these concerns. Let me know!

I was on the fence like you or I would have come back and suggested I now had a preference. it is hard to predict what will be the best future approach..

jugglinmike:
Well, an error is an error. If we ever come down off this fence, we can change the implementation without worrying about breakage. I'll merge this as-is

jugglinmike:
Actually, it looks like some changes in `master` are incompatible with this patch (I'm guessing ae4896670a46247ef59df9d3ea9cf3814c864559). Rebasing takes place without conflict, but the resulting patch no longer passes its own tests.

lukeapage:
ok,  I will take a look.

jugglinmike:
Trying to be more constructive, I did some research on this. Here's a patch that resolves the new issue; it also includes a minor refactoring in how BindingElements are parsed in order to limit duplicating the warning:

https://github.com/jugglinmike/jshint/commit/d74303a9ae3464bddfa5c9708bc8e877938c74eb

lukeapage:
@jugglinmike your patch looks good

lukeapage:
Added your commit @jugglinmike and rebased

PR #2543, opened 2015-07-12
[[FIX]] Regular args can come after args with default

### LINKED ISSUE #1779: Rest parameters and default parameters
JSHint reports `Regular parameters cannot come after default parameters` if you use both default parameter and rest parameter, but I think this is a valid use case:
```js
function f (x = 0, ...y) {}
```

lukeapage:
Fixes #1779 

Not a breaking change since it used to be an error, which could not be overridden, this makes it a warning which can be overridden.

Other option is to remove the warning completely.

### DIFF
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -1438,7 +1438,7 @@ exports.testDefaultArguments = function (test) {
     .addError(14, "'bar' is not defined.")
     .addError(14, "'num3' was used before it was declared, which is illegal for 'param' variables.")
     .addError(15, "'num4' was used before it was declared, which is illegal for 'param' variables.")
-    .addError(11, "Regular parameters should not come after default parameters.")
+    .addError(18, "Regular parameters should not come after default parameters.")
     .addError(27, "'c' is not defined.")
     .addError(33, "'d' was used before it was defined.")
     .addError(36, "'e' was used before it was declared, which is illegal for 'param' variables.")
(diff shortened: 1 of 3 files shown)


lukeapage:
rebased on master

tests/unit/core.js
```diff
@@ -1438,7 +1438,7 @@ exports.testDefaultArguments = function (test) {
     .addError(14, "'bar' is not defined.")
     .addError(14, "'num3' was used before it was declared, which is illegal for 'param' variables.")
     .addError(15, "'num4' was used before it was declared, which is illegal for 'param' variables.")
-    .addError(18, "Regular parameters cannot come after default parameters.")
+    .addError(11, "Regular parameters should not come after default parameters.")
```
nicolo-ribaudo: Why did you change the line number?
```js
/* Line 11 */ testUseOfVariables: function(num = foo, // Ok, defined above
                             num1 = num, // Ok, previous argument
                             num2 = num3, // warn, next argument
                             num3 = bar, // warn, not defined
                             num4 = num4) { // warn, self reference
},
/* Line 18 */ testBadDefault: function(num, num2=1, num3) {
  return num === num2;
},
```

lukeapage: I have no idea how that happened - the tests fail as well because of this. I've pushed a commit to fix it. Thanks.

jugglinmike:
Squashed and merged to `master` at commit f2a59f1ea8b31491f9dcc2c5ded32d8ca0024d74. Thanks, Luke!

PR #2544, opened 2015-07-12
[[FIX]] Immediately-invoked arrow functions' parameter doesn't need parentheses.

### LINKED ISSUE #2351: [es6] "Bad invocation" when an immediately-invoked arrow function hasn't parenthesis around its argument
This works:
```javascript
/*jshint esnext: true */

((b) => {})(0);
```

This doesn't work: (But is the same of the first example)
```javascript
/*jshint esnext: true */

(b => {})(0);
```

With the second I get this warning: `Bad invocation`

nicolo-ribaudo:
Fixes #2351

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2548,7 +2548,7 @@ var JSHINT = (function() {
           addInternalSrc(left, p[0].value);
         }
       }
-      if (!left.identifier && left.id !== "." && left.id !== "[" && left.value !== "=>" &&
+      if (!left.identifier && left.id !== "." && left.id !== "[" && left.id !== "=>" &&
           left.id !== "(" && left.id !== "&&" && left.id !== "||" && left.id !== "?" &&
           !(state.option.esnext && left["(name)"])) {
         warning("W067", that);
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -2548,9 +2548,9 @@ var JSHINT = (function() {
           addInternalSrc(left, p[0].value);
         }
       }
-      if (!left.identifier && left.id !== "." && left.id !== "[" &&
-          left.id !== "(" && left.id !== "&&" && left.id !== "||" &&
-          left.id !== "?" && !(state.option.esnext && left["(name)"])) {
+      if (!left.identifier && left.id !== "." && left.id !== "[" && left.value !== "=>" &&
```
lukeapage: is it safer to use `left.id` instead of `left.value` ? my only concern.

nicolo-ribaudo: You are right: using `.value`, this code passes the test:
```js
("=>")();
```

lukeapage:
Thanks!

lukeapage:
Committed - https://github.com/jshint/jshint/commit/d261071e7b523549499077bb13e1affb7e25158e

PR #2546, opened 2015-07-12
Loopfunc fixes

### LINKED ISSUE #2153: loopfunc not enforced when using arrow functions
```js
var trouble = [];
for(var i=0; i < 10; i++) {
    trouble.push(() => i);
}
```

It should warn as if it was `trouble.push(function() {return i;});`

lukeapage:
[[FIX]] use of params is not capturing loopfunc
Previously the scope manager only considered the function body for usages
that went outside scope, indicating "capturing" that should trigger
loopfunc. This corrects that.

Also affected are function names, but in the case of declarations there is
a different warning and in the case of expressions the name available to
the function will be correct even within a loop.

Note - see this code
```
var a = function b(arg) { if (!arg) { b(true); } else console.log("first"); }, c = function b(arg) { if (!arg) { b(true); } else console.log("second"); }; a(); c();
```

[[FIX]] Report loopfunc for all function types
Previously only function expressions caused loopfunc, meaning arrow
functions that captured variables were not warned.
Fixes #2153

Note: if you consider this breaking code because it will start warning about loopfunc's (I think it is enough within the spirit of the option and obscure enough that its okay in a minor release) then I'll split into 2 PR's as the first commit should really go in asap.


tests/unit/fixtures/loopfunc.js
```diff
@@ -29,3 +29,15 @@ function boo(p) {
 while (true) {
   var d = function z() { return z(); };
 }
+
+for(i = 0; i < 5; i++) {
```
mc10: Minor nit: You should add a space after `for`.

lukeapage:
> Minor nit: You should add a space after for.

It was in the test fixture file so it mattered very little, but I fixed it for consistency.

lukeapage:
@jugglinmike the first commit here fixes a regression and should go in before any release.

lukeapage:
rebased on master and added a commit (the fixup! one)

src/scope-manager.js
```diff
@@ -13,21 +13,26 @@ var marker = {};
...
+  var _scopeStack = [];
+
+  function _newScope(type) {
+    _current = {
       "(labels)": Object.create(null),
       "(usages)": Object.create(null),
       "(breakLabels)": Object.create(null),
-      "(parent)": _current
+      "(parent)": _current,
+      "(type)": type,
```
jugglinmike: I like the new `"(type)"` attribute!

src/scope-manager.js
```diff
@@ -483,19 +504,6 @@ var scopeManager = function(state, predefined, exported, declared) {
...
     },
 
-    /**
-     * Tell the manager we are entering a new scope with parameters
-     * Call stack after all parameters have been added
-     * @param isFunction Whether the scope is a function or not
-     */
-    stackParams: function(isFunction) {
-      _current = _newScope();
-      _current["(isParams)"] = isFunction ? "function" : true;
```
jugglinmike: Also nice to see `"(isParams)"` going away :)

src/scope-manager.js
```diff
@@ -446,6 +459,10 @@ var scopeManager = function(state, predefined, exported, declared) {
           "(type)" : type,
           "(token)": token,
           "(unused)": true };
+
+        if (!_current["(params)"]) {
+          _current["(params)"] = [];
```
jugglinmike: This seems like initialization logic. Does it really need to be dynamic? Since
  the `_newScope` function now receives the block type, is there any chance we
  can extend the logic there to create the array where necessary? My goal here
  is to make that function more descriptive of the "shape" of the scope object.
  As it stands, it's a little cagey about the `"(params)"` attribute (both its
  type and when it may be used). What do you think of:

```diff
diff --git a/src/jshint.js b/src/jshint.js
index 08dc777..187fa33 100644
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3872,7 +3872,7 @@ var JSHINT = (function() {
       advance("catch");
       advance("(");
 
-      state.funct["(scope)"].stack();
+      state.funct["(scope)"].stack("catchparams");
 
       if (checkPunctuators(state.tokens.next, ["[", "{"])) {
         var tokens = destructuringPattern();
diff --git a/src/scope-manager.js b/src/scope-manager.js
index e97a7e8..0494b84 100644
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -24,7 +24,7 @@ var scopeManager = function(state, predefined, exported, declared) {
       "(breakLabels)": Object.create(null),
       "(parent)": _current,
       "(type)": type,
-      "(params)": null
+      "(params)": (type === "functionparams" || type === "catchparams") ? [] : null
     };
     _scopeStack.push(_current);
   }
@@ -254,7 +254,9 @@ var scopeManager = function(state, predefined, exported, declared) {
 
     /**
      * Tell the manager we are entering a new block of code
-     * @param {string} [type] - the type of the block. Valid values are "functionparams", "functionouter"
+     * @param {string} [type] - the type of the block. Valid values are
+     *                          "functionparams", "catchparams", and
+     *                          "functionouter"
      */
     stack: function(type) {
       var previousScope = _current;
@@ -465,9 +467,6 @@ var scopeManager = function(state, predefined, exported, declared) {
           "(token)": token,
           "(unused)": true };
 
-        if (!_current["(params)"]) {
-          _current["(params)"] = [];
-        }
         _current["(params)"].push(labelName);
       }
```

jugglinmike:
This patch no longer applies cleanly, but the resolution is trivial. I *think* the "proper" way to handle this mid-review is to use a merge commit that can be squashed prior to landing. I've never done that before, though, so if it's wrong or too much of a hassle, feel free to simply rebase (but please keep non-conflict-related changes in separate commits).

lukeapage:
> I think the "proper" way to handle this mid-review is to use a merge commit that can be squashed prior to landing. I've never done that before, though, so if it's wrong or too much of a hassle, feel free to simply rebase (but please keep non-conflict-related changes in separate commits).

I think that would work as long as you squash before rebasing. In the past I've had a branch with merges and wanted to keep the commits and rebased, but then you find you have to re-do the conflicts with each commit.

src/scope-manager.js
lukeapage: > This seems like initialization logic. Does it really need to be dynamic? Since
> the _newScope function now receives the block type, is there any chance we
> can extend the logic there to create the array where necessary? My goal here
> is to make that function more descriptive of the "shape" of the scope object.
> As it stands, it's a little cagey about the "(params)" attribute (both its
> type and when it may be used). What do you think of:

Yes, thats better. I'm sure I had it like this and changed it, I think I was trying to minimise the number of magic values to stack and make it more flexible - but yes wanting to know the shape of the object was why I put (params) in set to null, this is fine.

lukeapage:
> I think that would work as long as you squash before rebasing. In the past I've had a branch with merges and wanted to keep the commits and rebased, but then you find you have to re-do the conflicts with each commit.

I just noticed it is two Fix commits for seperate issues - I think you'll have trouble rebasing them now without squashing them together. The merge conflict is quite trivial - the incoming conflict is just a rename of _currentFunct to _currentFunctBody

jugglinmike:
Squashed, rebased, and merged to `master` at the following commits:

- 827e335b29f402ccc8dc2260a82d5df4c071329b
- 4d4cfcdfbd97adbe93e044e67bfa3bc81bd128af

For posterity, the rebased commits produce an equivalent trees:

    $ git diff 4d4cfc 91356b9
    $

Thanks, Luke!

PR #2547, opened 2015-07-12
[[FIX]] Allow line breaking after yield if `asi: true`

### LINKED ISSUE #2530: Unexpected 'Bad line breaking' warning
```js
// jshint asi: true, esnext: true, undef: false
function* gen() {
  yield
  fn()
}
```

JSHint warns `Bad line breaking before 'fn'.`, so it is suggesting me to rewrite that code as this:
```js
// jshint asi: true, esnext: true, undef: false
function* gen() {
  yield fn()
}
```

But these codes are different: the first one is
```js
function* gen() {
  yield;
  fn();
}
```
While the second is
```js
function* gen() {
  yield fn();
}
```

Spec: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-rules-of-automatic-semicolon-insertion

nicolo-ribaudo:
Fixes #2530

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -6006,13 +6006,18 @@ exports["test for line breaks with 'yield'"] = function (test) {
     "function* gen() {",
     "  yield",
     "  fn();",
+    "  yield*",
+    "  fn();",
     "}"
   ];
 
-  TestRun(test, "gh-2530").test(code2, { esnext: true, undef: false, asi: true });
+  TestRun(test, "gh-2530")
+    .addError(5, "Bad line breaking before 'fn'.")
+    .test(code2, { esnext: true, undef: false, asi: true });
 
   TestRun(test, "gh-2530")
     .addError(3, "Bad line breaking before 'fn'.")
+    .addError(5, "Bad line breaking before 'fn'.")
     .test(code2, { esnext: true, undef: false });
 
   test.done();
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -6001,6 +6001,20 @@ exports["test for line breaks with 'yield'"] = function (test) {
     .addError(13, "Missing semicolon.");
 
   run.test(code, {moz: true});
+
+  var code2 = [
+    "function* gen() {",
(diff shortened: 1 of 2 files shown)


tests/unit/parser.js
```diff
@@ -6001,6 +6001,20 @@ exports["test for line breaks with 'yield'"] = function (test) {
     .addError(13, "Missing semicolon.");
 
   run.test(code, {moz: true});
+
+  var code2 = [
+    "function* gen() {",
+    "  yield",
```
lukeapage: I think its a good example of why you need to know what your doing if you don't add semi-colons.

tests/unit/parser.js
```diff
@@ -6001,6 +6001,20 @@ exports["test for line breaks with 'yield'"] = function (test) {
...
+    "  fn();",
```
jugglinmike: Could you add another statement for `yield*`? Something like

```diff
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -6006,13 +6006,18 @@ exports["test for line breaks with 'yield'"] = function (test) {
     "function* gen() {",
     "  yield",
     "  fn();",
+    "  yield*",
+    "  fn();",
     "}"
   ];
 
-  TestRun(test, "gh-2530").test(code2, { esnext: true, undef: false, asi: true });
+  TestRun(test, "gh-2530")
+    .addError(5, "Bad line breaking before 'fn'.")
+    .test(code2, { esnext: true, undef: false, asi: true });
 
   TestRun(test, "gh-2530")
     .addError(3, "Bad line breaking before 'fn'.")
+    .addError(5, "Bad line breaking before 'fn'.")
     .test(code2, { esnext: true, undef: false });
 ```

It's not a problem with the patch, but it's a likely source of regression, so I'd like an explicit test for it.

nicolo-ribaudo:
Done

jugglinmike:
Squashed and merged to `master` at commit 728c84bc4230c9b7f12a79de7845af9f5fbc5464 . Thanks, Nicolò!

PR #2556, opened 2015-07-16
[[FIX]] Labels shadowing within a function is a syntax error

### LINKED ISSUE #2419: label shadowing is a syntax error but is only warned under shadow inner
from discussion here: https://github.com/jshint/jshint/issues/2394#issuecomment-106620733

The following code:
```
/* jshint shadow:true */
a: while (true) {
  a: while (true) {}
}

function b() {
  c: while (true) {
  	c: while (true) {}
	}
}
```

does not error, even though it is a syntax error. With `shadow: inner` the 2nd case will warn.

lukeapage:
Fixes #2419

### DIFF
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -694,7 +694,10 @@ var scopeManager = function(state, predefined, exported, declared) {
 
       addBreakLabel: function(labelName, opts) {
         var token = opts.token;
-        if (state.option.shadow === "outer") {
+        if (scopeManagerInst.funct.hasBreakLabel(labelName)) {
+          warning("E011", token, labelName);
+        }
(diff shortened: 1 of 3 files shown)


jugglinmike:
Thanks for the quick turnaround, Luke! This "upgrades" the warning generated on `master` (W004) to an error (E011), so users will no longer be able to ignore it. I think this is acceptable for a minor release because the condition describes syntactically invalid JavaScript. Would you agree?

lukeapage:
Yes  that is exactly why I did it...

jugglinmike:
Just making sure we're on the same page.

Anyway, I've merged this to `master` at 124e00fecd7df0c16554dcc6b1263151cd0ae8c8. Thanks again

PR #2559, opened 2015-07-18
[[FIX]] Improve support for `__proto__` identifier

jugglinmike:
Information about identifiers used are stored internally as properties
of objects whose key values are the identifier value. Because of
non-standard implementations of the `__proto__` property, input source
code that uses it as an identifier can have inconsistent side effects
across environments.

When using an object as a lookup table, the built-in
`Object.hasOwnProperty` is generally preferable for detecting property
membership. Unfortunately, due to the non-standard implementations
described above, this check has environment-specific results.

An alternate approach to membership detection is coercing direct
property references to a boolean value. This is not appropriate in cases
where the entries may themselves be "falsey" values, but in the
identifier lookup tables, entries are consistently object literals.

This second approach yields the same results across all legacy
environments and is also compatable with the standardized specification
for the `__proto__` attribute (provided the table is initialized to have
no prototype).

`var o = {};`

                                           | Compliant | Node.js v0.10.40 | IE9   | PhantomJS v1.9
------------------------------------------ | --------- | ---------------- | ----- | --------------
**Before `__proto__` assignment**          |           |                  |       |
!!o.__proto__                              | true      | true             | false | true
'__proto__' in o                           | true      | true             | false | true
Object.hasOwnProperty.call(o, '__proto__') | false     | false            | false | true
Object.keys(o).length                      | 0         | 0                | 0     | 0
**After `__proto__` assignment**           |           |                  |       |
!!o.__proto__                              | true      | true             | true  | true
'__proto__' in o                           | true      | true             | true  | true
Object.hasOwnProperty.call(o, '__proto__') | false     | false            | true  | true
Object.keys(o).length                      | 0         | 0                | 1     | 0

`var o = Object.create(null);`

                                           | Compliant | Node.js v0.10.40 | IE9   | PhantomJS v1.9
------------------------------------------ | --------- | ---------------- | ----- | --------------
**Before `__proto__` assignment**          |           |                  |       |
!!o.__proto__                              | false     | false            | false | false
'__proto__' in o                           | false     | true             | false | true
Object.hasOwnProperty.call(o, '__proto__') | false     | false            | false | true
Object.keys(o).length                      | 0         | 0                | 0     | 0
**After `__proto__` assignment**           |           |                  |       |
!!o.__proto__                              | true      | true             | true  | true
'__proto__' in o                           | true      | true             | true  | true
Object.hasOwnProperty.call(o, '__proto__') | true      | false            | true  | true
Object.keys(o).length                      | 1         | 0                | 1     | 0

Compliant enviroments: latest SpiderMonkey (and Firefox), latest V8 (and
Chrome), Node.js v0.12.7, Iojs v.2.3.4

Update all lookup tables to be created without a prototype, and update
all membership tests to simply coerce direct attribute references.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4616,7 +4616,7 @@ var JSHINT = (function() {
       name = tkn.value;
     }
 
-    if (props[name]) {
+    if (props[name] && name !== "__proto__") {
       warning("W075", state.tokens.next, msg, name);
     } else {
       props[name] = Object.create(null);
@@ -4652,7 +4652,7 @@ var JSHINT = (function() {
     state.nameStack.set(tkn);
 
     if (props[name]) {
-      if (props[name].basic || props[name][flagName]) {
+      if ((props[name].basic || props[name][flagName]) && name !== "__proto__") {
         warning("W075", state.tokens.next, msg, name);
       }
     } else {
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3061,7 +3061,7 @@ var JSHINT = (function() {
     // Check for lonely setters if in the ES5 mode.
     if (state.inES5()) {
       for (var name in props) {
-        if (_.has(props, name) && props[name].setterToken && !props[name].getterToken) {
+        if (props[name] && props[name].setterToken && !props[name].getterToken) {
           warning("W078", props[name].setterToken);
         }
       }
@@ -3086,7 +3086,7 @@ var JSHINT = (function() {
   (function(x) {
     x.nud = function() {
       var b, f, i, p, t, isGeneratorMethod = false, nextVal;
-      var props = {}; // All properties, including accessors
+      var props = Object.create(null); // All properties, including accessors
 
       b = state.tokens.curr.line !== startLine(state.tokens.next);
       if (b) {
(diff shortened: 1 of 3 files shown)


jugglinmike:
Once this lands, we will be able merge gh-2128 (it's currently failing for [this test](https://github.com/jshint/jshint/blob/124e00fecd7df0c16554dcc6b1263151cd0ae8c8/tests/unit/core.js#L1398-L1401)). Running automated tests in this environment will allow us to catch regressions in this fix. In the mean time, I can try to author failing test cases for the changes you've requested by locally applying that patch.

jugglinmike:
@lukeapage I've pushed up two `fixup!` commits, one for each of the changes you requested. As usual, I'm happy to squash whenever you think it's ready.

lukeapage:
I'll look tomorrow.

jugglinmike:
I'd be the last one to rush you :P

lukeapage:
Nice! Perfection! LGTM.

lukeapage:
Reordered commits and fixedup, committed to master
https://github.com/jshint/jshint/commit/925a9837867ff27882a758f444174bf44420e216
and https://github.com/jshint/jshint/commit/477d3add0a1ec973f3d1d2d427f1cea183153123

lukeapage:
See #2566 for PR to revert commit as a reminder it should be done for v3

Issue #2560, opened 2015-07-18
Warning about the 'module' option when setting function-level flags


gustavderdrache:
This input:
```javascript
/* jshint esnext: true, module:true */

function bar() {
  /* jshint validthis:true */
  return this;
}
```
warns "The 'module' option cannot be set after any executable code." on line 4 (`validthis: true`).

jugglinmike:
Confirmed. Thanks for the report!

nicolo-ribaudo:
@jugglinmike I didn't see you self-assigned this and I fixed the issue... can I open a pr or you have already started resolving this issue?

jugglinmike:
I have already started

nicolo-ribaudo:
Ok

lukeapage:
This has been merged and will be fixed in the next version of jshint.

PR #2561, opened 2015-07-18
[[FIX]] Relax restriction on `module` option

jugglinmike:
While the `module` option is intended to be set before executable code,
this restriction should only be enforced when it is first enabled (and
not with the application of every in-line directive).

Resolves gh-2560

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -199,13 +199,6 @@ var JSHINT = (function() {
 
     if (state.option.module) {
       /**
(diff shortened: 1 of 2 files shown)


lukeapage:
This looks very straight forward to me. Looks good.

lukeapage:
Committed - https://github.com/jshint/jshint/commit/16f5779c04204f5127a3f1f85d6031f01dcabc00

Issue #2471, opened 2015-06-18
Validthis breaks when adding function definition?


erwinw:
The following code gives no jshint warnings:

	'use strict';
	
	/*jshint validthis:true*/
	function a(){
		this.foo = 'bar';
	}
	/*jshint validthis:false*/

	a();

The following code does give a jshint warning:
	
	'use strict';
	
	function foo(){}

	/*jshint validthis:true*/
	function a(){
		this.foo = 'bar';
	}
	/*jshint validthis:false*/

	a();
	foo();

In other words, adding that (completely unrelated) function definition somehow makes jshint ignore the validthis flag.

nicolo-ribaudo:
Duplicate - #2436

Issue #1666, opened 2014-05-09
-W041 not in effect when a function uses `!= null` above the hint


SheetJSDev:
```
var x = 3, y;
function foo(f,g,h) { return h != null; } 
/*jshint -W041 */
console.log(x != null);
```

Expected behavior: only one warning for `h != null`

Actual behavior: two warnings

SheetJSDev:
Probably should clarify: putting the jshint comment above the first instance does the right thing:

```
var x = 3, y;
/*jshint -W041 */
function foo(f,g,h) { return h != null; } 
console.log(x != null);
```

Issue #1383, opened 2013-11-22
Could not disable warnings about assignment usage where JSHint expected conditional expression (W084)


narqo:
JSHint throws warning "Expected a conditional expression and instead saw an assignment (W084)" for code below:

``` javascript
function my() {
  var p, i, mparams;

  function wat() {}

  /* jshint -W084 */
  while (p = mparams[i++]) {
    b();
  }
  /* jshint +W084 */
}
```

The same time, if one comments `wat` function, everything looks fine for JSHint.

bucaran:
What is the `.jshintrc` that I should add to avoid adding that comment.

nicolo-ribaudo:
This is fixed by #2563

Issue #2436, opened 2015-06-02
maxparams cannot be overridden in some cases when dealing with prototypes


hristozov:
Consider the following JSHint configuration in `.jshintrc`:

    {
	"maxparams": 4
    }

and the following in `fail.js`:

    function Foo () {}

    // jshint maxparams: 9
    Foo.prototype.bar = function (a, b, c, d, e) {
      return a + b + c + d + e;
    };

Running the latest JSHint with the configuration and the source file produces a warning, although the comment explicitly overrides `maxparams`:

    $ jshint -c .jshintrc fail.js
    fail.js: line 4, col 30, This function has too many parameters. (5)

    1 error
    $ jshint -v
    jshint v2.8.0

However, moving the line to the top of the file does not produce any warnings. This file is OK:

    // jshint maxparams: 9
    function Foo () {}

    Foo.prototype.bar = function (a, b, c, d, e) {
      return a + b + c + d + e;
    };

Furthermore, replacing the prototype modification with a plain object assignment does not produce any warnings as well:

    var Foo = {};

    // jshint maxparams: 9
    Foo.bar = function (a, b, c, d, e) {
      return a + b + c + d + e;
    };

nicolo-ribaudo:
It looks like you can't overwrite the `maxparams` option after you have defined a function:

```js
// jshint maxparams: 1

function noop() {}

// jshint maxparams: 3

function sum(x, y) { // warning
  return x + y;
}
```

```js
// jshint maxparams: 1

var notAFunction;

// jshint maxparams: 3

function sum(x, y) { // ok
  return x + y;
}
```

nicolo-ribaudo:
I've tested a little.

JSHint doesn't pick up options which are defined immediately after a function:

```js
function x() {
  eval();
}

/*jshint evil: true */ // <-- Ignored
function y() {
  eval(); // Warning
}
```
```js
function x() {
  eval();
}

var something;

/*jshint evil: true */
function y() {
  eval(); // ok
}
```

nicolo-ribaudo:
I think fixing this issue would also fix this:

```js
// This doesn't warn, while it should because evil:true is set after eval()
eval()/*jshint evil: true */;
```

```js
eval();/*jshint evil: true */ // Warns, ok
```

PR #2563, opened 2015-07-18
support jshint options in comments after function declaration.

lukeapage:
Comments welcome, but currently a WIP.

Looks like 2nd approach might be doable, needs some tweaking as tests don't work at the moment.


lukeapage:
@nicolo-ribaudo I see why you described the problem but didn't attempt a fix! any ideas you had would be welcome.

src/jshint.js
```diff
@@ -698,9 +697,21 @@ var JSHINT = (function() {
   }
 
   // Produce the next token. It looks for programming errors.
+  var nextCommentOptions = [];
```
jugglinmike: Careful--the preceding comment is meant to describe the `advance` function.

jugglinmike: I know this is still early, but this value should probably be defined on the `state` object.

tests/unit/options.js
```diff
@@ -2932,3 +2932,27 @@ exports.module.declarationRestrictions = function( test ) {
...
+      "function Foo () {}",
+      "// jshint maxparams: 9", // <-- previously ignored
+      "Foo.prototype.bar = function (a, b, c, d, e) {",
+      "  return a + b + c + d + e;",
+      "};",
+      "var z = () => foo;",
+      "// jshint maxparams: 1",
+      "function y(a, b) {}",
+      "// jshint evil: false",
+      "eval()/*jshint evil: true */;"
```
nicolo-ribaudo:
```
✖ gh-2436
Assertion Message: 
  TestRun: [bold]{undefined}
  [yellow]{Errors thrown by JSHINT, but not defined in test run}
  [bold]{Line 14, Char 1} eval can be harmful.
```

I think here the warning is correct (`evil: true` is set after `eval()`), but this would be a breaking change :disappointed:

lukeapage: yes it is. It is kind of a breaking change, so it might need a major release, upto @jugglinmike. Its hard to fix the bug properly and not introduce a breaking change. Its another one of those where the existing behaviour is probably not expected so we have to guess whether this will likely effect many people.
There is another reason why it might be worth waiting to major, which I'll comment in a sec.

nicolo-ribaudo:
Passing `nt` into `assume` (and using it instead of `state.tokens.next` at lines 205 and 212) would fix these errors:
```
✖ module - behavior
Assertion Message: 
  TestRun: [bold]{undefined}
  [yellow]{Errors with wrong line number}
  [bold]{Line 2} The 'module' option is only available when linting ECMAScript 6 code. [red]{not in line(s)} [bold]{1}

✖ module - declarationRestrictions
Assertion Message: 
  TestRun: [bold]{undefined}
  [yellow]{Errors with wrong line number}
  [bold]{Line 3} The 'module' option cannot be set after any executable code. [red]{not in line(s)} [bold]{2}
```

src/jshint.js
```diff
@@ -401,8 +401,35 @@ var JSHINT = (function() {
     return i;
   }
 
-  function doOption() {
-    var nt = state.tokens.next;
+  function doLexOption(nt) {
```
lukeapage: I don't like having to add a doLexOption but I don't see a choice to make `maxlen` backwards compatible because lexer options are applied a token before. There are possibly ways to do this with a bigger refactor, but since maxlen will be removed in the next major, my plan was do this for now and remove this function along with maxlen.

lukeapage:
I've updated the PR, its ready for review

lukeapage:
> Passing nt into assume (and using it instead of state.tokens.next at lines 205 and 212) would fix these errors:

Thanks, that's what I ended up doing - I just gave up last night and wasn't sure I'd have time today, but I did have...

lukeapage:
Fixes #2471, #2436, #2407, #1852

nicolo-ribaudo:
Fixes #1666

nicolo-ribaudo:
Fixes #1383

lukeapage:
Thanks for tracking down those extra ones @nicolo-ribaudo

PR #2565, opened 2015-07-19
De-duplicate config finding logic

steelbrain:
Just like the diff says

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -623,7 +623,7 @@ var exports = {
         }
 
         if (filename && !config) {
-          config = exports.getConfig(filename)
+          config = exports.getConfig(filename);
         }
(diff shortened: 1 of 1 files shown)


lukeapage:
This seems okay - but first can we get the build and tests passing?

They are failing at the moment with this error...
```
src/cli.js: line 626, col 47, Missing semicolon.
```

steelbrain:
I don't add semi-colons, but it's a personal taste, I understand that you want to maintain your codebase's codestyle, so I've added it now.

lukeapage:
committed as https://github.com/jshint/jshint/commit/ba1caea4e704bbe0174e6bf776cb73ade9322272, thanks

PR #2566, opened 2015-07-19
[[FIX]] Revert regressions commit for __proto__

lukeapage:
So that jshint now warns on __proto__ as other variables.

So its not forgotten. See #2559


nicolo-ribaudo:
Does this replace #2379?

jugglinmike:
Haha you're very quick! It does; I'll close that now

jugglinmike:
This will resolve gh-2372

PR #2569, opened 2015-07-19
[[FIX]] Allow empty export and trailing comma

### LINKED ISSUE #2567: Parse errors with ES6 export {} declaration
Both of these lines cause an error about an expected identifier, but [the grammar](http://www.ecma-international.org/ecma-262/6.0/#sec-exports) indicates they should be allowed:

```
/* jshint esnext:true */

export {};
export {x,};
```

nicolo-ribaudo:
Fixes #2567

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4520,7 +4520,7 @@ var JSHINT = (function() {
       // ImportClause :: NamedImports
       advance("{");
       // Breaking change
-      // if (checkPunctuators(state.tokens.next, ["}"])) {
+      // if (checkPunctuator(state.tokens.next, "}")) {
       //   warning("W137", state.tokens.next, "import");
       // }
       for (;;) {
@@ -4616,11 +4616,11 @@ var JSHINT = (function() {
     if (state.tokens.next.value === "{") {
       // ExportDeclaration :: export ExportClause
       advance("{");
-      if (checkPunctuators(state.tokens.next, ["}"])) {
+      if (checkPunctuator(state.tokens.next, "}")) {
         warning("W137", state.tokens.next, "export");
       }
       var exportedTokens = [];
-      while (!checkPunctuators(state.tokens.next, ["}"])) {
+      while (!checkPunctuator(state.tokens.next, "}")) {
         if (!state.tokens.next.identifier) {
           error("E030", state.tokens.next, state.tokens.next.value);
         }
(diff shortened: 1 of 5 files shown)


src/jshint.js
```diff
@@ -4378,6 +4378,10 @@ var JSHINT = (function() {
     } else {
       // ImportClause :: NamedImports
       advance("{");
+      // Breaking change
```
lukeapage: So this is the same for imports, but currently it is allowed so unlike exports its a breaking change?

It would be good to have this in and then a undoing commit - we can then revert the undoing for v3

@jugglinmike do you agree empty imports and exports should be warned?

src/messages.js
```diff
@@ -216,7 +216,7 @@ var warnings = {
   W134: "The '{a}' option is only available when linting ECMAScript {b} code.",
   W135: "{a} may not be supported by non-browser environments.",
   W136: "'{a}' must be in function scope.",
-  W137: "Empty destructuring."
+  W137: "Empty {a}."
```
lukeapage: On another PR it was pointed out one of my messages was too simplistic and we could be more helpful. I am wondering if thats the case here, but I can't think of a more useful message - it seems pretty straightforward?

Empty destructuring, this is un-necessary and can be removed? I'm not sure, so only make that change if you agree.

src/jshint.js
jugglinmike: Yes, a warning sounds like a good idea to me

nicolo-ribaudo: > It would be good to have this in and then a undoing commit - we can then revert the undoing for v3

Why just don't open a new pr that will be merged for v3? So in the changelog there won't be both `Warn when an import is empty` and `Don't warn when an import is empty`, which is confusing.

lukeapage: Okay. I started doubting myself after I said it anyway.

nicolo-ribaudo:
@lukeapage Still `Needs work`?

nicolo-ribaudo:
Rebased

jugglinmike:
Thanks for this, @nicolo-ribaudo! I've rebased your work to account for improvements we recently made to the test harness and to update the (relatively) new whitelist for Test262 (this patch fixed six failures in that suite). I've merged the result to `master` at the following commits:

- aed66d7 [[DOCS]] Make `Empty {a}.` message more useful
- 631327e [[FIX]] Allow empty export and trailing comma

I've also submitted a follow-up to further improve the warning message, since `import {} from 'module.js';` is not completely superfluous. See gh-3185.

Issue #2570, opened 2015-07-19
`singleGroups` forbids grouping unary operations


jugglinmike:
The binding power of unary operations is not considered when `singleGroups` is enabled. For example, the grouping operator in the following expression has a substantive effect on the expression, but JSHint considers is unnecessary:

```js
(-3).toString();
```

jugglinmike:
Patch incoming!

lukeapage:
A fix for this has been merged and will be in the next release after the current 2.8.0

PR #2562, opened 2015-07-18
[[CHORE]] Remove commented out code from scope manager

lukeapage:
spotted this reviewing your PR @jugglinmike

PR #2571, opened 2015-07-19
Single groups fixes 2

jugglinmike:
Resolves gh-2570 and gh-2548

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2527,7 +2527,7 @@ var JSHINT = (function() {
           // Used to delineate an integer number literal from a dereferencing
           // punctuator (otherwise interpreted as a decimal point)
           (ret.type === "(number)" &&
-            checkPunctuators(pn, ".") && /^\d+$/.test(ret.value));
+            checkPunctuator(pn, ".") && /^\d+$/.test(ret.value));
       }
     }
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2508,7 +2508,11 @@ var JSHINT = (function() {
           // operator.
           (isFunctor(ret) && !isEndOfExpr()) ||
           // Used as the return value of a single-statement arrow function
-          (ret.id === "{" && preceeding.id === "=>");
+          (ret.id === "{" && preceeding.id === "=>") ||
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -2508,7 +2508,11 @@ var JSHINT = (function() {
           // operator.
           (isFunctor(ret) && !isEndOfExpr()) ||
           // Used as the return value of a single-statement arrow function
-          (ret.id === "{" && preceeding.id === "=>");
+          (ret.id === "{" && preceeding.id === "=>") ||
+          // Used to delineate an integer number literal from a dereferencing
+          // punctuator (otherwise interpreted as a decimal point)
+          (ret.type === "(number)" &&
+            checkPunctuators(pn, ".") && /^\d+$/.test(ret.value));
```
lukeapage: it should be checkPunctuators(pn,  ["."]) for consistency shouldn't it? at least till my pr is merged

tests/unit/options.js
```diff
@@ -2596,6 +2596,71 @@ singleGroups.lineNumber = function (test) {
   test.done();
 };
 
+singleGroups.unary = function (test) {
+  var code = [
+    "(-3).toString();",
```
lukeapage: I tried to think of missing tests and although not directly related to the bugs, I can't find a test for `(a++).toString()` - which requires a bracket?
Otherwise I can't think of anything else sensible to test.

jugglinmike:
@lukeapage Now that gh-2562 has landed, I've incorporated your feedback. I wanted to avoid rebasing to make review easier for you, so there's a merge commit in there. It should all be taken care of with the following command:

    $ git rebase -i --autosquash master

lukeapage:
ok will look later.

lukeapage:
looks good.
committed to master
6d6bd8f4feacebc3e7a97c2d6ce4873ffc2c88e0
8c265ca47342dd5f7dcf1f2a934cd7130c2a9c51
1bbcc64aa6963a2dae172a06458506ff0e720252

jugglinmike:
Thanks!

PR #2572, opened 2015-07-20
[[FIX]] Restrictions about `super()` and `this`'s position.

nicolo-ribaudo:
>- `super` can only be called in the constructor of classes which extend
  another class. It can't be used in other methods, in functions (except in
  arrow functions contained in `constructor()`), or in global scope.
- `super` must be called in the constructor of subclasses.
- `this` can only be used after `super()` in the constructor of subclasses.

I think this isn't a breaking change, as the new errors are reported for invalid code.


src/jshint.js
```diff
@@ -3592,6 +3604,8 @@ var JSHINT = (function() {
     if (state.tokens.next.value === "extends") {
       advance("extends");
       c.heritage = expression(10);
+    } else {
+      c.heritage = null;
```
lukeapage: this is just so its null rather than undefined?

src/jshint.js
```diff
@@ -3668,6 +3682,9 @@ var JSHINT = (function() {
         continue;
       }
 
+      var superCall = c.heritage && c.heritage.type !== "null" &&
```
lukeapage: nice catch on the extending null

src/jshint.js
lukeapage: it doesn't catch this highly contrived example
```
// jshint esnext: true
class a extends function(){ return null;}() {
  constructor() {
    super();
  }
}
```
but I think that is fine. If we take that into account we can't really do any static analysis of whether super is required/allowed.

src/jshint.js
lukeapage: it doesn't catch this highly contrived example
```
// jshint esnext: true
class a extends function(){ return null;}() {
  constructor() {
    super();
  }
}
```
but I think that is fine. If we take that into account we can't really do any static analysis of whether super is required/allowed.

src/jshint.js
```diff
@@ -4584,7 +4605,25 @@ var JSHINT = (function() {
   FutureReservedWord("public", { es5: true, strictOnly: true });
   FutureReservedWord("short");
   FutureReservedWord("static", { es5: true, strictOnly: true });
-  FutureReservedWord("super", { es5: true });
+  FutureReservedWord("super", {
+    es5: true,
+    nud: function() {
+      if (state.inESNext() && checkPunctuators(state.tokens.next, "(")) {
```
lukeapage: will this warn on this
```
// jshint esnext: false
function a() {
  super();
}
```
I noticed I don't get any warnings at all in 2.8.0 (e.g. super is reserved, super is not defined, super requires es6)

This would not be a regression so we can handle this seperately if you want? (unless I'm wrong of course)

src/jshint.js
```diff
@@ -4584,7 +4605,25 @@ var JSHINT = (function() {
...
+        var f = state.funct;
+        while (f) {
+          // If `super` is called in an arrow function, check the outer scope
+          if (!f["(arrow)"]) { break; }
+          f = f["(context)"];
+        }
+        if (!f["(super)"]) {
+          error("E061", state.tokens.curr);
+        } else {
+          f["(super)"] = "called";
```
lukeapage: is it okay to call super twice? again not a regression so we could handle in a different pr

lukeapage: if it is okay, it should perhaps be a warning.

tests/unit/core.js
```diff
@@ -1952,3 +1952,139 @@ exports.duplicateProto = function (test) {
...
+    "}"
+  ];
+
+  TestRun(test).test(code, { esnext: true });
+
+  var code2 = [
+    "class A {}",
+    "class B extends A {",
+    "  constructor() {",
+    "    this.x = 1;",
```
lukeapage: How about a test case for
```
constructor() {
   (()=>this.x = 1)();
   super();
}
```

lukeapage:
I think its a really nice thing for jshint to check, nice work.

src/jshint.js
nicolo-ribaudo: I don’t know why, but without this some tests (for example, the last new.target test) fail. However, by copying them in a file, JSHint throws the right errors.

lukeapage: thats strange, I guess it must be to do with re-using the jshint object

src/jshint.js
nicolo-ribaudo: Looking in the spec it seems it is allowed (and [babel doesn't warn](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=true&spec=false&playground=false&code=class%20A%20extends%20Array%20%7B%0A%20%20constructor()%20%7B%0A%20%20%20%20super()%3B%0A%20%20%20%20super()%3B%0A%20%20%7D%0A%7D))

lukeapage: ok, IMO it should warn. Maybe in a seperate pr for v3, seems to me like it would most commonly be a mistake.

nicolo-ribaudo: I will open a pr as soon as this gets merged.

tests/unit/core.js
nicolo-ribaudo: Done

lukeapage: Sorry, I didn't mean to make it warn.. I'm not sure whether it should or shouldn't..

Presumably in this case
```
constructor() {
   var a = ()=>this.x = 1;
   super();
   a();
}
```

it works?

in this case..
```
(()=>a)();
let a;
```

we don't error about use in the TDZ because we don't know that the function is immediately invoked or not - we do however give a latdef warning, if that option is on. There is no such latdef for super type option (it seems to much of a corner case to create one), so we have to decide.

@jugglinmike do you think we should warn in this case or not?

lukeapage:
For https://github.com/nicolo-ribaudo/jshint/commit/4fdd18b8d6e7c0461a83fdca83ff4ca3247cbe23

* You've used W119 - elsewhere W024 if used for future identifiers and W119 is used for use of future syntax. Whether this is syntax or use as an identifier.. I'm not sure. I wonder if W024 would be more backwards compatible.. 
* I don't think this is a *bad* breaking change because defining super warns W024
E.g. the following works at the moment..
```
// jshint es3: true, -W024

function super() {
}

var super = function(){};

super();
```
and as long as it still works after your change, I wonder if thats ok?

@jugglinmike would appreciate your opinion. It might be best marking this breaking change for inclusion in v3 since it shouldn't be too far away?

tests/unit/core.js
nicolo-ribaudo: Ops, you are right, it shouldn't warn.

nicolo-ribaudo:
Rebased

nicolo-ribaudo:
@lukeapage Still `Needs work`?

nicolo-ribaudo:
Is `super()` required if the class extends `null`? Currently this patch doesn't warn in this case, but...

[14.5.14 - Runtime Semantics: ClassDefinitionEvaluation](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-runtime-semantics-classdefinitionevaluation)
```
...
6. ...
   e. If superclass is null, then
      i. Let protoParent be null.
     ii. Let constructorParent be the intrinsic object %FunctionPrototype%.
   ...
```

... therefore the function has a parent constructor to call, even if it extends null.

nicolo-ribaudo:
Rebased.
I added another commit to warn when `super.property` is used before `super()`

nicolo-ribaudo:
Rebased. I added another commit (371f54a) which removes the check for `extends null`, because the specification says that if the class extends null, the class type is set to `"derived"` (so `this` is not initialized by default) and the `super` constructor is set to `Function.prototype` (an empty function), which must be called.

PR #2576, opened 2015-07-21
[[FIX]] Make `strict: func` have precedence over env options.

nicolo-ribaudo:
>Now it is possible to ovverride node, phantom, module and browserify's `strict: global` using
`strict: func`, so that JSHint warns if `"use strict";` isn't in a function.
Example:
```js
/* jshint node: true, strict: func */
"use strict"; // Warning: Use the function form of "use strict".
```

>Env options still have precedence over `strict: true`.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -198,6 +198,9 @@ var JSHINT = (function() {
     }
 
     if (state.option.module) {
+      if (state.option.strict === true) {
+        state.option.strict = "global";
+      }
       /**
        * TODO: Extend this restriction to *all* ES6-specific options.
        */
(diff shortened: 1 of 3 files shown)


lukeapage:
I like this and I think its kind-of just a bugfix to https://github.com/jshint/jshint/commit/8de82472dbdca9d7f55e0b99a119e9aae7aafba5

It needs to go in before the next release otherwise it will be a breaking change.

lukeapage:
committed - https://github.com/jshint/jshint/commit/d138db87e8fbe8d7022c8d5972e581a15b952851

PR #2588, opened 2015-07-25
[[FIX]] Dont crash when testing x == keyword if eqnull is on

### LINKED ISSUE #2587: You can make jshint crash with `true == if`
```
// jshint eqnull: true
true == if
```

I have a fix

lukeapage:
If eqnull was on jshint previously crashed if the right hand side of an
assignment was not an expression.
Fixes #2587

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -926,36 +926,79 @@ exports.comma = function (test) {
 
 exports["gh-2587"] = function (test) {
 
-  var ops = ["==", "!="];
-  var options = [{}, {eqeqeq: true}, {eqnull: true}, {eqeqeq: true, eqnull: true}];
+  TestRun(test)
+    .addError(1, "Expected an identifier and instead saw 'if'.")
+    .addError(1, "Unrecoverable syntax error. (100% scanned).")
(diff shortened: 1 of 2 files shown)


tests/unit/parser.js
```diff
@@ -924,6 +924,42 @@ exports.comma = function (test) {
   test.done();
 };
 
+exports["gh-2587"] = function (test) {
+
+  var ops = ["==", "!="];
+  var options = [{}, {eqeqeq: true}, {eqnull: true}, {eqeqeq: true, eqnull: true}];
+
+  ops.forEach(function(op) {
+    options.forEach(function(option) {
```
jugglinmike: People will be reading this test far more often than they will be trying to modify it. I'd much prefer a simpler style that declare expected input and output without abstraction, even if it means less-than-exhaustive coverage.

lukeapage: done.

jugglinmike:
Squashed and merged to `master` at commit 6afd373e9b3f9e22076122bc7d7e7a16f7b368dd. Thank you, Luke!

PR #2590, opened 2015-07-26
[[FIX]] Unify assign operation checks.

### LINKED ISSUE #2589: A number of warnings aren't reported for bitwise assignment
1.
```
try{} catch(e) {
   e |= 1;
}
```
no "Expected an identifier in an assignment and instead saw a function invocation."

2.
```
function a() {}
a |= 1;
```
no "Reassignment of 'a', which is is a function. Use 'var' or 'let' to declare bindings that may change."

3.
```
function a() {
  arguments.length |= 1;
}
```

no "bad assignment"

lukeapage:
The bitwise assignment checks for left hand side do not have the same
rigour as normal assignment so this unifies the two.
Fixes #2589

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1269,69 +1269,81 @@ var JSHINT = (function() {
     if (prototype) return walkNative(prototype);
   }
 
+  /**
+   * Checks the left hand side of an assignment for issues, returns if ok
+   * @param {token} left - the left hand side of the assignment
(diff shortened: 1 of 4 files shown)


jugglinmike:
Much cleaner! Merged to `master` at commit 06eb1d2c30288734a7c9a80c9cbb8c4ec60d35cf

PR #2593, opened 2015-07-26
[[FIX]] Report `++ +` as confusing just like `+ ++`

### LINKED ISSUE #2592: `+ ++` reports confusing plusses, but `++ +` is fine
from our own tests
```
b = a + ++b;
b = a++ + a;
```

only one warns.

lukeapage:
Fixes #2592


tests/unit/parser.js
```diff
@@ -54,15 +54,17 @@ exports.confusingOps = function (test) {
...
   ];
 
   var run = TestRun(test)
     .addError(1, "Confusing minuses.")
     .addError(2, "Confusing plusses.")
     .addError(3, "Confusing minuses.")
-    .addError(4, "Confusing plusses.");
+    .addError(4, "Confusing plusses.")
+    .addError(5, "Confusing minuses.")
+    .addError(6, "Confusing plusses.")
```
nicolo-ribaudo: Here there should be a semicolon

PR #2597, opened 2015-07-29
[[FIX]] maxcomplexity doesn't take into account `&&`

### LINKED ISSUE #840: High cyclomatic complexity on switch statements
Currently, every case in a switch block increments the cyclomatic complexity regardless of whether or not it falls through.

This has a complexity of 4:

```
function(someVal) {
    switch (someVal) {
        case 1:
        case 2:
        case 3:
            doSomething();
            break;
        default:
            doSomethingElse();
            break;
    }
}
```

This has a complexity of 2 while being logically equivalent:

```
function(someVal) {
    if (someVal === 1 || someVal === 2 || someVal === 3) {
        doSomething();
    } else {
        doSomethingElse();
    }
}
```

I'm not an expert in cyclomatic complexity by any means, so I'm honestly not sure if this is correct or not. Even if this is the expected behavior, it seems like it would be nice to at least have a flag which could make these two function calculate the same cyclomatic complexity.

lukeapage:
Fixes #840

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1105,6 +1105,12 @@ var JSHINT = (function() {
     });
   }
 
+  function infixDefaultLed(left, token, precedence) {
+    token.left = left;
+    token.right = expression(precedence);
+    return token;
+  }
+
   function infix(s, f, p, w) {
     var x = symbol(s, p);
     reserveName(x);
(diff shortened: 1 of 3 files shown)


lukeapage:
interestingly neither does http://jscomplexity.org/

I don't understand why not..

jugglinmike:
Thanks, Luke! And sorry for the egregious delay.

PR #2611, opened 2015-08-08
empty case statements should not increase complexity #840

paul-wade:
case only increases complexity if the next statement is not immediately another case.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4002,9 +4002,11 @@ var JSHINT = (function() {
 
         advance("case");
         this.cases.push(expression(0));
-        increaseComplexityCount();
         g = true;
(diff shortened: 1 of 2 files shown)


lukeapage:
see https://github.com/jshint/jshint/issues/840#issuecomment-128913071
i don't think this is wanted. 
please feel free to disagree

paul-wade:
I'm not sure I agree. In visual studio and intellij in C# and java respectively, complexity is NOT increased by an empty case statement. I also wonder outside of consistency with other operators within jshint how increasing complexity of an empty case makes sense?

jugglinmike:
Superseded by gh-2597

Issue #2581, opened 2015-07-24
Consider using conventional-changelog to generate changelog


stevemao:
we have [conventional-changelog](https://github.com/ajoslin/conventional-changelog) or [conventional-github-releaser](https://github.com/stevemao/conventional-github-releaser) (there are also grunt/gulp tasks) to automate the process of writing the changelog. As you are using formatted commit messages so the logs can be generated well according to the parts of the message. We use handlebars for templating the logs. It can also ignore reverted commits. There are many other features. We have a jshint preset for people using jshint commit format.
The modules are actively maintained.

stevemao:
A sample generated based on your commits

<a name="2.8.0"></a>
# 2.8.0 (2015-07-24)


### Bug Fixes

* Add `HTMLCollection` to browser environment. ([e92d375](https://github.com/jshint/jshint/commit/e92d375)), closes [#2443](https://github.com/jshint/jshint/issues/2443)
* Add `window.performance` to browser vars ([3ff1b05](https://github.com/jshint/jshint/commit/3ff1b05)), closes [#2461](https://github.com/jshint/jshint/issues/2461)
* Allow `__proto__` when using ES6 ([06b5764](https://github.com/jshint/jshint/commit/06b5764)), closes [#2371](https://github.com/jshint/jshint/issues/2371)
* Allow default values in destructuring. ([04ace9a](https://github.com/jshint/jshint/commit/04ace9a)), closes [#1941](https://github.com/jshint/jshint/issues/1941)
* Allow non-identifier PropertyNames in object destructuring. ([aa8a023](https://github.com/jshint/jshint/commit/aa8a023)), closes [#2467](https://github.com/jshint/jshint/issues/2467)
* Declare `func` as a property of `state` ([3be8d36](https://github.com/jshint/jshint/commit/3be8d36))
* Do not accept empty values for directives ([a5bfefb](https://github.com/jshint/jshint/commit/a5bfefb))
* Don't warn when Array() is used without 'new'. ([5f88aa7](https://github.com/jshint/jshint/commit/5f88aa7)), closes [#1987](https://github.com/jshint/jshint/issues/1987)
* handle no 'home' environment variables ([946af3e](https://github.com/jshint/jshint/commit/946af3e))
* Honor `ignore` directive more consistently ([0971608](https://github.com/jshint/jshint/commit/0971608))
* labeled break and continue semantics ([da66f70](https://github.com/jshint/jshint/commit/da66f70))
[...]

Issue #2331, opened 2015-04-22
How do you keep in sync with new options of jshint?


jakub-g:
Hello,
thanks for everyone for contributing to jshint, it's an invaluable tool.

The biggest pain point I've noticed is that I fail to keep up with new features :)
Every now and then new flags land in jshint but I did not find any simple way to find them out - other than iterating over http://jshint.com/docs/options/ and comparing the options to the ones in my `.jshintrc`.

The Github release notes are also not that helpful since they usually list all the commits, unfiltered.

Is there some changelog hidden  somewhere that documents which new options are added in which jshint version?

jugglinmike:
A more user-friendly change log would definitely be a good thing. Related: gh-1864

jugglinmike:
We're now using conventional-changelog to generate a formal CHANGELOG.md file, which is consumed by the jshint.com build process to generate blog posts. Hopefully these new resources will be useful when updating!

PR #2612, opened 2015-08-09
[[FEAT]] Use conventional-github-releaser

stevemao:


jugglinmike:
Thanks for the patch! I think I prefer the host-agnostic "conventional-changelog" module you originally proposed in gh-2581. I'd like to check the changelog in to the project for discoverability. We could then parse that file to generate blog posts on jshint.com for each release. What do you think?

jugglinmike:
(Doing that would address an open issue against this project: gh-2331.)

stevemao:
Yeah, of course! I didn't know you guys already have issues with changelog but I think it can address those :)

stevemao:
Done. A quick link to the generated changelog
https://github.com/stevemao/jshint/blob/changelog/CHANGELOG.md

jugglinmike:
Thanks Luke. Care to comment?

lukeapage:
> Thanks Luke. Care to comment?

No, I don't really have any comments ?

jugglinmike:
@stevemao I've been playing with this locally, and there's just one last thing I'd like to confirm. It seems as though, configured in this way, the tool will insert new entries into the changelog only (in other words, it will not re-generate the existing log). I think this is ideal, because it would allow for manual entry of commentary that is not defined in the project history. (As stated above, I want to re-use the generated file to create pages on http://jshint.com/blog/ , and I'll be starting by manually inserting the content of existing blog posts into the `CHANGELOG.md` file.)

Is this correct? Do you know if anyone is currently using the tool in combination with manual modification of the file?

stevemao:
I didn't manually generate the logs
`conventional-changelog -p jshint -i CHANGELOG.md -w -r 0`
and
`conventional-github-releaser -p jshint -r 0`

stevemao:
This is why I regenerated the logs for you for the first time. Of course you could do that by yourself :)

jugglinmike:
Sorry, I didn't explain myself properly. Let me put it another way:

If we accept this change, then I will obviously want to use the tool to update the `CONTRIBUTING.md` file when it comes time to release JSHint 2.9.0. After I do that, is it "safe" for me to manually edit that file, adding some additional notes for the release that were not automatically inserted?

I ask because the tool could have been built in such a way that the entire changelog file is re-created every time it is run. If that were the case, I would lose all the manual editions when I released the *next* version of JSHint (2.9.1, for example).

Testing locally, it looks like this is not how the tool works. In other words, it should be "safe" to insert manual additions because the tool only inserts new content. I just wanted to verify this with you because this will be an important part of how I would like to maintain the changelog.

stevemao:
This is correct :) 

Longer answer: We don't want user to lose their previous manually edited logs too unless you do `conventional-changelog -p jshint -i CHANGELOG.md -w -r 0`
For `conventional-github-releaser` it will never overwrite any release. You have to delete first then make a new one.

stevemao:
BTW, `conventional-changelog` and `grunt-conventional-changelog` projects themselves are doing this way. We always leave previous generated logs. We sometimes add some entries in them too.

jugglinmike:
Perfect! I've landed this with some modifications to the commit subject lines (using "DOCS" and "CHORE" as described by the project's contribution guidelines). I followed up by [removing the old changelog script](https://github.com/jshint/jshint/commit/c59ce02f4ec4c77fcc499350faa070a1cb43936d) and [importing additional changelog content from jshint.com](https://github.com/jshint/jshint/commit/d4db77129e4f6daf74a944ab3b83af05c2c891a5). Thanks a bunch, Steve--this is a much-needed improvement to our process and documentation!

stevemao:
Thanks @jugglinmike

Issue #2048, opened 2014-12-19
plusplus options should be in "Enforcing options" section


clark800:
At http://jshint.com/docs/options/

Also "strict" option.

Issue #2045, opened 2014-12-19
Remove `enforceall` option


jugglinmike:
The `enforceall` option is a shorthand for enabling all available "enforcing" options. A feature that allows consumers to automatically opt-in to all features makes it technically impossible to publish a [SemVer 2.0](http://semver.org/)-compliant minor release. As long as JSHint supports the `enforceall` option, _any_ new "enforcing" option has to be considered a breaking change and therefore requires a new major version.

This option should be removed in JSHint 3.0 so that new features can safely be added between major releases.

jugglinmike:
I appreciate your perspective, @Krinkle, but I'm not quite ready to give up on semver. I have been thinking about this, though, and I think there's a way to satisfy the "concise configuration" use case in a semver-compliant way. We remove `enforceall` and implement a new option called `featureset`. This option takes a major/minor version pair (e.g. `featureset: "2.3"` and enables the features available in JSHint at that release. This would ensure consistent behavior across minor version releases in spite of new feature additions.

Because we may remove options altogether between major versions, I expect we would also remove support for values in the previous release (i.e. setting `featureset: "2.6"` would trigger an error JSHint 3.0). This would allow us to avoid technical debt of maintaining options we decided to remove. Removing support these values in this way would itself be a breaking change, but that's what major versions are for.

It would *also* force a best practice on JSHint, namely: explicitly encoding the version that each option was introduced. This information could be displayed in the JSHint website's documentation (which are now procedurally generated).

paul-wade:
To me a breaking change is something like removing an option causing existing configs to be invalid or changing the CLI so IDE plugins need to be updated. I don't see how semver plays any role in WHAT enforceall turns on.

The fact that new options are included in enforceall is what people wanted. "we want everything on, we always want the most strict rules, we don't want to maintain a giant config with everything on" If you add a new feature people using enforceall are saying yes we want that turned on and that fact in itself has nothing to do with a major vs minor change.

jugglinmike:
> To me a breaking change is something like removing an option causing existing
> configs to be invalid

From a user's perspective, a failed build due to a newly-invalidated
configuration file is not much different than a failed build due to a
newly-activated linting rule. They both interrupt the workflow in the same way,
and they both require some intervention to resolve (e.g. pinning to a prior
release or updating the project source).

> "we want everything on, we always want the most strict rules, we don't want
> to maintain a giant config with everything on"

I've heard people cite the first and third reasons for wanting `enforceall`.
[...]

PR #2619, opened 2015-08-16
Update options.js

pansay:
moved funcscope from enforcing to relaxing

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -127,26 +127,6 @@ exports.bool = {
     forin       : true,
 
     /**
-     * This option suppresses warnings about declaring variables inside of
-     * control
-     * structures while accessing them later from the outside. Even though
-     * JavaScript has only two real scopes—global and function—such practice
-     * leads to confusion among people new to the language and hard-to-debug
-     * bugs. This is why, by default, JSHint warns about variables that are
-     * used outside of their intended scope.
-     *
-     *     function test() {
-     *       if (true) {
-     *         var x = 0;
-     *       }
-     *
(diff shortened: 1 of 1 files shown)


caitp:
What's the motivation for this? It seems like a pointless breaking change.

caitp:
Ah i see, might be ok

nicolo-ribaudo:
@pansay Can you sign the [cla](https://www.clahub.com/agreements/jshint/jshint)?

jugglinmike:
Thanks for the patch! Unfortunately, this introduces the same kind of regression described in gh-2048. From that conversation:

> [...] this is not *just* a documentation issue. Moving those options will change the behavior of JSHint for users with the `enforceall` option enabled. Unfortunately, this means we probably shouldn't apply that patch until version 3.0 .
>
> (Note that I have already filed https://github.com/jshint/jshint/issues/2045 to discuss removing `enforceall` completely so we can avoid problems like these in the future.)

In this case, the following code passes on `master` but fails with this patch applied:

```js
// jshint enforceall: true
(function() {
  { var x; }
  x = null;
}());
```

The change is definitely valid, though. Labeling for consideration in the next major release.

pansay:
@jugglinmike you're right, it's not just a documentation issue then, it's a bugfix: since it's false by default, and true with `enforceall`, that means that a user who activated `enforceall`, intending to have *more* strict validation, will in fact have *less* warnings for this case, thus missing potential errors although intending to have the strictest validation possible.

caitp:
Really it's kind of a nonsense warning anyway, but yeah

PR #2623, opened 2015-08-17
[[FEAT]] Warn when invoking a class

### LINKED ISSUE #2579: detect bad class invocations
this... is a typo
```
class A {
}

var b = A();
```

and I think we should warn as it needs to be `new A()`

nicolo-ribaudo:
Fixes #2579

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2433,8 +2433,12 @@ var JSHINT = (function() {
     var p = [];
 
     if (left) {
-      if (left.type === "(identifier)") {
-        if (left.value.match(/^[A-Z]([A-Z0-9_$]*[a-z][A-Za-z0-9_$]*)?$/)) {
+      if (left.type === "class") {
+        warning("W064", left);
+      } else if (left.type === "(identifier)") {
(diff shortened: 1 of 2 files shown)


lukeapage:
LGTM, since I raised the proposal though I'd prefer someone else to chip in and say its a good idea before this is merged.

nicolo-ribaudo:
Note that with decorators classes can be made callable:

```js
function callable(class_) {
	const original = Symbol();
	const fn = function (...args) {
		if (new.target === undefined) {
			return class_.call(...args);
		}
		if (Object.hasOwnProperty.call(new.target, original)) {
			return new class_(...args);
		}
		return new (new.target)(...args);
	};
	Object.defineProperties(fn, {
		name: { value: class_.name, enumerable: false },
		prototype: { value: class_.prototype, enumerable: false },
		[original]: { enumerable: false, configurable: false, writable: false },
	});
	Object.assign(fn, class_);
	return fn;
}

@callable
class A {

	constructor() {
		console.log("constructed");
	}

	static call() {
		console.log("called");
	}

}

A(); // "called"
new A(); // "constructed"
```

PR #2629, opened 2015-08-20
[[FIX]] Warn about using `var` inside `for (...)` when `varstmt: true`

### LINKED ISSUE #2627: var inside for loop with varstmt=true
I set `varstmt` to true, but when I have some code like this:
```javascript
for (var test in [1,2,3,4]) {
  // code here
}
```
JsHint doesn't warm about the use of `var`

nicolo-ribaudo:
Fixes #2627

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3608,7 +3608,7 @@ var JSHINT = (function() {
         lone = true;
       }
 
-      if (!prefix && report && state.option.varstmt) {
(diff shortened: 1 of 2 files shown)


jugglinmike:
Merged to `master` at commit f1ab6383e2926b46aa89c4bf0f635ee71eb01b07. Thanks!

PR #2631, opened 2015-08-20
[[FIX]] Throw W119 instead of "Unexpected '`'" when using templates in ES5 mode

nicolo-ribaudo:
Commit message:
> Previously, linting this code:
```
var a = ``;
```
JSHint threw three errors:
- Unexpected '`'
- Unexpected early end of program
- Unrecoverable syntax error.

> Now it throws "'template literal syntax' is only available in ES6 (use 'esversion: 6')."

So JSHint doesn't stop if it encounters a template literal and esversion is less than 6

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -1056,10 +1056,15 @@ Lexer.prototype = {
     var startChar = this.char;
     var depth = this.templateStarts.length;
 
-    if (!state.inES6(true)) {
-      // Only lex template strings in ESNext mode.
-      return null;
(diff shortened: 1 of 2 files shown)


lukeapage:
LGTM, will merge after the next release

nicolo-ribaudo:
Rebased

jugglinmike:
Merged to `master` at commit 87064e8801dbea77ae66acbac8054d51bd5d10e4. Thanks!

PR #2632, opened 2015-08-20
[[FIX]] Allow latedef in the initialiser of variable

### LINKED ISSUE #2628: Possible latedef regression
@lukeapage The following code issues a warning on `master` (but not v2.8.0):

```js
// jshint latedef: true
var a = function() { return a; };
```

Although we can't statically differentiate patterns like that from patterns like

```js
// jshint latedef: true
var a = function() { return a; }();
```

I think we need to be lax in this case (e.g. forgive all references in the right-hand side of binding statements). Patterns like this are actually fairly common (I discovered it while running the `master` branch against a co-worker's code base):

```js
var myModule = {
  method1: function() { /* etc. */ },
  method2: function() { myModule.method1(); }
};
```

If included in a minor release, this change in application of `latedef` would likely cause trouble for our consumers. I think we ought to be conservative in this case: continue to allow references to the identifier being initialized within the right hand side of a binding. For example:

```js
// jshint latedef: true
void a; // warning
var a;
var b;
void b; // accepted
var c = c; // accepted
```

What do you think?

lukeapage:
Fixes #2628

@jugglinmike as always it wasn't as simple as it seemed as I discovered another minor bug. in master its

```
function(a) {
var a;
return a;
}
```
reports the parameter as unused, when in fact it is used (its not masked by the unnecessary `var a`. I had fixed it for case `function(a) { var a = a; }` but not the above. This now fixes it better. note block scoped vars `function(a) { var b = a; let a;}` will still mark the param as used, but thats okay because its an error to declare block scoped vars with the same name as the parameter.

I started trying to fix `let a = a;` but its probably involved - my start is here - https://github.com/lukeapage/jshint/commit/2b3c5792e0a32167c3c01a12238cdc7e6e01b1f3 I may or may not get time to work on it for a long time.

Feel free to alter this patch as much as you like, you have my permission to use as you wish and get the release out. feel free to finish off my wip if you feel like it.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3521,25 +3521,6 @@ var JSHINT = (function() {
         warning("E012", state.tokens.curr, state.tokens.curr.value);
       }
 
-      // absorb the assignment before processing the label so that use of the label
-      // can be detected as before declaration
-      if (state.tokens.next.id === "=") {
-        advance("=");
-        if (!prefix && state.tokens.next.id === "undefined") {
-          warning("W080", state.tokens.prev, state.tokens.prev.value);
-        }
-        if (!prefix && peek(0).id === "=" && state.tokens.next.identifier) {
-          warning("W120", state.tokens.next, state.tokens.next.value);
-        }
-        // don't accept `in` in expression if prefix is used for ForIn/Of loop.
-        value = expression(prefix ? 120 : 10);
-        if (lone) {
-          tokens[0].first = value;
-        } else {
-          destructuringPatternMatch(names, value);
-        }
-      }
-
       for (var t in tokens) {
         if (tokens.hasOwnProperty(t)) {
           t = tokens[t];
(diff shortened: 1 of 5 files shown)


lukeapage:
> I started trying to fix let a = a; but its probably involved - my start is here - lukeapage@2b3c579 I may or may not get time to work on it for a long time.

Its definitely difficult to get right, I don't know if I'll ever fix it. Added another sketching commit on this branch -  https://github.com/lukeapage/jshint/tree/fix-var-a-eq-a

src/jshint.js
```diff
@@ -3561,6 +3542,25 @@ var JSHINT = (function() {
         }
       }
 
+      // absorb the assignment before processing the label so that use of the label
```
jugglinmike: It seems  like this comment (here and below) is now inaccurate--it's describing the behavior that this change removes. Is that right? If so, with your permission, I'll remove it from the patch when I merge.

tests/unit/options.js
```diff
@@ -736,11 +736,18 @@ exports.undef = function (test) {
 
   // block scope cannot use themselves in the declaration
   TestRun(test)
-    .addError(1, "'a' was used before it was declared, which is illegal for 'let' variables.")
-    .addError(2, "'b' was used before it was declared, which is illegal for 'const' variables.")
+//    .addError(1, "'a' was used before it was declared, which is illegal for 'let' variables.")
```
jugglinmike: I'd like to add an explicit `TODO` here to more clearly describe the intention of the commented-out assertions. Something like, "JSHint does not currently enforce the correct temporal dead zone semantics in this case." Sound good?

lukeapage:
I agree with your comments, though i would probably delete the test and raise an issue instead. 
I commented because i thought id be able to fix it. 
im out of time (shouls be a dad in the next 24 hours!) so please make whatever changes you like to get this merged, I wont be able to for a bit.

jugglinmike:
Merged to `master` with the discussed changes: 18f8775a05c10fc8c157ef7b79c16d36ba0bbc7b

I've opened gh-2637 to follow up on the TDZ enforcement issue. I've kept the commented-out assertions and included a comment in order to avoid confusing future contributors who are fixing that bug.

Congratulations, Luke! Best of luck with the sleepless nights to follow :P

Issue #2459, opened 2015-06-11
Warn when using let but not mutating


lukeapage:
Probably needs a new option and probably also needs to be used with the "varstmt" option.

```
function name() {
    let obj = { x: true };  // warn that obj has not been re-assigned, so should be `const`
    let i = 1;                  // do not warn
    i++;
    return obj.x;
}
```

PR #2633, opened 2015-08-21
[[FEAT]] Implement `preferconst` option

nicolo-ribaudo:
Do you think the `preferconst` name is ok?

#2459


lukeapage:
i like this rule alot.

src/options.js
```diff
@@ -236,6 +236,21 @@ exports.bool = {
...
+     * This option prohibits the use of `let` for variables which don't mutate.
+     * Using `const` would make the code easier to review.
+     *
+     *     // jshint preferconst: true
+     *
+     *     let a = 2; // a is never modified, so JSHint warns.
+     *     let b = a;
+     *     b++;
+     *
+     * This option doesn't warn when using `var` but not mutating because `var`
```
lukeapage: could be helpful to reference the `varstmt` option

src/scope-manager.js
```diff
@@ -280,12 +280,25 @@ var scopeManager = function(state, predefined, exported, declared) {
       var i, j;
       var currentUsages = _current["(usages)"];
       var currentLabels = _current["(labels)"];
+      var labelNameList = Object.keys(currentLabels);
```
lukeapage: `labelNameList` is only used by `preferconst` so I'd put it inside the if-block.
My reasoning for these kind of changes is that now `let` exists, one day this code might be made to es6 and if you were using `let`, you'd probably do that?

lukeapage:
> Do you think the preferconst name is ok?

I think its a good name

lukeapage:
I've thought about this for a bit and I think it looks good.

I had one minor concern ;-
```
let [a, b] = [1,2]; // warns about `b`
a++;
```
The above code, espescially with more complex destructuring you might not want to prefer a const in.

But after thinking a bit, I think it's fine, because the case is not going to be that common and b) you could argue it makes the code clearer to extract all the destructured variables as const and then put into a new mutated variable if you need one (and defaults can be handled in the destructuring).

nicolo-ribaudo:
>
```js
let [a, b] = [1,2]; // warns about `b`
a++;
```

I think JSHint shouldn't warn at all or it should warn if both `a` and `b` should be `const`

PR #2639, opened 2015-08-25
[[DOCS]] Link to inline configuration from globals option

davidjb:
For users reading the JSHint Options documentation, there is no indication on this page about how to use globals as inline configuration.

It's easy and explained clearly if you arrived at the documentation on main Docs page, but not if you arrived at Options first (eg a search for "jshint options").  Hence, a simple link to direct people to that section of the Docs page.


jugglinmike:
Good idea, and thank you!

Merged to `master` at commit 325c379642df77f7c79cf98ca95b409f0014cd9d.

PR #2640, opened 2015-08-25
in the comments, changing version to 2.8.0

tisalph:


nicolo-ribaudo:
I think this PR isn't need, because the version will be increased to 2.9.0 in the next release

tisalph:
Ok, thank you.

PR #2645, opened 2015-08-26
[[FIX]] Allow binary and octal numbers, and templates when using inline esnext

nicolo-ribaudo:
This commit fixes a regression introduced in gh-2519


jugglinmike:
Merged to `master` at commit b5ba7d63c72777eb488df392bd6185f060c38f06. Thanks!

PR #2649, opened 2015-08-27
Fix overriding of comma function with var

gtanner:
There is already a comma function (https://github.com/jshint/jshint/blob/master/src/jshint.js#L1010), this would override that value and cause errors when attempting to execute the function.

Renamed to firstcomma to prevent this problem


gtanner:
I have signed the CLA @lukeapage

jugglinmike:
Thanks for the patch! This is actually a problem elsewhere in the codebase. Would you have any interest in fixing it generally? Then we could set `shadow: outer` for the project's `.jshintrc` and be sure we never run across this problem again.

gtanner:
Hey @jugglinmike, good idea.  I rebased the commit to enable that rule and fix all of the occurrences of that bug.

gtanner:
Just checking in @jugglinmike and @lukeapage.

Anything else I need to do to get this merged in?

caitp:
I've just gone over the diff, and all i see are renamed variables/properties/parameters. Worse than that, i don't particularly agree with the renaming.

I'd rather the mass renaming didn't happen, and that this patch only included functional changes

src/jshint.js
```diff
@@ -4225,7 +4225,7 @@ var JSHINT = (function() {
     var i = 0;
     var inof = ["in", "of"];
     var level = 0; // BindingPattern "level" --- level 0 === no BindingPattern
-    var comma; // First comma punctuator at level 0
+    var firstcomma; // First comma punctuator at level 0
```
gtanner: I don't really care about the mass renaming.  this rename is needed because it is causing issues when the code overwrites the comma function.

All I need is this rename merged in.

gtanner: The mass renaming was based off of @jugglinmike's feedback, which is valid because it looked like there was a few other places where global functions will be overwritten and cause jshint to not work

caitp: That may be, but I don't think that's the right approach to fixing it --- the reason is that the old names were more descriptive and easier to understand.

A better approach would be installing all of the global functions as properties on an object, so that they can be fully qualified when needed

gtanner: How about we merge this in to fix the bug (or just my one rename to fix the issue with comma).  Then maybe someone in the future can do that refactoring.

All I want is jshint to work again and I can update without repatching the comma fix in every version.

jugglinmike: > All I want is jshint to work again and I can update without repatching the comma fix in every version.

My apologies: until you wrote this, I didn't realize that JSHint was actually non-functional for you. Up until now, I thought you were interested in improving the readability of the source--that is why I have treated it as lower-priority than the work I have been doing recently. The fact that the change is necessary to fix a problem changes things a bit.

It also makes me curious, though: all of the project's unit tests are passing *without* this change (or your original change) in place. Can you explain more about the error you are experiencing (and how you are running JSHint itself)?

gtanner: Hey sorry, 

We use this at http://www.bithound.io/ and I have seen it on a few private repos that we analyze.

We don't do anything magic, just call the JSHINT msg in our code:

```
if (!JSHINT(source + '\n', config, config.globals)) {
      JSHINT.errors.forEach(function (error) {
        if (error) {
          errors.push(error);
        }
      });
    }
}
```

jugglinmike: That's very strange! Is there any chance you can share a script that demonstrates the crash? I'd like to turn that into a unit test so we can be sure to avoid this problem in the future.

gtanner: Since it was on a private repository that we analyzed on production I don't have access to the actual code that was causing this.

But you can see the offending call to the `comma` function https://github.com/gtanner/jshint/blob/master/src/jshint.js#L4362 that can crash before the rename of the local var to `firstcomma`.

jugglinmike:
@gtanner It's no problem about not being able to share the full source. Actually, in
this case, it would be more useful to have a simplified (read: original and
safe-to-share) code sample.

The good news is that the information that you have shared has been enough to
reverse engineer a test case. The following input will crash JSHint:

```js
for ( 0,
      ,; ; ) {}
```

```js
for ( ; ; 0,
    ,) {}
```

This code is not valid, but (as you've recognized) invalid code should produce
a syntax error without crashing the program. Your fix correctly mitigates this
crash, so thanks!

With that mystery solved, I've been able to focus on the substance of the
variable renaming. I tend to agree with @caitp's feedback: many of the new
names leave something to be desired. I have specific thoughts about better
names, but it's all pretty subjective and shouldn't block your bug fix.

Here's what I have in mind:

Could you limit the patch to *only* address the overloading of the `comma`
binding (and add the test cases I shared above)? In this case, I think it makes
more sense to rename the "outer" `comma` to `parseComma`--as a function value,
a verb name more clearly communicates the type.

I recognize that it was my idea to expand the scope of your patch, but I don't
want to throw your additional work away! From there, we could collaborate on a
follow-up patch to find the best names for the remaining shadowed bindings.

Breaking the change apart like this would help us to get your fix in the
(imminent) 2.9.1 release. I'd understand if you've had just about enough of all
this; if you only want to do the first part, I'd be happy to follow up on my
own. In any event, I appreciate the patience and effort you've demonstrated so
far!

What do you say?

gtanner:
Hey @jugglinmike, rebased with just the comma change.

Thanks for trying to get this in.

nicolo-ribaudo:
I think @jugglinmike asked to rename the `comma` *function* to `parseComma`, not the `comma` var.

gtanner:
Good catch @nicolo-ribaudo, I misread the instructions.  Updated the PR

nicolo-ribaudo:
@gtanner Can you also add a test (in tests/unit/parser.js) using the code
jugglinmike wrote?


-- 
Nicolò Ribaudo

jugglinmike:
@gillesruppert This looks good, but we still need those tests in there. If you like, I can add them myself in a separate commit. I'll do this if I don't hear back from you in the next day or so.

gtanner:
Thanks @jugglinmike and @gillesruppert, I don't think I can carve off some time to get those tests in.

gtanner:
tests added @jugglinmike, @gillesruppert

gtanner:
ping

jugglinmike:
Hey @gtanner, thanks for following up on this. While waiting for those tests, we published a release candidate. I'd like to avoid any changes that aren't directly related to regressions until the stable version is published, which includes this bug fix. This should be "good to go" for the next cycle, though!

gtanner:
So following up since 2.9.1 just landed, I am assuming you will be merging this in now @jugglinmike ?

jugglinmike:
Don't worry, @gtanner, I haven't forgotten about your patch! I added the "Good to land" label when we last spoke to make sure of that. Anyway, this changeset now conflicts with `master`, so either you or I will need to resolve that before this can be accepted.

gtanner:
Probably best a core member to do the merge so I don't trigger the need for another review or introduce a regression with a bad rebase/merge

jugglinmike:
The conflict (introduced by 767c47d6482a577d7e1a3fbfad664b6a957fe56e) was easy to resolve; I just updated the indentation of your change. Beyond that, I squashed your commits into one and updated the message according to [our contribution guidelines](https://github.com/jshint/jshint/blob/2e0026f8f461425bb05da966570708002a9f4f0a/CONTRIBUTING.md#commit-message-guidelines).

The fix is now in `master` at commit 2e0026f8f461425bb05da966570708002a9f4f0a. Thanks for your help!

Issue #2657, opened 2015-09-03
[Regression] Function scope of "use strict" is now required even in node mode


mzgol:
As of JSHint 2.9.0 function scope of `"use strict"` is now required even in `node` mode. This is breaking current jQuery `master` build so it's pretty blocking for us.

TuckerWhitehouse:
+1

ice2morrow:
+1. Breaking change

andyk1989:
+1

andyk1989:
Looks like https://github.com/jshint/jshint/blob/9e33d37739ddcb8021d15a2c8a70b2ee7df30e88/src/jshint.js#L5290 is the culprit

It is no longer checking for `state.option.node` and other values as it was in v2.8.0 https://github.com/jshint/jshint/blob/2.8.0/src/jshint.js#L5377

andrespr:
+1

tran2:
+1

nightpool:
+1

ice2morrow:
Is there any updates on this?

lukeapage:
@ice2morrow the linked pr is under review

Issue #2661, opened 2015-09-03
Regression: Use the function form of "use strict".


Raynos:
We depend on jshint with `^` and our travis builds started failing complaining about use function form of "use strict".

We have the following set:

```
{
    "strict": false,
    "globalstrict": true,
    ...
}
```

This is a breaking change and will probably break other people as well.

Raynos:
looks like 8de8247 is a regression

```
-        if (state.isStrict()) {
-          if (!state.option.globalstrict) {
+        if (state.directive["use strict"]) {
+          if (state.option.strict !== "global") {
```

Raynos:
There are more regressions:

```
> tchannel@2.8.1 lint /home/raynos/uber/tchannel/node
> jshint .

benchmarks/bench_server.js: line 114, col 9, Reassignment of 'onPing', which is is a function. Use 'var' or 'let' to declare bindings that may change.
benchmarks/bench_server.js: line 117, col 9, Reassignment of 'onSet', which is is a function. Use 'var' or 'let' to declare bindings that may change.
benchmarks/bench_server.js: line 120, col 9, Reassignment of 'onGet', which is is a function. Use 'var' or 'let' to declare bindings that may change.

test/hyperbahn/advertise.js: line 88, col 13, 'client' was used before it was defined.
```

Please dont publish a new minor version with 3 months worth of commits -.-

lukeapage:
@Raynos please raise the other "regressions" in a different issue with code samples. it is possible some of those new warnings are rule violations that were not previously detected ratger than regressions.

lukeapage:
possible duplicate of https://github.com/jshint/jshint/issues/2657

nightpool:
I think this is the inverse of #2657—code that wasn't strict, beforehand, is now getting interpreted as strict. 

I think. @0xazure does that make sense?

lukeapage:
@Raynos We have fixed the reported regression with use strict.
Investigating your other regressions, looking at https://github.com/uber/tchannel/blob/master/node/ - the advertise.js isn't in the repo so I cannot comment on that, but the new errors in bench_server are as I suggested, warnings that were not given previously.

E.g. in 2.8.0

```
function a() {
}
a = 1; // warns
b = 1; // does not warn
function b() {
}
```
 in 2.9.0
```
function a() {
}
a = 1; // warns
b = 1; // warns
function b() {
}
```

it is a difficult line between a regression and a bugfix but in this case we decided it was a bugfix and should now warn.

[...]

PR #2469, opened 2015-06-16
[[FEAT]] Option to assume strict mode

nicolo-ribaudo:
See https://github.com/jshint/jshint/issues/924#issuecomment-16981384

Issue #924, opened 2013-03-13
Option to assume strict mode (for nodejs)


coolaj86:
Now that nodejs is es5 compliant it can be run in strict mode by default without including `(function () { "use strict"; /* code here */ }());`.

Will you add an option `implicitstrict` so that it's not required to have the boilerplate code?

englercj:
Nearly two years since this was opened, is this actually implemented?

adriengibrat:
It's not yet implemented, a [patch](https://github.com/Demeterr/jshint/commit/7b3589f2f33934525f937b207108d9c548100020) was refused because it was mixed with other unrelated stuffs.

Please, it would be nice to implement what @valueof had in mind: [```strict```  multivalue option](https://github.com/jshint/jshint/issues/924#issuecomment-16981384)

englercj:
Any word on when we will have a release with this in it? I've been waiting for it for a while, and it looks like it landed in master.

lukeapage:
@jugglinmike can tell you,  I think one is due soon.

mik01aj:
This gets very annoying with Babel, as it [puts 'use strict'; in front of everything](https://github.com/babel/babel/issues/1844) and explicit `'use strict'` declarations are not necesssary anymore. @lukeapage, could we have a new `2.8.1` release?

lukeapage:
@jugglinmike is working hard to get the next release out. It has alot of
work in it and a few regressions were identified which slowed things down.
we hope it will be soon.

PR #2663, opened 2015-09-03
[[FIX]] `use strict`-related regressions.

nicolo-ribaudo:
Fixes gh-2657 and gh-2661

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -5286,7 +5286,10 @@ var JSHINT = (function() {
         directives();
 
         if (state.directive["use strict"]) {
-          if (state.option.strict !== "global") {
+          if (state.option.strict !== "global" &&
+              !((state.option.strict === true || !state.option.strict) &&
+                (state.option.globalstrict || state.option.module || state.option.node ||
+                 state.option.phantom || state.option.browserify))) {
             warning("W097", state.tokens.prev);
           }
         }
--- a/tests/unit/envs.js
+++ b/tests/unit/envs.js
@@ -86,6 +86,13 @@ exports.node = function (test) {
     .addError(1, "Read only.")
     .test(overwrites, { es3: true, browserify: true });
 
+  TestRun(test, "gh-2657")
+    .test("'use strict';var a;", { node: true });
+
+  // Implied `strict: global` if `strict` is `true`
+  JSHINT("", { node: true, strict: true });
+  test.strictEqual(JSHINT.data().options.strict, "global");
+
   test.done();
 };
 
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -1685,6 +1685,9 @@ exports.globalstrict = function (test) {
   code = '(function (test) { "use strict"; return; }());';
   TestRun(test).test(code, { globalstrict: true, strict: true });
 
+  TestRun(test, "gh-2661")
+    .test("'use strict';", { strict: false, globalstrict: true });
+
   test.done();
 };
 


src/jshint.js
```diff
@@ -213,6 +213,8 @@ var JSHINT = (function() {
     if (state.option.module) {
       if (state.option.strict === true) {
         state.option.strict = "global";
+      } else if (!state.option.strict) {
+        state.option.strict = "global?";
```
lukeapage: should we document `global?` or was the intention that it be internal only?
What does it mean - I'm wondering if it would be better having a different word e.g. `ignore` or `global-implied` or something else?

nicolo-ribaudo: I've renamed `global?` to `ignore` and documented it.

lukeapage:
LGTM but will leave for @jugglinmike for a 2nd opinion.

jugglinmike:
Even after reviewing this, I'm having trouble understanding what `strict: ignore` actually means. The overlap with `strict: false` make it seem more like an afterthought than an intentionally-designed feature. To me, the weak distinction detracts from the usability of the option as a whole.

I also feel reluctant to allow a regression to motivate a new feature. I know it wasn't originally implemented this way, and that @lukeapage suggested that we expose this and document. Based on his wording, though, I don't think he was entirely sure of that idea.

That said, there's definitely some tension in supporting an option value for internal use only. It makes the code harder to follow, and consumers might be confused when they see one value in the source but are unable to specify it directly.

An alternative way to resolve this tension is to avoid introducing a new option value altogether (whether "public" or "private"). For instance, we might be able to set a flag on the `state` object describing this distinction and reference as appropriate. (I don't really care for spit-balling when it comes to reviewing the contributions of others--my inclination would be to prototype something as a more constructive way to share feedback. But that would take time I don't have at the moment, and @lukeapage has encouraged me to lighten up on that, so here we are.)

What do you two think?

lukeapage:
typo.. 

Use `strict: global` is `strict` was `true`,

should be if.

lukeapage:
I wondered if it should be exposed if it was a valid behaviour that people want..  e. g.  globalstrict is deprecated so we are  removing the option..

In addition as you described I didn't like having an option that wasn't documented,  so that was my other motivation, if it should not be exposed I agree state is a nicer way of doing it. 

Does ignore mean we are in strict mode all the time and don't warn about anything,  whereas false is not in strict mode,  don't warn about anything? If so would ignore => implied-ignore be better and have ignore === false as another option? or should we go back and conside not deprecating globalstrict?

jugglinmike:
Specifying `globalstrict: true` simply allows you to include the `"use strict";` directive in the global scope. This is the same behavior as the recently-implemented `strict: global`, so deprecating `globalstrict` still seems like a good idea. Or am I misunderstanding some subtlety here? If not, then I think the current set of values already contains all the behavior for which users have expressed interest (which is why I didn't want to expose another value).

nicolo-ribaudo:
As you said, it *allows* (not *requires*).

`strict: global` is like `strict: true, global strict: true` (You have to use the "use strict" directive, which can be in the global scope), but there isn't a `strict` value which allows you to put that directive in the global scope,  but doesn't warn if it is missing (`strict: false, globalstrict: true`)

jugglinmike:
@nicolo-ribaudo I'm starting to get confused.

> As you said, it *allows* (not *requires*).

That is what I'm seeing on `master` right now, where the global `"use strict"`
directive is normally disallowed:

```
$ ./bin/jshint -
'use strict';
a = 1;
stdin: line 1, col 1, Use the function form of "use strict".
stdin: line 2, col 1, 'a' is not defined.

2 errors
```

And while with `strict: global`, the global `"use strict"` directive is not
*required*:

```
$ ./bin/jshint -
// jshint strict: global
a = 1;
```

...that option value makes it *allowed*:

```
$ ./bin/jshint -
// jshint strict: global
'use strict';
a = 1;
stdin: line 3, col 1, 'a' is not defined.

1 error
```

Your follow-up description doesn't match those observations:

> `strict: global` is like `strict: true, global strict: true` (You have to use
> the "use strict" directive, which can be in the global scope), but there
> isn't a `strict` value which allows you to put that directive in the global
> scope, but doesn't warn if it is missing (`strict: false, globalstrict:
> true`)

lukeapage:
@jugglinmike 
```
// jshint strict: global
a = 1;
```
produces
```
test.js: line 2, col 1, 'a' is not defined.

1 error
```
As you said, but

```
// jshint strict: global
a = 1;
var a;
```

produces

```
test.js: line 3, col 6, Missing "use strict" statement.

1 error
```

E.g. just putting an assignment on a line for some reason doesn't trigger jshint into giving the warning.

I tried
```
// jshint strict: global
var a;
```
and that works.

We should file this as a separate bug.

nicolo-ribaudo:
What do you think about this? nicolo-ribaudo/jshint@48018b411e28abd806bb53013cfcc38720bf72ee

jugglinmike:
> or should we go back and conside not deprecating globalstrict?

The way I see it, the basic question we need to answer is this: Does anyone want to *allow* the global "use strict" directive without *enforcing* it? When we deprecated the "globalstrict" option and introduced `strict: global` (via https://github.com/jshint/jshint/pull/2469), we decided, "no, no one wants to do that". I can't find any discussion that substantiates this decision, though; the closest thing I can find is [this passing suggestion by Anton from two years ago](
https://github.com/jshint/jshint/issues/924#issuecomment-16981384)--hardly a sign of careful design!

Since feature deprecation is a big deal, I'd like to make sure we're all on-board with that change. For my part, I agree. The use case of "allow but don't enforce" (which is currently served by `globalstrict: true, strict: false`) seems inconsistent and uncommon. In a `strict: global` world, users who actually desire this behavior could achieve it by setting `strict: func` for their whole project and then `-W097` for the specific files that they wish to use a global "use strict" directive. This is definitely more cumbersome than the current method, but I think that's acceptable for two reasons:

1. It's an uncommon use case
2. It's more explicit

@lukeapage @rwaldron I consider this to be the last issue to clear up before moving forward with the release process. Can you please let me know if you agree with my reasoning?

lukeapage:
> In a strict: global world, users who actually desire this behavior could achieve it by setting strict: func for their whole project and then -W097 for the specific files that they wish to use a global "use strict" directive.

or setting the correct `//jshint strict: global` at the top of the file that has a different pattern

> It's more explicit

I think thats a good reason.

> I consider this to be the last issue to clear up before moving forward with the release process. Can you please let me know if you agree with my reasoning?

yes, sorry if I confused things earlier when I didn't really grasp the issue.

If we agree we still have to do one thing - make sure `//jshint strict: false, globalstrict: true` resolves to `//jshint strict: false` ? Or should we ask people who have this setting to remove the `globalstrict: true` as it is confusing (for them and us) ?

nicolo-ribaudo:
I can fix gh-2657 and gh-2661 without introducing the `ignore` value (see nicolo-ribaudo/jshint@48018b4) so that JSHint 2.9.1 can be released.
Then you can open a new issue where to continue this discussion.

jugglinmike:
Not exactly. We're discussing the viability of a new feature, so we really shouldn't release until there's consensus about that feature (otherwise, we risk creating technical debt by exposing a new option that we decide to deprecate soon after).

Don't worry though: I pinged @rwaldron earlier today, so we'll be able to move the discussion forward soon.

src/options.js
```diff
@@ -365,7 +365,8 @@ exports.bool = {
      *
      * For more info about strict mode see the `strict` option.
      *
-     * @deprecated Use `strict: "global"`.
+     * @deprecated Use `strict: global` is `strict` was `true`,
+     *             or `strict: ignore` is it was `false`.
```
rwaldron: I don't follow this description

nicolo-ribaudo: It should be `Use 'strict: ignore' to allow a global "use strict" directive or 'strict: global' to require it.`

rwaldron:
@jugglinmike and I discussed this today and agree that `strictmode: global|function|implied` is a solid path forward, however _any_ overlapping use of `globalstrict` will fail hard and fast (for those accidental cases). `globalstrict` should be maintained as-is until 3.x, then dropped.

nicolo-ribaudo:
Should I cherry-pick the commit I linked? (It fixes the issue without introducing the `ignore` value)

jugglinmike:
Sounds good to me!

nicolo-ribaudo:
I removed the `strict: ignore` commits and cherry-picked the new one.

jugglinmike:
Merged to `master` at 04b43d22755a766311f3ed12a2e6ce51d6ac4a3a . Thanks!

PR #2665, opened 2015-09-04
[[FIX]] Don't warn that an exported function is used before it is defined

### LINKED ISSUE #2658: latedef:true erroneously flags named exports as used before define
Version: 2.9.0, regressed from ~2.8.0

```
/*jshint latedef: true, esnext: true*/
export function hi() {
 return 1;
}
```

results in:

```
2	'hi' was used before it was defined.
```

nicolo-ribaudo:
Fixes #2658

### DIFF
--- a/tests/unit/fixtures/latedef-esnext.js
+++ b/tests/unit/fixtures/latedef-esnext.js
@@ -65,3 +65,4 @@ export function de() {}
 export class df {}
 export var dg;
 export let dh;
+export const di = {};
--- a/tests/unit/fixtures/latedef-esnext.js
+++ b/tests/unit/fixtures/latedef-esnext.js
@@ -60,3 +60,8 @@ let dc = {
         return dc;
     }
 };
+
+export function de() {}
+export class df {}
+export var dg;
+export let dh;
--- a/tests/unit/fixtures/latedef-esnext.js
+++ b/tests/unit/fixtures/latedef-esnext.js
@@ -61,6 +61,7 @@ let dc = {
     }
 };
 
+// Regression test for gh-2658
 export function de() {}
 export class df {}
 export var dg;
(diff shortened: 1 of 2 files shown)


lukeapage:
LGTM

tests/unit/fixtures/latedef-esnext.js
```diff
@@ -60,3 +60,8 @@ let dc = {
         return dc;
     }
 };
+
+export function de() {}
+export class df {}
+export var dg;
+export let dh;
```
mikesherov: Maybe add one for const?

lukeapage: yes, even though let and const now share the same code, there is no reason to leave it off, sorry I missed that.

nicolo-ribaudo: Done

mikesherov:
LGTM

tests/unit/fixtures/latedef-esnext.js
```diff
@@ -60,3 +60,9 @@ let dc = {
         return dc;
     }
 };
+
```
jugglinmike: Could you add a comment here referencing the regression at gh-2658 so future contributors understand the purpose of these statements?

jugglinmike:
@nicolo-ribaudo This looks good! Just one documentation-related thing, then this should be good to go.

tests/unit/fixtures/latedef-esnext.js
nicolo-ribaudo: Done.

jugglinmike:
Squashed and merged to `master` at commit d0433d23e1a48c74522b9ed32b87c45624bef4c5. Thanks, @nicolo-ribaudo!

PR #2666, opened 2015-09-04
[[FIX]] Support property assignment when destructure assigning

### LINKED ISSUE #2659: array destructuring assignment to an object property errors incorrectly
Regression in 2.9

```
let a = {};

[a.b] = ['b'];
```

yields

```
Expected ',' and saw '.'
```

### LINKED ISSUE #2660: array destructuring assignment to an object property where the object is a const errors incorrectly
Regression is 2.9

```
const a = {};

[a.b] = ['b'];
```

yields:
```
Attempting to override 'a' which is a constant.
```

lukeapage:
Fixes #2659 and fixes #2660

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3384,7 +3384,7 @@ var JSHINT = (function() {
           if (!identifierToken.identifier) {
             warning("E030", identifierToken, identifierToken.value);
           }
-          var assignTarget = expression(150);
+          var assignTarget = expression(155);
           if (assignTarget && assignTarget.identifier) {
             ident = assignTarget.value;
           }
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1370,7 +1370,9 @@ var JSHINT = (function() {
     } else if (left.id === "{" || left.id === "[") {
       if (allowDestructuring && state.tokens.curr.left.destructAssign) {
         state.tokens.curr.left.destructAssign.forEach(function(t) {
-          state.funct["(scope)"].block.modify(t.id, t.token);
+          if (t.id) {
+            state.funct["(scope)"].block.modify(t.id, t.token);
+          }
         });
       } else {
         if (left.id === "{" || !left.left) {
@@ -3356,12 +3358,14 @@ var JSHINT = (function() {
     var ids;
     var identifiers = [];
     var openingParsed = options && options.openingParsed;
+    var isAssignment = options && options.assignment;
+    var recursiveOptions = isAssignment ? { assignment: isAssignment } : null;
     var firstToken = openingParsed ? state.tokens.curr : state.tokens.next;
 
     var nextInnerDE = function() {
       var ident;
       if (checkPunctuators(state.tokens.next, ["[", "{"])) {
-        ids = destructuringPatternRecursive();
+        ids = destructuringPatternRecursive(recursiveOptions);
         for (var id in ids) {
           id = ids[id];
           identifiers.push({ id: id.id, token: id.token });
@@ -3374,9 +3378,22 @@ var JSHINT = (function() {
         advance(")");
       } else {
         var is_rest = checkPunctuator(state.tokens.next, "...");
-        ident = identifier();
-        if (ident)
+
+        if (isAssignment) {
+          var identifierToken = is_rest ? peek(0) : state.tokens.next;
+          if (!identifierToken.identifier) {
+            warning("E030", identifierToken, identifierToken.value);
+          }
+          var assignTarget = expression(150);
+          if (assignTarget && assignTarget.identifier) {
+            ident = assignTarget.value;
+          }
+        } else {
+          ident = identifier();
+        }
+        if (ident) {
           identifiers.push({ id: ident, token: state.tokens.curr });
+        }
         return is_rest;
       }
       return false;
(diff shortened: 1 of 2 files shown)


lukeapage:
comments welcome (and ideas for any more tests) - I got it working but I don't know if there is a better way.

nicolo-ribaudo:
This code is allowed by this pr, while it shouldn't:
```js
let a;
[a()] = [];
```

lukeapage:
Well spotted.

> What do you think about using checkLeftSideAssign?

I've fixed the bug by setting the precedence to 155 rather than 150 and I've verified no other operators have a precedence higher. But I think you are right, there are checks in `checkLeftSideAssign` that should be being run. I'll handle this next.

lukeapage:
> But I think you are right, there are checks in `checkLeftSideAssign` that should be being run. I'll handle this next.

done - there were a couple of bugs because of this even in master (e.g. `({e} = {});` in a try-catch).

mikesherov:
:+1: AFAICT

tests/unit/parser.js
```diff
@@ -1643,7 +1646,48 @@ exports["destructuring const errors"] = function (test) {
...
+    .addError(5, "Missing semicolon.")
+    .addError(5, "Expected an identifier and instead saw ']'.")
+    .addError(5, "Expected an assignment or function call and instead saw an expression.")
+    .addError(5, "Missing semicolon.")
+    .test(code, {esnext: true, unused: true, undef: true});
+
+  test.done();
+};
+
+exports["destructuring const errors"] = function (test) {
```
jugglinmike: Looks like some weird copy-paste behavior might have tripped you up here--this test is identical to the one that precedes it. It's safe to remove.

src/jshint.js
```diff
@@ -3374,9 +3380,27 @@ var JSHINT = (function() {
...
+          var identifierToken = is_rest ? peek(0) : state.tokens.next;
+          if (!identifierToken.identifier) {
+            warning("E030", identifierToken, identifierToken.value);
+          }
+          var assignTarget = expression(155);
+          if (assignTarget) {
+            checkLeftSideAssign(assignTarget);
+
+            // if the target was a simple identifier, add it to the list to return
+            if (assignTarget.identifier) {
```
jugglinmike: The converse of this condition is not currently tested. In other words, this also passes:

```diff
-// if the target was a simple identifier, add it to the list to return
-if (assignTarget.identifier) {
   ident = assignTarget.value;
-}
```

...but that's clearly wrong. I'm having trouble thinking of a test case for that, though. Any ideas?

lukeapage:
> The converse of this condition is not currently tested. In other words, this also passes:

oh,  really? for `[ a.b] = [] ` isnt assignTarget the '.' with a left and right identifier? maybe value is not set on it which means it would be ignored,  explaining why it passes tests with if commented out? 

I've been trying not to return null identifiers from this function but still guarding against them. 

I will debug and double check tomorrow, just mentioning in case this explains something for you.

tests/unit/parser.js
lukeapage: woops,  I was going to split the test but I changed my  mind.  looks like i didn't undo far enough

caitp:
I haven't read this yet, but a good test case ( for both destructuring assignment and vanilla assignment) is `new.target` for assignment target. Definitely want to warn about that even though it is an early error in VMs

lukeapage:
yep testing that :)

tests/unit/parser.js
lukeapage: fixed.

lukeapage:
So, if I change the code to

```
// if the target was a simple identifier, add it to the list to return
if (assignTarget.identifier) {
    ident = assignTarget.value;
} else {
    console.log("not an identifier");
    console.log("value =", assignTarget.value);
}
```

and run the tests I get lots of 

```
not an identifier
value = .
not an identifier
value = .
not an identifier
value = [
not an identifier
value = [
not an identifier
value = ...
```

which initially confused me because if I remove the if, things still pass.. I would expect lots of `'[' is not defined.`. 
However the usage of that identifier list only does
```
if (t.id) {
    state.funct["(scope)"].block.modify(t.id, t.token);
}
```
and modify doesn't "use" the variable - that is done when an identifier is parsed.

So I've added a couple of tests that make sure we get a `'x' is not defined` warning when destruct assigning and that `[...i] = [1]` uses `i`.

jugglinmike:
Let me see if I'm understanding this correctly: due to the way the `destructuringPattern` function's return value is used, it currently makes no difference if it erroneously includes entries for non-identifier tokens. Is that right?

lukeapage:
yes.

I could change that by making scope managers modify throw an error if it is
not first used, but I'm reluctant to add an assertion unless it only
asserts during tests (in case it introduces a bug).

jugglinmike:
Yeah, I'd be leery of bugs and also of the assertion's effect on performance. We should look into a build step for the main release--that would let us strip out assertions (and it would also pave the way for other transformations, like ES6 syntax).

That's work for another day, though. I've squashed these commits into one and merged that to `master` at revision b6df1f22633ee597df433591b0fcf925c98801f0. Thanks, Luke!

PR #2667, opened 2015-09-04
[[FIX]] `latedef` shouldn't warn when marking a var as exported

### LINKED ISSUE #2662: Exported var was used before it was defined in 2.9.0
Using JSHint 2.9.0, I get this error:  `common-lib.js: line 2, col 5, 'commonLib' was used before it was defined.` and in previous versions of JSHint I did not.  Moving the `/* exported ... /*` line to the end of the file resolves the symptom.  Here is a minimum sample repro project:

.jshintrc:
```
{
  "latedef": true
}
```

common-lib.js:
```
/* exported commonLib */
var commonLib = (function () {
  'use strict';

  return 'the common lib';
}());
```

command I run:
`jshint .`

nicolo-ribaudo:
Fixes #2662

### DIFF
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -65,16 +65,6 @@ var scopeManager = function(state, predefined, exported, declared) {
     }
   }
 
-  function _addUsage(labelName, token) {
-
-    _setupUsages(labelName);
-
-    if (token) {
-      token["(function)"] = _currentFunctBody;
-      _current["(usages)"][labelName]["(tokens)"].push(token);
-    }
-  }
-
   var _getUnusedOption = function(unused_opt) {
     if (unused_opt === undefined) {
       unused_opt = state.option.unused;
@@ -772,7 +762,12 @@ var scopeManager = function(state, predefined, exported, declared) {
           token.ignoreUndef = true;
         }
 
-        _addUsage(labelName, token);
+        _setupUsages(labelName);
+
+        if (token) {
+          token["(function)"] = _currentFunctBody;
+          _current["(usages)"][labelName]["(tokens)"].push(token);
+        }
       },
 
       reassign: function(labelName, token) {
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -75,11 +75,6 @@ var scopeManager = function(state, predefined, exported, declared) {
     }
   }
 
-  var exportedLabels = Object.keys(exported);
-  for (var i = 0; i < exportedLabels.length; i++) {
-    _addUsage(exportedLabels[i]);
-  }
-
   var _getUnusedOption = function(unused_opt) {
     if (unused_opt === undefined) {
       unused_opt = state.option.unused;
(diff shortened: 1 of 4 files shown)


lukeapage:
I would add a new test run with latedef option to the exported test.

lukeapage:
I considered this approach but I wasn't sure the exported directive always means global scope.. I've been thinking though and I can't think of an example of non global leader scope so maybe that's ok.

lukeapage:
does this work if the directive is after the declaration? did it work in 2.8?

nicolo-ribaudo:
> does this work if the directive is after the declaration? did it work in 2.8?

Yes - yes

lukeapage:
ah yes, but only if the directive is in the global scope. why not just check the global scope rather than only doing it if current is global? or..  wdyt?

lukeapage:
because that wouldn't work with a var in an if block...  ok.

nicolo-ribaudo:
> because that wouldn't work with a var in an if block... ok.

It works:
```js
// jshint unused: strict, esversion: 6
if (true) {
  /*exported a, b*/
  let a; // `a` is defined but never used - which is ok, because `a` *isn't* really exported
  var b;
}
```

lukeapage:
yes sorry for the brief comments, away from computer again.. thats what i
meant,  I was answering my own question.

lukeapage:
one small not very important regression remaining is code like this...

```
// jshint unused: true

function foo() {
}

if (true) {
   /* exported foo */
}
```

which doesn't fail at the moment because peeking means the lexer runs the directive before entering the block - that may and probably will change. However this does fail at the moment...

```
// jshint unused: true

function foo() {
}

if (true) {
    for(var i = 0; i < 1; i++) {
       // dont peek
    }
   /* exported foo */
}
```
I'm not sure if there are any realistic examples of this.

To fix it, instead of...

```
 } else if (_current["(type)"] === "global" && _.has(_current["(labels)"], labelName)) {
```

you would have to use _getLabel but tell it to only look in the global scope.. I'm not sure it is worth it.

lukeapage:
`_addUsage` can be inlined since it is now only used in one place.

lukeapage:
> > I would add a new test run with latedef option to the exported test.

> Added

Great, that covers that. For extra coverage though, how about..

here - https://github.com/nicolo-ribaudo/jshint/commit/a58a91b8ca7e18e19822a721c32aecdb8989d8b8#diff-59929b5758a2716d37551959b0155cbaR1126

could you also add

```
run.test(src, {latedef: true, unused: true });
```

nicolo-ribaudo:
Done

lukeapage:
The trouble with that fix is that I don't think this works

```
// jshint unused: true
if (true) {
  var a;
}
/* exported a */
```

again I'm not sure how important these corner cases are...

nicolo-ribaudo:
> The trouble with that fix is that I don't think this works

Fixed

lukeapage:
Thanks, LGTM

lukeapage:
@jugglinmike I don't mind merging this but I'll keep it around till you comment in case you have any other observations. If you want to merge it, go ahead of course!

jugglinmike:
@lukeapage I've been watching this patch come together, and it looks good to me. I'll let you do the honors since the scope manager is your baby (figuratively).

lukeapage:
squashed and committed - https://github.com/jshint/jshint/commit/c630994e9470bd82d704aba05a0ece07d1582147

PR #2677, opened 2015-09-11
[[FIX]] Allow comments and new lines after /* falls through */

nicolo-ribaudo:


jugglinmike:
This is nice because it moves appropriate responsibilities into the lexer. I was tempted to suggest encapsulating the `caseFallsThrough` property assignment in `doOption`, but "falls through" isn't exactly an option. Maybe one day a `doDirective` function will be appropriate, but for now, it's probably a case of YAGNI.

Anyway, merged to `master` at commit 3b1c92581bfd65a3ba2ac44643a412a613dd8671. Thank you!

Issue #2668, opened 2015-09-07
lack of "use strict" directive doesn't always warn


lukeapage:
This code:
```
// jshint strict: true, globalstrict: true
a = 1;
```
(or its 2.9.x + equivalent)
```
// jshint strict: global
a = 1;
```
both do not warn about missing a `"use strict"` directive.

I'm guessing only very short code files suffer this problem (or this would be a big problem) but it requires investigation.

Found during investigation of #2663

nicolo-ribaudo:
Another test case:
```js
// jshint strict: global
a + b;
```

nicolo-ribaudo:
JSHint doesn't warn if the statement is a function or a punctuator (`a = 2` is `=`, so JSHint doesn't warn, while `var a = 2` is `var`, so JSHint warns)

Issue #1075, opened 2013-05-09
JSHint with asi: true still thinks that "use strict" should use a semicolon


mcandre:
I've set `"asi": true` in my `.jshintrc`, but JSHint still complains:

```
$ jshint .
index.js: line 1, col 1, Missing semicolon.

$ less index.js
'use strict'
...
```

dav-:
This is still an issue for me

caitp:
@dav- you will still need a semicolon depending on what comes after the "use strict" directive, regardless of your ASI settings. The reason for this is, for instance the following:

```js
"use strict"
(function() {
})()
```

This ends up being:

{String} {Call}{Call} --- in other words, the string is not processed as a directive, it's processed as a plain old string expression, which is the LHS of a call expression, and the result of that first call is itself called.

So this confuses people, but yeah I think we're right to warn about things like that. There are other productions where you'll get similar results, too. However, if you think the statement following should be allowed, please open a new issue with a reproduction

PR #2678, opened 2015-09-11
[[FIX]] Always warn about missing "use strict" directive

nicolo-ribaudo:
Fixes gh-2668

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -6935,6 +6935,7 @@ exports.testStrictDirectiveASI = function (test) {
     .test("'use strict'\n(function fn() {})();", options);
 
   TestRun(test, 5)
+    .addError(2, "Missing \"use strict\" statement.")
     .test("'use strict'\n[0] = '6';", options);
 
   TestRun(test, 6)
@@ -6944,6 +6945,7 @@ exports.testStrictDirectiveASI = function (test) {
     .test("'use strict',function fn() {}\nfn();", options);
 
   TestRun(test, 7)
+    .addError(1, "Missing \"use strict\" statement.")
     .test("'use strict'.split(' ');", options);
 
   TestRun(test, 8)
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1667,7 +1667,9 @@ var JSHINT = (function() {
 
     r = expression(0, true);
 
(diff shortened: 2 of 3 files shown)


tests/unit/parser.js
```diff
@@ -6944,6 +6945,7 @@ exports.testStrictDirectiveASI = function (test) {
     .test("'use strict',function fn() {}\nfn();", options);
 
   TestRun(test, 7)
+    .addError(1, "Missing \"use strict\" statement.")
```
jugglinmike: This one is my bad; as written, that test was clearly invalid. Thanks for correcting it.

tests/unit/parser.js
```diff
@@ -6935,6 +6935,7 @@ exports.testStrictDirectiveASI = function (test) {
     .test("'use strict'\n(function fn() {})();", options);
 
   TestRun(test, 5)
+    .addError(2, "Missing \"use strict\" statement.")
```
jugglinmike: This change looks right to me, but since it's a change to an existing test, I'd like a third opinion. @caitp what do you think?

lukeapage: https://github.com/jshint/jshint/issues/1075

lukeapage: and related: https://github.com/twbs/bootstrap/issues/1905

caitp: What exactly is the behavior of this patch? There's a use strict directive in the source, so i don't understand why it's being warned about?

caitp: Oh, the test source isn't visible in the diff, my bad

caitp: no, I still don't understand the warning that is being emit here. Why is this warning being expected?

nicolo-ribaudo: Because `;` isn't automatically inserted if the next token is `[`, so that code is like
```js
'use strict'[0] = '6';
```

jugglinmike: I talked this over with Cait on IRC, and we agreed that this is valid. This may be a good candidate for a more specific warning, but we can investigate that in a follow-up patch.

jugglinmike:
Rebased and merged to `master` at commit e85c2a139b4314c13c563d8b359912198ab48f7b. Thanks!

PR #2684, opened 2015-09-18
Add the meteor platform

raix:


jugglinmike:
Unfortunately, accepting library-specific globals is a slippery slope towards supporting all sorts of lesser-used libraries. Instead of trying to define which libraries are "popular enough" to justify inclusion, we now reserve environment settings for true environments.

Thanks anyway!

PR #2686, opened 2015-09-18
[[FIX]] Allow let and const to be in a block outside of a block

### LINKED ISSUE #2685: let declaration not directrly within block
```js
while(true)
  if(true)
  {
    let a = 1;  // let declaration not directly within block
  }
```
also
```js
if(true)
  //;
else while(true)
{
  let a = 1; // let declaration not directly within block
}
```

with {}:
```js
while(true)
{
  if(true)
  {
    let a = 1;
  }
}
// without errors
```

nicolo-ribaudo:
Fixes #2685

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1802,6 +1802,7 @@ var JSHINT = (function() {
 
       // create a new block scope
       state.funct["(scope)"].stack();
+      state.funct["(noblockscopedvar)"] = false;
 
       line = state.tokens.curr.line;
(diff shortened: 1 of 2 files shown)


jugglinmike:
This looks good to me. @lukeapage mind giving it a once-over?

jugglinmike:
Merged to `master` at 84a9145bb19a925652b77169a08d632e571cec28. Thanks!

PR #2687, opened 2015-09-19
Fix ES5 default value

niedzielski:
ES5 option is now set per default.


jugglinmike:
Merged to `master` at commit 5c96bf8fb89b747aceb8d10626153ad2fc4e5f36. Thanks!

PR #2688, opened 2015-09-21
[[FIX]] Enforce TDZ within initializer of lexical declaration

### LINKED ISSUE #2637: TDZ is not enforced within initializer of lexical declaration
The following code should produce [error E056](https://github.com/jshint/jshint/blob/18f8775a05c10fc8c157ef7b79c16d36ba0bbc7b/src/messages.js#L74) but currently does not:

```js
// jshint esversion: 6
let a = a;
const b = b;
```

See also: gh-2632

nicolo-ribaudo:
Fixes #2637

### DIFF
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -1989,3 +1989,42 @@ exports.duplicateProto = function (test) {
 
   test.done();
 };
+
+exports["TDZ within initializer of lexical declarations"] = function(test) {
+  var code = [
+    "let a = a;",
+    "const b = b;",
+    "let c = () => c;",
+    "const d = () => d;",
+    // line 5
+    "let e = {",
+    "  x: e,",
+    "  y: () => e",
+    "};",
+    "const f = {",
+    "  x: f,",
+    "  y: () => f",
+    "};",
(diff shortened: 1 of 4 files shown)


tests/unit/core.js
```diff
@@ -1989,3 +1989,42 @@ exports.duplicateProto = function (test) {
...
+    "};",
+    "const f = {",
+    "  x: f,",
+    "  y: () => f",
+    "};",
+    // line 13
+    "let g, h = g;",
+    "const i = 0, j = i;",
+    "let [ k, l ] = l;",
+    "const [ m, n ] = n;",
```
nicolo-ribaudo: Is it correct to throw on lines 15 and 16? (Babel doesn't)

nicolo-ribaudo:
Just added other TDZ checks

jugglinmike:
Thanks for the patch, @nicolo-ribaudo! I have a different approach in mind--could you tell me what you think of my changes in gh-2733?

nicolo-ribaudo:
Your approach looks better than mine, I'm closing this pr.

PR #2691, opened 2015-09-22
[[FIX]] Add `File` to browser global variables

### LINKED ISSUE #2690: Is it intentional that "File" is left out of browser default global variables?
In [default variables for browser](https://github.com/jshint/jshint/blob/7f5806f86cbd6f30424fa6e0fd9ae659376925fc/src/vars.js#L58-L381) I don't see `File`. Although `File` constructor is not used directly it can be used for type checking. For example in my drag and drop code I wrote:

```js
if ($scope.draggedFiles instanceof File) {
```

`File` spec: http://www.w3.org/TR/FileAPI/#dfn-file

mohsen1:
File was left out of browser global variables

Fixes #2690

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -84,6 +84,8 @@ exports.browser = {
   Event                : false,
   event                : false,
   fetch                : false,
+  File                 : false,
+  FileList             : false,
   FileReader           : false,
   FormData             : false,
   focus                : false,


lukeapage:
LGTM, most of everything else in that spec is in the list.

Would you add FileList too ?

mohsen1:
Added FileList

jugglinmike:
Merged to `master` at commit 7f2a729. Thanks, Mohsen!

PR #2692, opened 2015-09-22
[[CHORE]] Update dependencies and change jscs accordingly

BridgeAR:
I checked all tests / build steps etc. and also checked the changelogs for all major / minor versions without any indication that this would break anything.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -5,7 +5,10 @@
       "allowComments": true,
       "allowRegex": true
     },
-    "validateIndentation": 2,
+    "validateIndentation": {
+      "allExcept": ["comments"],
+      "value": 2
+    },
 
     "disallowMultipleLineStrings": true,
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch! I appreciate that you did the research to update the project's dependencies, but please understand that this particular task is really a burden of the maintainers. It may take some time before any of us are able to conduct a thorough review for this changeset.

BridgeAR:
Shall I split this into smaller patches? E.g. the dev dependencies should not be an issue? And I could also add the changelog for the other three dependencies if that helps or check the commits for things that changed, if you trust me on that.

jugglinmike:
I don't think that will be necessary. Thanks, though!

jugglinmike:
I'm closing this because the dependency information no longer reflects the latest versions of the modules we're using. When we update dependencies, we'll need to use the latest versions.

PR #2701, opened 2015-09-29
[[FIX]] Don't throw W080 when the initializer starts with `undefined`

nicolo-ribaudo:
> Example:
   ```js
var a = undefined === 0; // a is 'false', not 'undefined'
```

> Fixes gh-2699

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3441,6 +3441,8 @@ var JSHINT = (function() {
         }
       }
     };
+
+    var id, value;
     if (checkPunctuator(firstToken, "[")) {
       if (!openingParsed) {
         advance("[");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jugglinmike** [comment]
Thanks for the patch! It's missing a few test cases, though. Can you add assertions for destructuring patterns, eg:

```js
var [ a = undefined === 0 ] = [];
[ a = undefined === 0 ] = [];
var { b: b = undefined === 0 } = {};
({ b: b = undefined === 0 } = {});
```

nicolo-ribaudo:
> Can you add assertions for destructuring patterns

Done

jugglinmike:
Squashed and merged to `master` at commit 0d879196bab1c733f1299d2dbc470362b64c9bc6. Thanks!

PR #2703, opened 2015-10-04
[[FIX]] Disallow incompatible option values

jugglinmike:
This was requested by @rwaldron during the discussion of gh-2663.

Commit message:

> The recently-implemented "global" value for the "strict" option is
> incompatible with the "globalstrict" value, so their combined use should
> be disallowed.
>
> Introduce a generic error message for incompatible option values, and
> utilize it to restrict the combined usage of "strict: global" and
> "globalstrict".

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -210,6 +210,10 @@ var JSHINT = (function() {
       combine(predefined, vars.ecmaIdentifiers[6]);
     }
 
+    if (state.option.strict === "global" && "globalstrict" in state.option) {
+      error("E059", state.tokens.next, "strict", "globalstrict");
+    }
+
     if (state.option.module) {
       if (state.option.strict === true) {
         state.option.strict = "global";
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -210,6 +210,10 @@ var JSHINT = (function() {
       combine(predefined, vars.ecmaIdentifiers[6]);
     }
 
+    /**
+     * Use `in` to check for the presence of any explicitly-specified value for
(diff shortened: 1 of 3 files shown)


src/jshint.js
```diff
@@ -210,6 +210,10 @@ var JSHINT = (function() {
       combine(predefined, vars.ecmaIdentifiers[6]);
     }
 
+    if (state.option.strict === "global" && "globalstrict" in state.option) {
```
rwaldron: I'd add a note explaining _why_ you've used `prop in obj`, ie. that both `false` and `true` must be considered.

src/messages.js
```diff
@@ -73,7 +73,8 @@ var errors = {
   E055: "The '{a}' option cannot be set after any executable code.",
   E056: "'{a}' was used before it was declared, which is illegal for '{b}' variables.",
   E057: "Invalid meta property: '{a}.{b}'.",
-  E058: "Missing semicolon."
+  E058: "Missing semicolon.",
+  E059: "Incompatible values for the '{a}' and '{b}' linting options."
```
rwaldron: Nice

rwaldron:
Just this: https://github.com/jshint/jshint/pull/2703/files#r41664499 then it LGTM

jugglinmike:
@rwaldron Thanks for the review. I just pushed up a commit to address your feedback.

jugglinmike:
Merged to `master` at commit 72ba5ad20bfe51cf6fb2f03d4f3036d65d99ccd4.

PR #2705, opened 2015-10-05
[[FIX]] Don't warn that `let is ES6 or moz` about let blocks and let expressions

nicolo-ribaudo:
```js
let (a = 1) {
  console.log(a + let (a = 2) a;)
}
```

JSHint shouldn't warn that `'let' is available in ES6 (use 'esversion: 6') or Mozilla JS extensions (use moz).`.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -895,7 +895,7 @@ var JSHINT = (function() {
       state.funct["(scope)"].stack();
       advance("let");
       advance("(");
-      state.tokens.prev.fud();
+      state.tokens.prev.fud({ letExpr: true });
       advance(")");
     }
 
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3528,19 +3528,20 @@ var JSHINT = (function() {
...
+    var isLetExpr = context && context.letExpr;
+    var isLetBlock = isLet && checkPunctuator(state.tokens.next, "(");
+    var tokens, lone, value;
 
-    if (!state.inES6()) {
+    if (!isLetExpr && !isLetBlock && !state.inES6()) {
       warning("W104", state.tokens.curr, type, "6");
     }
 
-    if (isLet && state.tokens.next.value === "(") {
```
caitp: I'd rather we just got rid of the parsing code for handling this, rather than adding a new setting to allow it. The reason is, we really want to discourage extensions and other users from making use of this at all. There isn't really a good reason to now, even for ESR releases of Firefox

PR #2711, opened 2015-10-08
[[FIX]] Arrow function precedence

nicolo-ribaudo:
This is a little hacky because `=>` has different precedence on left and right.

I created the `Wrap the arrow function in parens.` error because throwing `Expected '(' abd instead saw '{b}'` would be confusing when `ArrowParameters` is `CoverParenthesizedExpressionAndArrowParameterList` (`Expected '(' and instead saw '('` :confused: )


tests/unit/options.js
```diff
@@ -2611,25 +2613,44 @@ singleGroups.arrowFunctions = function (test) {
...
+  run.test(loneArgs, options);
+
+  code = [
+    // unnecessary
+    "var a = (x) => (3 || {});",
+
+    // necessary
+    "var b = ((x) => {}) || 3;",
+    "var c = 3 || ((x) => 3);",
+    "var d = (x) => ({} || 3);",
```
nicolo-ribaudo: Here it isn't needed to wrap the `||` operator, but JSHint doesn't warn because without parenthesis `{}` wouldn't be an object literal anymore. (It could be written as `(x) => ({}) || 3`)

Issue #2714, opened 2015-10-12
Semicolon ignore doesn't work with strings as the only statement


wthit56:
(All examples apply for single-quoted strings and double-quoted strings.)
Semicolon warning shows for: (`//` marks the line that the warning appears on)
```js
function() { "code" } //
function() { "code"; } //
function() {
"code" }  //
function() { "code"; "code" } //
function() { "code", "code" } //
```

But not for:
```js
function() { 1 }
function() { true }
function() { window }
function() { "code"
}
function () { 1; "code" }
function() { 1
"code" }
function() { string = "code" }
```

**.jshintrc**:
```json
{
	"expr": true,
	"shadow": true,
	"asi": true,
	"evil": true,
	"laxbreak": true
}
```

PR #2716, opened 2015-10-12
[[FIX]] Correctly recognize asi after directives

nicolo-ribaudo:
Fixes gh-2714

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1732,32 +1732,19 @@ var JSHINT = (function() {
 
   /*
    * read all directives
-   * recognizes a simple form of asi, but always
-   * warns, if it is used
    */
   function directives() {
     var i, p, pn;
 
     while (state.tokens.next.id === "(string)") {
-      p = peek(0);
-      if (p.id === "(endline)") {
-        i = 1;
-        do {
-          pn = peek(i++);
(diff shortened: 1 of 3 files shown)


src/jshint.js
```diff
@@ -1732,32 +1732,20 @@ var JSHINT = (function() {
...
-        }
-      } else if (p.id === "." || p.id === "[") {
+      pn = peek();
+      i = 0;
+      do {
+        p = pn;
+        pn = peek(++i);
+      } while (p.id === "(endline)");
+      if (p.lbp > 0 &&
+          (startLine(pn) !== p.line || !checkPunctuators(p, ["++", "--"]))) {
```
nicolo-ribaudo: Are `++` and `--` the only tokens which may be not related to the previous one if they are on a different line?

jugglinmike:
Squashed and merged to `master` at commit 039ee2e7f26b5d4b6f503d21b66664387185d940. 

The extensions to `isEndOfExpr` implemented in gh-2977 make this patch's logic concerning the prefix increment/decrement operators a little duplicative. I'm experimenting with the best way to re-use the new code; I'll let you know when/if I come up with something worth looking at.

In any event, thanks for the fix!

PR #2721, opened 2015-10-18
[[DOCS]] Mark the 'curly' option as deprecated

ricordisamoa:
Commit 4e8921666a5fd54c5f58ebc66f3c61d6c60f7d85 marked many options
like 'camelcase', 'indent' and 'quotmark' as deprecated in favor of
the respective JSCS rules.

JSCS' requireCurlyBraces is the equivalent of JSHint's curly, as stated
at http://jscs.info/rule/requireCurlyBraces, but more customizable.

In order not to split efforts between projects, JSHint should not
enforce code style.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -47,6 +47,11 @@ exports.bool = {
      *     while (day)
      *       shuffle();
      *       sleep();
+     *
+     * @deprecated JSHint is limiting its scope to issues of code correctness.
+     *             If you would like to enforce rules relating to code style,
+     *             check out [the JSCS
+     *             project](https://github.com/jscs-dev/node-jscs).
      */
     curly       : true,
 


jugglinmike:
Thanks for the patch! I don't think deprecation is appropriate here, though, since the use of block delimeters for single-statement blocks is more than a matter of coding style. [Last year's iOS TLS bug](https://www.imperialviolet.org/2014/02/22/applebug.html) is a good example of how this consideration can actually help developers write safer code.

I think the documentation does a pretty good job describing this, but I'd be open to suggestions for making this point more clear.

ricordisamoa:
Are you suggesting that JSCS should drop its option? Is redundancy any good here?

jugglinmike:
I think that the decision to remove functionality from a library is best left to its maintainers. If the folks at JSCS want to keep `requireCurlyBraces`, then I think it's a great idea. The additional control there might be useful for users who do not recognize this as a code safety issue--if you consider this strictly a stylistic consideration, you may have different preferences for different structures.

PR #2722, opened 2015-10-19
[[DOCS]] Update notation about `esnext` option

hakatashi:
ES6 is already finalized on June 17.
Ref: http://www.ecma-international.org/news/Publication%20of%20ECMA-262%206th%20edition.htm


caitp:
"finalized" isn't the word I'd used, more like "ratified". But LGTM

jugglinmike:
Thank you, @hakatashi

PR #2727, opened 2015-10-21
[[FEAT]] Allow multipler reporters to run

### LINKED ISSUE #1702: Allow multiple reporters for a single run
When running JSHint in Jenkins, we want the regular output for the build log (emailed to developers for quick fixes) and the `checkstyle` reporter for the Jenkins Checkstyle plugin to process.

Is this possible? If not, it would be very handy so we don't have to run JSHint twice.

jsr6720:
Added new helper fuction getReporter that removes logic from interpret.
Now can call jshint example.js --reporter=unix,checkstyle. Fully backwards
compatible.

Fixes #1702

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -204,7 +204,13 @@ function loadIgnores(params) {
     return [];
   }
 
-  var lines = (file ? shjs.cat(file) : "").split("\n");
+  var name = (file ? shjs.cat(file) : "");
+
+  if (!name) {
+    return [];
+  }
+
+  var lines = name.split("\n");
   var exclude = params.exclude || "";
   lines.unshift.apply(lines, exclude.split(","));
 
@@ -632,7 +638,12 @@ var exports = {
         mergeCLIPrereq(config);
 
         lint(extract(code, opts.extract), results, config, data, filename);
-        (opts.reporter || defReporter)(results, data, { verbose: opts.verbose });
+
+        // invoke each reporter
+        for (var index in opts.reporters) {
+          (exports.getReporter(opts.reporters[index]))(results, data, { verbose: opts.verbose });
+        }
+
         cb(results.length === 0);
       });
 
@@ -668,7 +679,17 @@ var exports = {
       }
     });
 
-    (opts.reporter || defReporter)(results, data, { verbose: opts.verbose });
+    if (opts.hasOwnProperty('reporter')) {
+      // passed in manually 'old' way without going through entry point of script (interpret)
+      (opts.reporter)(results, data, { verbose: opts.verbose });
+    }
+    else {
+      // invoke each reporter as set up on
+      for (var index in opts.reporters) {
+        (exports.getReporter(opts.reporters[index]))(results, data, { verbose: opts.verbose });
+      }
+    }
+
     return results.length === 0;
   },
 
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -681,6 +702,54 @@ var exports = {
...
+   *
+   * @param {string} name of reporter to fetch, pre-defined or custom.
+   */
+  getReporter: function(reporterName) {
+    var reporter = defReporter; // default
+
+    if (!reporterName) {
+      return reporter; // return default
+    }
+    switch (true) {
```
nicolo-ribaudo: Who not `switch (reporterName) {`?

src/cli.js
```diff
@@ -632,7 +638,12 @@ var exports = {
         mergeCLIPrereq(config);
 
         lint(extract(code, opts.extract), results, config, data, filename);
-        (opts.reporter || defReporter)(results, data, { verbose: opts.verbose });
+
+        // invoke each reporter
+        for (var index in opts.reporters) {
+          (exports.getReporter(opts.reporters[index]))(results, data, { verbose: opts.verbose });
```
nicolo-ribaudo: Can you remove the parentheses around `exports.getReporter(...)`? (Just for style consistency)

src/cli.js
jsr6720: Want me to switch it to switch(reporterName)? You're right the original switch was (true) because it had multiple variable comparison.

jsr6720:
Any more feedback on this pull request? We're hoping to see this merged and use it @Paychex
Thanks.

src/cli.js
```diff
@@ -204,7 +204,13 @@ function loadIgnores(params) {
     return [];
   }
 
-  var lines = (file ? shjs.cat(file) : "").split("\n");
+  var name = (file ? shjs.cat(file) : "");
```
nicolo-ribaudo: Unnecessary parentheses

src/cli.js
```diff
@@ -668,7 +679,17 @@ var exports = {
       }
     });
 
-    (opts.reporter || defReporter)(results, data, { verbose: opts.verbose });
+    if (opts.hasOwnProperty('reporter')) {
+      // passed in manually 'old' way without going through entry point of script (interpret)
+      (opts.reporter)(results, data, { verbose: opts.verbose });
```
nicolo-ribaudo: Unnecessary parentheses

src/cli.js
```diff
@@ -668,7 +679,17 @@ var exports = {
...
+    }
+    else {
+      // invoke each reporter as set up on
+      for (var index in opts.reporters) {
+        exports.getReporter(opts.reporters[index])(results, data, { verbose: opts.verbose });
```
nicolo-ribaudo: Can you use `for` loop instead of `for...in`? (Because `opts.reporters` is an array)

src/cli.js
```diff
@@ -681,6 +702,55 @@ var exports = {
...
+        reporter = "./reporters/non_error.js";
+        break;
+
+      // Custom reporter. not to be confused with 'default' if undefined above
+      default:
+        reporter = path.resolve(process.cwd(), reporterName);
+        break;
+    }
+
+    if (reporter) {
```
nicolo-ribaudo: `reporter` is never `false`/`null`/`""`, so this condition is not necessary

nicolo-ribaudo:
Is it possible to run the default reporter along with another reporter?

src/cli.js
```diff
@@ -668,7 +679,17 @@ var exports = {
       }
     });
 
-    (opts.reporter || defReporter)(results, data, { verbose: opts.verbose });
+    if (opts.hasOwnProperty('reporter')) {
+      // passed in manually 'old' way without going through entry point of script (interpret)
+      (opts.reporter)(results, data, { verbose: opts.verbose });
+    }
+    else {
```
nicolo-ribaudo: (style) `else` should be on the same line of `}`

src/cli.js
jsr6720: Agreed. The code handles false/null/"". This 'default:' was an interpretation of the old switch statement that evaluated (true) and then had this block for detecting custom reporters.

The 'default' in this new switch statement is for handling custom reporters.

Am I understanding the requirements correctly?

##Original 'default' line
```javascript
// Custom reporter
    case options.reporter !== undefined:
      options.reporter = path.resolve(process.cwd(), options.reporter);
```

jsr6720:
@nicolo-ribaudo Is it possible to run the default reporter along with another reporter?

Patched this in. Now you can run the default reporter using keyword 'default'.

```
$ bin/jshint examples/reporter.js --reporter=default,checkstyle
```

jsr6720:
@nicolo-ribaudo Thank you for the feedback, you've got a great eye for detail. I'm looking forward to further conversation on this PR.

nicolo-ribaudo:
> Now you can run the default reporter using keyword 'default'.

This can be a breaking change... (cc @jugglinmike)
If you have a reporter in a file named `default.js`, running `jshint --reporter=default file.js` that file won't be loaded anymore.

src/cli.js
nicolo-ribaudo: I meant this:
```diff
@@ -737,16 +737,14 @@ var exports = {
       default:
         reporter = path.resolve(process.cwd(), reporterName);
         break;
     }

-    if (reporter) {
-      reporter = loadReporter(reporter); // copy over path with require obj
-      if (reporter === null) {
-        cli.error("Can't load reporter file: " + reporterName);
-        exports.exit(1);
-      }
+    reporter = loadReporter(reporter); // copy over path with require obj
+    if (reporter === null) {
+      cli.error("Can't load reporter file: " + reporterName);
+      exports.exit(1);
     }

     return reporter;
   },
   /**
```

jsr6720:
@nicolo-ribaudo @jugglinmike 
reporter expects a .js file extension to trigger finding a custom reporter. This is the same functionality as before the change. The switch statement is only for reporter 'options' that are bundled with jshint.

jsr6720:
@nicolo-ribaudo @jugglinmike 
Need anything else from me on this PR #2727

jsr6720:
@nicolo-ribaudo @jugglinmike 
Need anything else from me on this PR #2727

jugglinmike:
Thanks, but it looks like there is a more powerful solution available that doesn't require any changes to JSHint--please see gh-1702 for details!

PR #2728, opened 2015-10-21
Update .travis.yml

paladox:
Allow testing on travis ci new server which are faster.

PR #2732, opened 2015-10-25
Update .travis.yml

paladox:
This makes test run faster and on the new travis ci server.


nicolo-ribaudo:
@paladox  This is exactly the same of #2728, can you close this pr?

PR #2733, opened 2015-10-25
Enforce TDZ within initializer of lexical declaration (refactored)

### LINKED ISSUE #2637: TDZ is not enforced within initializer of lexical declaration

jugglinmike:
This patch includes work by @nicolo-ribaudo (originally submitted in gh-2701) and a refactoring of my own design.

I initially expected that this approach would reduce some of the noise
associated with explicit maintence of the standalone "definition" objects (e.g.
the calls to `state.funct["(scope)"].definition.reset`). While this turned out
to be true, I found that it *also* introduced a different kind of noise--see
the extra calls to `state.funct["(scope)"].initialize`.

I think the latter is preferable because it makes the source code more closely
match the spec language. Readers are more likely to know what it means to
"initialize a label" than what it means to "clear a definition."

Resolves #2637

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3576,7 +3576,9 @@ var JSHINT = (function() {
         }
       }
 
-      state.funct["(scope)"].definition.reset();
+      if (!prefix) {
+        state.funct["(scope)"].definition.reset();
+      }
 
       statement.first = statement.first.concat(names);
 
@@ -4275,6 +4277,10 @@ var JSHINT = (function() {
       //     for ( LeftHandSideExpression of AssignmentExpression ) Statement
       expression(bindingPower);
       advance(")", t);
+
+      if (letscope) {
+        state.funct["(scope)"].definition.reset();
+      }
 
       if (nextop.value === "in" && state.option.forin) {
         state.forinifcheckneeded = true;
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3553,6 +3553,7 @@ var JSHINT = (function() {
             if (lone && inexport) {
               state.funct["(scope)"].setExported(t.token.value, t.token);
             }
+            state.funct["(scope)"].definition.add(t.id, type);
           }
         }
       }
@@ -3574,6 +3575,8 @@ var JSHINT = (function() {
           destructuringPatternMatch(names, value);
         }
       }
+
+      state.funct["(scope)"].definition.reset();
 
       statement.first = statement.first.concat(names);
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3549,11 +3549,6 @@ var JSHINT = (function() {
               type: type,
               token: t.token });
             names.push(t.token);
-
-            if (lone && inexport) {
-              state.funct["(scope)"].setExported(t.token.value, t.token);
-            }
-            state.funct["(scope)"].definition.add(t.id, type);
           }
         }
       }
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
> I think the latter is preferable because it makes the source code more closely
match the spec language.

:thumbsup:

> I initially expected that this approach would reduce some of the noise associated with explicit maintence of the standalone "definition" objects (e.g. the calls to `state.funct["(scope)"].definition.reset`). While this turned out to be true, I found that it also introduced a different kind of noise--see the extra calls to `state.funct["(scope)"].initialize`.

You could make the `addlabel` function accept an `initialized` parameter (whose default value is `true`) so that you don't need the extra calls to `state.funct["(scope)"].initialize`.
e.g.
```diff
diff --git a/src/jshint.js b/src/jshint.js
index 0758967..4bb41f0 100644
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3570,4 +3570,5 @@ var JSHINT = (function() {
             state.funct["(scope)"].addlabel(t.id, {
               type: type,
+              initialized: false,
               token: t.token });
             names.push(t.token);
@@ -3743,4 +3744,5 @@ var JSHINT = (function() {
       state.funct["(scope)"].addlabel(this.name, {
         type: "class",
+        initialized: false,
         token: state.tokens.curr });
diff --git a/src/scope-manager.js b/src/scope-manager.js
index 11e141a..767abf2 100644
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -649,5 +649,5 @@ var scopeManager = function(state, predefined, exported, declared) {
         }

-        scopeManagerInst.block.add(labelName, type, token, !isexported);
+        scopeManagerInst.block.add(labelName, type, token, !isexported, opts.initialized);

       } else {
@@ -813,9 +813,9 @@ var scopeManager = function(state, predefined, exported, declared) {
        * Adds a new variable
        */
-      add: function(labelName, type, tok, unused) {
+      add: function(labelName, type, tok, unused, initialized) {
         _current["(labels)"][labelName] = {
           "(type)" : type,
           "(token)": tok,
-          "(initialized)": false,
+          "(initialized)": initialized !== false,
           "(blockscoped)": true,
           "(unused)": unused };
```

jugglinmike:
I had some similar code when I first drafted this patch, but I held off on it because of a gut reaction to the boolean trap for the `add` method. That could be resolved by refactoring to use an `options` object with well-named attributes, but I'm also (overly?) sensitive to object allocations costs in this codepaths.

It's been over two weeks, though, and I've come around on the idea of simply adding that additional boolean parameter. It's an internal method, so the usability concern is kind of minimal.

That said, I decided to implement `initialized` as "false by default". It's true that there are more cases in the code base where we create initialized bindings than uninitialized bindings, but I still think this is preferable. "False by default" is easier to reason about because it aligns with JavaScript's natural treatment of values (`undefined` coerces to `false`), and it also makes "pre-initialized" bindings opt-in only. My thinking is that we don't want to accidentally creating bindings this way, and an explicit API makes that less likely.

jugglinmike:
@rwaldron I've rebased this patch (the original version which conflicts with `master` is available at https://github.com/jugglinmike/jshint/tree/tdz-initializer-presquash for posterity).

rwaldron:
Oh very cool :)

I will try to review this tomorrow. Feel free to nudge me if I haven't reported back by noon

jugglinmike:
@rwaldron Nudging, as requested.

rwaldron:
@jugglinmike is this ready to land when I'm done reviewing?

jugglinmike:
Yes, sir.

rwaldron:
I poked and prodded this one for a bit. Nice work!

jugglinmike:
Thanks, Rick!

jugglinmike:
And thanks, @nicolo-ribaudo!

PR #2734, opened 2015-10-25
Find .jshintignore relative to file. Fixes #2595

### LINKED ISSUE #2595: .jshintignore not respected if jshint is run from different directory
`.jshintignore` isn't always respected. Hopefully the following log shows the issue well. Notice I demonstrate that `.jshintrc` is working as expected.

```
[chris /tmp/a]$ jshint -v
jshint v2.8.0
[chris /tmp/a]$ ls -a1
.
..
.jshintignore
.jshintrc
foo.js
[chris /tmp/a]$ cat .jshintignore 
**
[chris /tmp/a]$ cat .jshintrc 
{
  "esnext": true
}
[chris /tmp/a]$ cat foo.js 
let x = 123;
console.log("no semicolon")
[chris /tmp/a]$ jshint foo.js 
[chris /tmp/a]$ cd ..
[chris /tmp]$ jshint a/foo.js 
a/foo.js: line 2, col 28, Missing semicolon.

1 error
```

AgentME:
This makes it so `.jshintignore` is now also loaded from the directories of processed files like `.jshintrc` is, instead of being loaded based on the current working directory. This fixes #2595.

There's also a commit improving findFile's caching. Previously, if you ran `findFile("e.js", "/a/b/c/d")`, and only "/a/e.js" existed, then the only cache entry added would be for "/a/e.js", which would only allow calls to `findFile("e.js", "/a")` to be answered by the cache. Future calls to `findFile("e.js", "/a/b/c/d")` would still check for "/a/b/c/d/e.js", "/a/b/c/e.js", "/a/b/e.js" before finally being able to make use of the cached entry.

Now, every recursive call makes a cache entry, so future calls to `findFile("e.js", "/a/b/c/d")` can immediately result in "/a/e.js".

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -188,7 +188,9 @@ function findFile(name, cwd) {
     return null;
   }
 
-  return findFile(name, parent);
+  var result = findFile(name, parent);
+  findFileResults[filename] = result;
+  return result;
 }
 
 /**
--- a/src/cli.js
+++ b/src/cli.js
@@ -563,7 +563,8 @@ var exports = {
    *
    * @param {object} post-processed options from 'interpret':
    *                   args     - CLI arguments
-   *                   ignores  - A list of files/dirs to ignore (defaults to .jshintignores)
+   *                   ignores  - A list of files/dirs to ignore (defaults to
+   *                     the contents of .jshintignore)
    *                   extensions - A list of non-dot-js extensions to check
    */
   gather: function(opts) {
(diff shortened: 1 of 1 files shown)


AgentME:
Is there anything I can do to push this along? Is the bug that this fixes clear? Are more tests needed?

hybridview:
Just FYI. This fix resolved my issue where ignore list was not used when launching from VStudio 2015 as an external tool.

cobexer:
Why is this not merged yet? I absolutely expected this behavior!

Any ETA?

cobexer:
@jugglinmike could you push this along?

PR #2742, opened 2015-10-31
[[FIX]] Do not enable `newcap` within strict mode

jugglinmike:
@rwaldron You're gonna love this.

Since the first public release of JSLint (ca120a7), this codebase has
automatically enabled the `newcap` option for all strict mode code:

    function use_strict() {
        if (nexttoken.value === 'use strict') {
            advance();
            advance(';');
            strict_mode = true;
            option.newcap = true;
            option.undef = true;
            return true;
        } else {
            return false;
        }
    }

Commit 8de8247 unified the way JSHint detects and enforces strict mode
across contexts. Among other modifications, it included the following
change:

    - if (state.tokens.curr.value === "use strict") {
    + if (state.isStrict()) {
        if (!state.option["(explicitNewcap)"]) {
          state.option.newcap = true;
        }
        state.option.undef = true;
      }

This change introduced a regression: it enables the `newcap` option for
class bodies and ES2015 module code as well.

Because ES5's strict mode and the warnings enabled by the `newcap`
option are orthogonal (and because as a stylistic concern, the `newcap`
is deprecated) correct the regression by relaxing the behavior for all
cases--do not automatically enable `newcap` when entering strict mode,
for whatever reason.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -738,9 +738,6 @@ var JSHINT = (function() {
             state.option[key] = (val === "true");
           }
 
-          if (key === "newcap") {
-            state.option["(explicitNewcap)"] = true;
-          }
           return;
         }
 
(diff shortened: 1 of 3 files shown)


rwaldron:
My name is Rick Waldron and I approve this change. Nice work @jugglinmike !

PR #2749, opened 2015-11-05
[[FIX]] Don't throw '(NaN% scanned)'

nicolo-ribaudo:
A very low priority fix...

(note `0% scanned` instead of `NaN% scanned`)

```
nicol@DESKTOP-EDGLMME MINGW64 ~/Documents/jshint (master)
$ bin/jshint demo.js
demo.js: line 1, col 1, Unrecoverable syntax error. (NaN% scanned).

1 error

nicol@DESKTOP-EDGLMME MINGW64 ~/Documents/jshint (0%-scanned)
$ bin/jshint demo.js
demo.js: line 1, col 1, Unrecoverable syntax error. (0% scanned).

1 error
```

demo.js
```js
}
```

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -330,7 +330,7 @@ var JSHINT = (function() {
 
   // Produce an error warning.
   function quit(code, line, chr) {
-    var percentage = Math.floor((line / state.lines.length) * 100);
+    var percentage = Math.floor((line / state.lines.length) * 100) || 0;
     var message = messages.errors[code].desc;
 
     throw {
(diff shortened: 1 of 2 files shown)


jugglinmike:
This solution hides the underlying problem: that `line` is `undefined`. A more explicit would look like this:

```diff
-    var percentage = Math.floor((line / state.lines.length) * 100) || 0;
+    var percentage = Math.floor(((line || 0) / state.lines.length) * 100);
```

...though even that seems to be treating a symptom. Maybe the lexer should be assigning a `line` property to that token.

nicolo-ribaudo:
I set `(begin)`'s position to 0 so that it isn't needed to set a fallback value.
I also made the `quit` function accept a token (instead of the line and the character), so it have the same signature of `warning` and `error`

PR #2754, opened 2015-11-10
[[FEAT]] Support QUnit's global notOk

leobalter:
Adds support for notOk exposed to the global object since QUnit 1.18.0

Ref https://github.com/jquery/qunit/releases/tag/1.18.0


jugglinmike:
This patch is **not** okay.

jugglinmike:
Thanks, Leo!

Issue #2761, opened 2015-11-13
[2.9.1-rc1] Inline "-W020" doesn't work if there is "+W020" later


nicolo-ribaudo:
```js
/* global foo: false */

// jshint -W020
foo = 2;
// jshint +W020
```

```
$ bin/jshint demo.js --verbose
demo.js: line 4, col 1, Read only. (W020)

1 error
```

The same is for W021

PR #2763, opened 2015-11-13
[[FIX]] Allow ignoring W020 and W021

nicolo-ribaudo:
I couldn't just move the checks to the `reassign` function because it would have generated warnings when a variable is defined after its assignment.
e.g.
```js
/* global foo: false */
(function() {
  foo = 2;
  var foo;
})();
```

Fixes gh-2761


jugglinmike:
I have some reservations about this solution, but it's following an established
convention, and it fixes the problem (which is the priority for a release
candidate). Besides, a cleaner approach would require some refactoring, and
that's not really appropriate while we're working on a release candidate. I'll
let you know when/if I get around to that.

Anyway, thanks for the patch! Merged to `master` at commit 46db9238da535a2d82d7565deac887de6e602ad4.

PR #2770, opened 2015-11-18
Fix the check if a file is ignored

thorn0:
Without this fix JSHint fails if the three conditions are true: 
1. it's run on Windows,
2. the current path contains characters like `++`,
3. ignore patterns are specified via .jshintignore or command line.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -238,7 +238,7 @@ function isIgnored(fp, patterns) {
       return true;
     }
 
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Can you sign the [cla](https://www.clahub.com/agreements/jshint/jshint)?

jugglinmike:
Thanks for the patch! We're currently running our tests on Windows as well as Linux; can you add a test to help us avoid regressions?

thorn0:
Done.

Issue #2308, opened 2015-04-13
Rhino fails on jshint@2.7.0


segrey:
Rhino isn't working with jshint@2.7.0

```
$ java -jar ./js.jar -debug ./jshint-2.7.0.js
js: "jshint-2.7.0.js", line 1975: uncaught JavaScript runtime exception: TypeError: Cannot read property "prototype" from undefined
	at jshint-2.7.0.js:1975 (runInContext)
	at jshint-2.7.0.js:8270 (anonymous)
	at jshint-2.7.0.js:1525 (anonymous)
	at jshint-2.7.0.js:1515 (anonymous)
	at jshint-2.7.0.js:6 (s)
	at jshint-2.7.0.js:6 (anonymous)
	at jshint-2.7.0.js:12384 (anonymous)
	at jshint-2.7.0.js:6 (s)
	at jshint-2.7.0.js:17963 (anonymous)
	at jshint-2.7.0.js:4
```

```
$ java -version
java version "1.8.0_40"
Java(TM) SE Runtime Environment (build 1.8.0_40-b25)
Java HotSpot(TM) Server VM (build 25.40-b25, mixed mode)
```

Reproduced on [rhino-js-1_7R4](https://developer.mozilla.org/ru/docs/Rhino/Download_Rhino) and [rhino-js-1_7R5](https://github.com/mozilla/rhino/releases/tag/Rhino1_7R5_RELEASE): js.jar could be found inside rhino-js installation.

File jshint-2.7.0.js:  https://raw.githubusercontent.com/jshint/jshint/2.7.0/dist/jshint.js
File jshint-2.6.3.js:  https://raw.githubusercontent.com/jshint/jshint/2.6.3/dist/jshint.js

However rhino works fine for jshint@2.6.3 (empty output):
```
$java -jar rhino1_7R5/js.jar -debug ./jshint-2.6.3.js
$
```

rwaldron:
> File jshint-2.7.0.js: https://raw.githubusercontent.com/jshint/jshint/2.7.0/dist/jshint.js
> File jshint-2.6.3.js: https://raw.githubusercontent.com/jshint/jshint/2.6.3/dist/jshint.js

That file isn't for Rhino, this one is: https://github.com/jshint/jshint/blob/master/dist/jshint-rhino.js

segrey:
[...]

Issue #2767, opened 2015-11-17
2.9.1rc1 rhino build does not contain platforms/rhino.js


MaxiBul:
2.9.1rc1 rhino build (dist/jshint-rhino.js) does not contain platforms/rhino.js file. In result it doesn't work

Plus for work with Rhino, first line in dist file has to be:
///
var window = this;
///

not
//
var window = {};
//

JDougherty:
Indeed, rhino support has been broken for at least six months (as you noted in issue #2308), and as far back as [two years ago](https://github.com/jshint/jshint/issues/645) it was said:

> Rhino version is deprecated

MaxiBul:
If it can be fixed in 1 line, why not add this fix?
Support of Rhino allow to use jshint in any editor, which support execute external program as source validator, without installing any third party huge platforms like Node (portability). Let it be deprecated, but workable, if it can.

Issue #2330, opened 2015-04-22
New 'lodash' Dependency Breaks JavaScriptCore


bdkjones:
I use JavaScriptCore (the low-level C library that powers JavaScript in WebKit) to run JSHint. (**Note:** This is NOT the JSC CLI runtime that you guys deprecated support for a long time ago.)

When I attempted to run JSHint 2.7.0 in this environment, it failed. In the past, this was caused by a bad version of Browserify. (See #1963). This time, it appears to be caused by the recent replacement of `underscore` with `lodash`. 

This commit is the issue: https://github.com/jshint/jshint/commit/2fc1e2377ac590eb0455f0a2b0ffef76f3f0b0ef

### How I Diagnosed the Issue

The first thing I did was roll back all the dependencies in `package.json` to their values before this commit: https://github.com/jshint/jshint/commit/6f83a1a78a5f1edec3fc7ac4cb4fa116661c7d7e

I then ran `npm install` and `./bin/build`, but the built version of JSHint continued to fail.

At that point, I noticed that `underscore` had been replaced by `lodash`. I undid that change, ran `npm install` and built JSHint. The compiled `jshint.js` file worked flawlessly in JavaScriptCore. 

### lodash 3.6.x and 3.7.x Are Buggy

Next, I wanted to see if `lodash` fixed whatever bug causes this in their 3.7.x release. So I installed a fresh copy of JSHint from npm and then edited `package.json` to require `lodash 3.7.x` instead of `lodash 3.6.x`. That made no difference. The built version of jshint still failed.

### Reason for Dependency Change?

Was there a pressing need to use `lodash` instead of `underscore`? What was the reason for this switch?

jugglinmike:
JSHint doesn't officially support JavaScriptCore, but there's no reason why it shouldn't. If we extend the testing environment to run the project in that engine, then we'll be able to avoid breaking support (for a third time) in the future.

You'll have to bear with us on this: believe it or not, we're still working on testing JSHint in a browser (see gh-2128). JSC isn't nearly as familiar to me, so it may take some time to figure out the necessary extensions to the testing infrastructure. The good news is that once that's done, it *should* be easy to update our continuous integration process; I think we can just install the `rhino` package from apt on the TravisCI Ubuntu machine and use `rhino-jsc`.

Apologies for the late reply!

caitp:
we could be able to build a native ObjC app that links with the system JSC library on travis in order to test JSC support, that seems like a pretty fun side project actually

[...]

PR #2774, opened 2015-11-20
fix support Rhino (#2308 ?)

MaxiBul:


nicolo-ribaudo:
Can you sign the [Contributor's License Agreement](https://www.clahub.com/agreements/jshint/jshint)?

MaxiBul:
done.

jugglinmike:
Thanks for the patch, @MaxiBul! As @JDougherty pointed out in gh-2767, previous maintainers of this project have been lukewarm about Rhino support. I see no reason why we can't/shouldn't continue to support Rhino, but I'm reluctant to do so without tests that verify the fix. Otherwise, we risk disappointing users once again. As much as I think Rhino users would like the latest version, I think it would be *more* disruptive to publish a fixed version only to accidentally break it again in some later release.

I would like to work with you to extend our testing infrastructure to cleanly execute JSHint's unit tests in the Rhino environment. [I partially abstracted the tests in service of testing in browser environments](https://github.com/jshint/jshint/pull/2128), but we'll need to go further still. This work is closely related to [the open issue to re-implement support for the JavaScriptCore environment](https://github.com/jshint/jshint/issues/2330), and I suspect it will use some form of the code I've proposed there.

Obviously this is going to require much more work, though! The first question is: are you interested in contributing to this effort?

MaxiBul:
I can try, but currently don't have much time for this.
In any way, latest builds totally lost support of Rhino. Currently you can mark that you support it unofficial and keep it at least with possibility to run.

segrey:
@MaxiBul @jugglinmike Probably, it's not related directly to this PR, but what do you think about maintaining `jshint-lib.js` -  a single file containting all JSHint functionality, like `dist/jshint.js`. The difference is that it will be a JavaScript environment agnostic library, i.e it will work in browser, Rhino, JavaScriptCore, maybe even node (though it's of no use). Then `jshint-rhino.js` can be simply built as a concatenation of `jshint-lib.js` and [rhino.js wrapper](https://github.com/jshint/jshint/blob/master/src/platforms/rhino.js). The reason why I'm suggesting such an approach is because it allows to compile `jshint-lib.js` with rhino (in the same java process) and then communicate with JSHint via Rhino API.
Well, sad thing here is that I'm planning to move JSHint integration from Rhino to Node, that would allow to bring a better integration in IDE. But I can help here.

MaxiBul:
@segrey I'm beginner in jshint and did not work with Node. => I can't suggest here anything.
I just use jshint with EditPlus.
Primary tried to use jshint via wscript.exe - but jshint is too large for it.
With Rhino it work fine.
May be in future Microsoft will update wscript.exe to use Edge engine - I'll try it and drop Rhino.
But currently I hope you will not drop Rhino, while it can work with your library. Even without testing :)

segrey:
@MaxiBul Sorry, but why would you prefer me not dropping Rhino? Are you dependent on it somehow?

JDougherty:
@segrey - I can't speak for MaxiBul, but I am dependent on Rhino. The system where I need to use JS Hint runs on SPARC architecture, which node doesn't support but Rhino does.

segrey:
@JDougherty My bad, sorry. I was wondering if @MaxiBul is dependent somehow on me using JSHint+Rhino in my project. I'm definitely not in a position of dropping Rhino in jshint project.

MaxiBul:
@jugglinmike Looks like #2855 is better solution with tests. Now I use node standalone dist (like jshint-rhino, without external library dependences). It is faster than Rhino :)

PR #2781, opened 2015-11-26
Update jscs to 2.6 and jshint to 2.9

paladox:


.jscsrc
```diff
@@ -5,7 +5,6 @@
       "allowComments": true,
       "allowRegex": true
     },
-    "validateIndentation": 2,
```
nicolo-ribaudo: What about using this?
```json
"validateIndentation": {
  "value": 2,
  "allExpect": [ "comments" ]
}
```

paladox: Thanks.

PR #2783, opened 2015-11-26
Update jquery to 1.11.3

paladox:


jugglinmike:
The version of jQuery included in this project's `test/` directory is used as input for JSHint, not direct execution. Because of that, its outdatedness is not a concern. Changing it correctly would require also changing the expected errors encountered, but that isn't desirable in this case. We lint the jQuery source code as a way to identify regressions between releases, so stability there is important. Thanks anyway!

Issue #2654, opened 2015-09-03
Prevent '{variable} is already defined' when in pseudo-scope


pnstickne:
The following code causes "'x' is already defined" errors when running under anything but `shadow:true` (which is "disable all shadowing checks", and is not desirable when one *wants* shadowing on parameters and whatnot).

    function f() {
      {
        var x = 1;
      }
      {
        var x = 2;
      }
  }

Or, with `for(;;)` pseudo-blocks,

    function f() {
      for (var x = 1; ..) { }
      for (var x = 2; ..) { }
  }

While both in the same scope in JavaScript this is a sometimes useful construct for logic grouping or isolated (ie. can be removed individually) construct.

However, there is a missing duality of [`funcscope`](http://jshint.com/docs/options/#funcscope) which "by default, warns about variables that are used outside of their intended scope".

In the above examples there is the *same* pseudo-scope in play and thus a relax-rule on shadowing would complement such.

Ie (fictions): `pseudoscope`:

> When true, pseudo-scopes are treated as real scopes such that variable declarations 'inside' sibling scopes cannot shadow each other.

----

As http://stackoverflow.com/questions/25102262/is-there-a-way-to-silence-jshints-variable-is-already-defined-warning points out one can set `shadow:true` individually in a function.

jugglinmike:
I like the point you're making about the "missing duality." If we really agree that the two behaviors (block-scoped reference semantics and block-scoped shadowing checks) are coupled, then a new option value is not ideal. Configurations like `funcscope: true, pseudoscope: true` and `funcscope: false, pseudoscope: false` are essentially invalid as they don't reflect a consistent opinion on the treatment of the `var` statement.

A more drastic solution would be to *only enable* inter-block shadowing warnings when `funcscope` is enabled. Then, instead of writing `shadow: true` and opting out of *all* shadowing warnings, the default behavior of JSHint would allow this.

What that would mean is that, by default, JSHint consistently considers `var` as block-scoped. Only via `funcscope: true` would JSHint treat `var` as function-scoped.

Some code examples will probably make this more clear. Note that only in the final case is a change necessary, and the change is to make JSHint *less* strict (so it is backwards-compatible):

```js
// NO CHANGE: referencing `var`s declared in nested blocks (when opting in to `funcscope`)

// jshint funcscope: true
(function() {
{
  var x;
}
void x;
}());
// Today:    no warning
// Proposed: no warning
```

```js
[...]

PR #2789, opened 2015-12-03
[[Fix]] Expand effect of `funcscope` option

jugglinmike:
JSHint generally treats `var` bindings as block-scoped in order to
promote code readability and safety. It exposes the `funcscope` option
to disable this behavior and instead interpret `var` bindings as
function-scoped.

The `funcscope` option *only* controls reporting of W038--the "used out
of scope" warning. In order to consistently apply var-scoping semantics
to `var` statements, the option should also control the emission of
W098--the "unused" warning.

Update JSHint's scope manager to track variable bindings according to
the state of the `funcscope` option.

Closely related: gh-2654

### DIFF
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -1900,6 +1900,31 @@ exports.scope = function (test) {
   test.done();
 };
 
+exports.scopeUnused = function (test) {
+  var src = [
+    '(function() {',
+    '  var x;',
+    '  {',
+    '    var x;',
+    '    void x;',
+    '  }',
+    '  {',
+    '    var x;',
+    '  }',
+    '}());'
+  ];
+
+  TestRun(test)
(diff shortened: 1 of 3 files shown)


tests/unit/options.js
```diff
@@ -1900,6 +1900,31 @@ exports.scope = function (test) {
...
+    '}());'
+  ];
+
+  TestRun(test)
+    .addError(2, "'x' is defined but never used.")
+    .addError(8, "'x' is defined but never used.")
+    .test(src, { shadow: true, unused: true });
+
+  TestRun(test)
+    .test(src, { funcscope: true, shadow: true, unused: true });
```
rwaldron: I like this, and I can't come up with anyways this will fail. How does this interact with block bodies of other forms?

jugglinmike:
Closing in favor of gh-2795.

Issue #2784, opened 2015-11-27
'i' is defined but never used.


ficristo:
The following code pass with 2.8.0 but will fail with jshint 2.9.1-rc1 with:
`test.js: line 4, col 14, 'i' is defined but never used.`
To me seems a bug. (I'm not sure if it is a duplicate or not)

```javascript
/*jshint unused:vars, -W004*/
function test() {
    var array = ['a', 'b', 'c'];
    for (var i in array) {
        var elem = array[i];
        elem[i] = 0;

        var i = "test";
        i = 0;
    }
}
test();
```

jugglinmike:
Thanks for checking out the release candidate and reporting back! Although this
is a new error for your code, I believe it is actually a bug fix.

Historically, JSHint has treated `var` bindings as block-scoped despite the
fact that they are function-scoped. It exposes an option dedicated to disabling
this (technically incorrect) behavior, [`funcscope`](). From the docs:

> This option suppresses warnings about declaring variables inside of control
> structures while accessing them later from the outside. Even though
> JavaScript has only two real scopes—global and function—such practice leads
> to confusion among people new to the language and hard-to-debug bugs. This is
> why, by default, JSHint warns about variables that are used outside of their
> intended scope.
>
>     function test() {
>       if (true) {
>         var x = 0;
>       }
>
>       x += 1; // Default: 'x' used out of scope.
>                 // No warning when funcscope:true
>     }

In keeping with that assumption, the first declaration of `i` in your code
should be scoped to [the special block reserved for the `for` statement's
head](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-forin-div-ofheadevaluation-tdznames-expr-iterationkind).
The second declaration of `i` shadows the first, and all further references are
to that second binding.

Here's a reduced test case:

```js
// jshint unused: true, shadow: true
(function() {
  var x;
  {
    var x;
    void x;
  }
}());
```

For consistency's sake (and even though it is a new error), JSHint should
consider the `var` from line 3 as unused.

But I don't think we're off the hook completely--if `funcscope` is really intended to control the way JSHint interprets `var` bindings, then it ought to effect more than just shadowing detection. [I've opened a pull request](https://github.com/jshint/jshint/pull/2789) to ensure that *this* code does not trigger warnings:

```js
// jshint funcscope: true, unused: true, shadow: true
(function() {
  var x;
  {
    var x;
    void x;
  }
}());
```

[...]

PR #2795, opened 2015-12-06
[[Fix]] Correct `unused` for function-scoped vars

jugglinmike:
In the absence of the `funcscope` option, JSHint validates
function-scoped variable references as though they were block scoped.
JSHint does not apply block-scoping semantics to reference counting,
however: re-declarations within nested blocks are *not* considered
distinct when determining whether a given binding is used.

Ensure that when a variable is re-declared within the same function
scope, usage tracking is "shared" across all identifiers.

This change required correcting two existing tests. Both alterations
removed warnings, meaning they do not reflect a backwards-incompatible
change in behavior.

1. In the following source code, the variable `ok` is used, so JSHint
   should not report otherwise:

        function(match) {
          var ch = this.string.charAt(this.pos);
          if (typeof match == "string") var ok = ch == match;
          else var ok = ch && (match.test ? match.test(ch) :
          if (ok) {++this.pos; return ch;}
        }

2. In the following source code, the `var` declaration for `d` is not
   technically used, but it is also invalid due to ES2015 block scoping
   semantics. Because JSHint continues to correctly emit an error for
   this code, the removal of the "unused" warning has little practical
   effect.

        function sup(a) {
          var b = 4;
          let c = 5;
          let d = 6;
          if (false) {
            var d = 7;
          }
          return b + c + a + d;
        }

### DIFF
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -386,24 +386,43 @@ var scopeManager = function(state, predefined, exported, declared) {
           });
       }
 
-      // if we have a sub scope we can copy too and we are still within the function boundary
+      // If this is not a function boundary, transfer function-scoped labels to
+      // the parent block (a rough simulation of variable hoisting). Previously
+      // existing labels in the parent block should take precedence so that things and stuff.
       if (subScope && !isUnstackingFunctionBody &&
         !isUnstackingFunctionParams && !isUnstackingFunctionOuter) {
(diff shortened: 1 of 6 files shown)


jugglinmike:
Oh, right: this resolves gh-2784.

rwaldron:
LGTM

Issue #2785, opened 2015-11-29
incorrect 'only available in ES6' warning (W119) with options `moz: true, esversion: 6`


myrdd:
__How to reproduce:__

Code:

```javascript
function toArray(...rest) {
  return rest;
}
```

Options:

```json
{
  "esversion": 6,
  "moz": true
}
```

__What happens?__

JSHint says: `'spread/rest operator' is only available in ES6 (use 'esversion: 6'). (W119)`

__What should happen?__

There should be no warning.

__Additional notes__

* I noticed that when `moz` is set to `false` the warning doesn't appear.
* I'm using jshint `v2.9.1-rc1`.

myrdd:
Looks like it happens for more (or all) `W119` warnings, for example when using arrow functions:

```javascript
let fn = (a, b) => a + b;
```

caitp:
It's a one line fix, just remove the requirement that `moz` is false.

myrdd:
I'm trying to understand this case. Referring to `jshint.js` and `state.js`, why is there a "strict" parameter to `inES6()`?
```javascript
// In `jshint.js`:

      if (!state.inES6(true)) {
        warning("W119", state.tokens.next, "spread/rest operator", "6");
      }

// in `state.js`:

  /**
   * @param {boolean} strict - When `true`, only consider ES6 when in
   *                           "esversion: 6" code and *not* in "moz".
   */
  inES6: function(strict) {
    // ...
  },
```

Btw, I also wonder why in commit cf5a699 somewhere [`state.option.esnext` has changed to `state.inES6(true)`](https://github.com/jshint/jshint/commit/cf5a699b0f6fbe2be61f4972ea0a2bbef51a0bd2#diff-7cd6d0e1db309008199ead0cb3a0b76fL1457).

nicolo-ribaudo:
> Btw, I also wonder why in commit cf5a699 somewhere state.option.esnext has changed to state.inES6(true).

It's my fault. I added that parameter because, without it, `...` and arrow functions would be allowed even if `esversion` isn't set and `moz` is true. To maintain the same behaviour of JSHint 2.8.0, `inES6(true)` should be replaced with `esversion === 6`

caitp:
No, Just remove the requirement that moz is false in inES6, it's not rocket science

PR #2796, opened 2015-12-06
[[Fix]] Do not disable ES6 when `moz` is set

jugglinmike:
Although environments implementing Mozilla extensions are not
necessarily ES6-compliant, they are not mutually exclusive. Ensure that
for configurations that specify both ES6 and `moz`, the semantics of
both environments are applied consistently.

Resolves gh-2785.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -22,11 +22,11 @@ var state = {
 
   /**
    * @param {boolean} strict - When `true`, only consider ES6 when in
-   *                           "esversion: 6" code and *not* in "moz".
+   *                           "esversion: 6" code.
    */
   inES6: function(strict) {
     if (strict) {
-      return !this.option.moz && this.option.esversion === 6;
+      return this.option.esversion === 6;
     }
     return this.option.moz || this.option.esversion >= 6;
   },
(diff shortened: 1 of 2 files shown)


src/state.js
```diff
@@ -22,11 +22,11 @@ var state = {
...
    * @param {boolean} strict - When `true`, only consider ES6 when in
-   *                           "esversion: 6" code and *not* in "moz".
+   *                           "esversion: 6" code.
    */
   inES6: function(strict) {
     if (strict) {
-      return !this.option.moz && this.option.esversion === 6;
+      return this.option.esversion === 6;
     }
     return this.option.moz || this.option.esversion >= 6;
```
rwaldron: It's always bothered me that `this.option.moz` is allowed to imply "ES6-ness"

jugglinmike: I agree, but disabling it now would break backwards compatibility. The nice thing is that [we've been careful to introduce more recent ES6 features like `new.target` using the "strict" setting](https://github.com/jshint/jshint/blob/46db9238da535a2d82d7565deac887de6e602ad4/src/jshint.js#L2342-L2344), which encourages new codebases to configure *both* `moz: true` and `esnext: true`

rwaldron:
LGTM

PR #2797, opened 2015-12-07
[[FIX]] Throw W033 instead of E058 when the ; after a do-while stmt is missing

nicolo-ribaudo:
In ES6, `;` is always inserted after a do-while statement (even if there isn't a line break): - [11.9.1 Rules of Automatic Semicolon Insertion](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-rules-of-automatic-semicolon-insertion).

This pr makes JSHint trigger a warning (which can be disabled using `asi: true`) instead of an error.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1597,7 +1597,7 @@ var JSHINT = (function() {
     }
   }
 
-  function parseFinalSemicolon() {
+  function parseFinalSemicolon(stmt) {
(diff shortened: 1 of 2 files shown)


jugglinmike:
Following up with gh-3066.

jugglinmike:
We landed an update version of this in gh-3066. Thanks, @nicolo-ribaudo!

PR #2798, opened 2015-12-07
Use pratt mechanics to parse tagged templates (rather than special casing them)

nicolo-ribaudo:


jugglinmike:
This looks reasonable to me, but since the "TODO" is @caitp's, I'd like to get her input. What do you think, Cait?

jugglinmike:
I decided to put my @caitp hat on (it has rabbit ears) and give this a closer look. I'm pretty certain the binding power of 15 is too low. I made a patch to increase it to a more accurate number and included tests by asserting the behavior of the `singleGroups` option:

https://github.com/jshint/jshint/compare/master...jugglinmike:pratt-tagged-template

@nicolo-ribaudo Please let me know if that looks good to you, or if I'm missing something about your choice of 15.

nicolo-ribaudo:
> @nicolo-ribaudo Please let me know if that looks good to you, or if I'm missing something about your choice of 15.

Sincerely I don't remember why I choosed a binding power lower than the operators' one, but it was obviously wrong. Thank you for having fixed that error in your branch.

jugglinmike:
Merged to `master` at commit 635dce969739026041a0c3bec6f8c32fe65dc49d. Thanks!

Issue #1600, opened 2014-04-01
Freeze option incorrectly catches assignments to local vars


kangax:
With freeze option:

```js
(function(){
  function Array() { }
  Array.prototype.foo = 1;
})();
```

incorrectly reports it as overwriting native Array.

PR #2799, opened 2015-12-07
[[FIX]] Make the 'freeze' option less strict

nicolo-ribaudo:
> Don't warn when the prototype of a local variable
named as a native object is modified.

> Fixes gh-1600

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1340,8 +1340,10 @@ var JSHINT = (function() {
       while (!obj.identifier && typeof obj.left === "object")
         obj = obj.left;
 
-      if (obj.identifier && natives.indexOf(obj.value) >= 0)
+      if (obj.identifier && natives.indexOf(obj.value) >= 0 &&
+          state.funct["(scope)"].isPredefined(obj.value)) {
         return obj.value;
+      }
     }
 
     var prototype = walkPrototype(left);
--- a/tests/unit/fixtures/nativeobject.js
+++ b/tests/unit/fixtures/nativeobject.js
@@ -17,3 +17,13 @@ Boolean.prototype = {
 NonArray.prototype.random = function () {
   return 4;
 };
+
+{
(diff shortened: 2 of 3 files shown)


src/jshint.js
```diff
@@ -1340,8 +1340,10 @@ var JSHINT = (function() {
       while (!obj.identifier && typeof obj.left === "object")
         obj = obj.left;
 
-      if (obj.identifier && natives.indexOf(obj.value) >= 0)
+      if (obj.identifier && natives.indexOf(obj.value) >= 0 &&
+          state.funct["(scope)"].isPredefined(obj.value)) {
```
jugglinmike: Taking this approach means `findNativePrototype` would no longer be purely functional. `jshint.js` has a lot of implicit dependencies between functions, so I think it's worthwhile to defend the rare cases where function behavior is fully defined by explicit input/output. This is easily resolved, though: just move the check to `checkLeftSideAssign`

jugglinmike:
@nicolo-ribaudo I have a request for a different approach. I've certainly taken my time getting to this patch, though, so I'm happy to do it myself (I just wanted to give you the option). Let me know either way, and we'll get this into `master`

src/jshint.js
nicolo-ribaudo: I think it is better to check if the variable is predefined in the `findNativePrototype` function so that it returns something only if you are actually modifying a native prototype. (which is the behaviour suggested by its name)

Anyway, this is a very little difference: feel free to go ahead and move it to `checkLeftAssign`

nicolo-ribaudo:
Merged to `master` at commit b76447cab445d67f26b218f770f889866cf4fa99

src/jshint.js
jugglinmike: You make a good point. The semantics of this function necessarily involve state outside of the expression itself!

Issue #2760, opened 2015-11-13
[2.9.1-rc1] "Incompatible options values" is reported too many times


nicolo-ribaudo:
```js
// jshint strict: global, globalstrict: true

// jshint unused: false

"use strict";
```

```
$ bin/jshint demo.js
demo.js: line 1, col 1, Incompatible values for the 'strict' and 'globalstrict' linting options.
demo.js: line 3, col 1, Incompatible values for the 'strict' and 'globalstrict' linting options.

2 errors
```

E059 should be reported only at line 1

PR #2801, opened 2015-12-09
[[Fix]] Abort in the presence of invalid config

jugglinmike:
> When an invalid set of configuration values is encountered, linting cannot
> continue because the expectations are ambiguous. In this event, report the
> error and cease linting.

Note that the current behavior of JSHint is to issue an error and continue linting. As @nicolo-ribaudo pointed out in gh-2760, a distinct "invalid configuration" error is reported for every inline directive encountered thereafter. In addition, because the program is in an invalid state, it's difficult to say how it will apply linting rules in general. That's why I'm proposing that we halt all linting in this case.

This change is **not** critical. If we want to play it safe while vetting 2.9.1's release candidate, then we can proceed with the existing behavior. The program will fail regardless, so even a patch release could implement this change.  Merging it now simply means that 2.9.1 will be less noisy in response to invalid configurations.

This depends on gh-2749. If we want to be super-conservative, [I wrote another implementation that does *not* depend on that patch](https://github.com/jshint/jshint/compare/jshint:46db923...jugglinmike:767c47d), but I think that version is weaker overall.

Resolves gh-2760

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -330,7 +330,7 @@ var JSHINT = (function() {
 
   // Produce an error warning.
   function quit(code, line, chr) {
-    var percentage = Math.floor((line / state.lines.length) * 100);
+    var percentage = Math.floor((line / state.lines.length) * 100) || 0;
     var message = messages.errors[code].desc;
 
     throw {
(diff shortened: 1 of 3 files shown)


jugglinmike:
@rwaldron I've renamed the variable as requested

rwaldron:
LGTM

PR #2803, opened 2015-12-13
[[Fix]] Disallow ambiguous configuration values

jugglinmike:
While working on gh-2801, @rwaldron and I identified another opportunity to
disallow ambiguous configurations. As of this writing, the `esversion` option
is not yet available in any stable release, which means that we are free to
introduce hard failures in its interpretation. This will not be possible
following the next stable release, and that is why we are considering it during
the release candidate phase.

@rwaldron: This required more refactoring than I expected, but (as I tried to
explain in the commit message) I think it's necessary. During your review,
please pay close attention to the handling of the `moz` option in particular.
As you know, the codebase has always made surprising assumptions about that
option. I believe the simplifications in this patch are all valid, but I would
especially appreciate your verification of that detail.

I want to be conservative with the changes we introduce to a release candidate;
do you this is warranted given the circumstances? We would of course vet this
with the recent fixes in another RC.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -187,19 +187,30 @@ var JSHINT = (function() {
   }
 
   function assume() {
+    var badESOpt = null;
     processenforceall();
 
     /**
      * TODO: Remove in JSHint 3
      */
-    if (!state.option.esversion && !state.option.moz) {
+    if (state.option.esversion) {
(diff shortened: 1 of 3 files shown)


rwaldron:
Nice work!

PR #2808, opened 2015-12-26
Update package.json

paladox:
Update packages to latest version per https://david-dm.org/jshint/jshint#info=dependencies&view=table

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,14 +42,14 @@
   "main": "./src/jshint.js",
 
   "dependencies": {
-    "cli":                 "0.6.x",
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the patch! As I mentioned in a similar pull request (gh-2692), this particular task is really a burden of the maintainers. It may take some time before any of us are able to conduct a thorough review for this changeset, and in the mean time, our dependencies may advance beyond the versions listed here. In the mean time, we're happy to review specific package updates on an as-needed basis.

PR #2809, opened 2015-12-26
Update package.json

paladox:
As jshint 2.9 had problems and was removed straight away. Updating to jshint 2.9.1-rc2 so .x is not needed yet. once 2.9.1 is released we can go back to doing for example 2.9.x.

Updated packages per https://david-dm.org/jshint/jshint#info=devDependencies&view=table

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,19 +53,19 @@
   },
 
   "devDependencies": {
-    "browserify":                     "9.x",
-    "conventional-changelog":         "0.4.x",
-    "conventional-github-releaser":   "0.4.x",
+    "browserify":                     "12.x",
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the patch! As I mentioned in a similar pull request (gh-2692), this particular task is really a burden of the maintainers. It may take some time before any of us are able to conduct a thorough review for this changeset, and in the mean time, our dependencies may advance beyond the versions listed here. In the mean time, we're happy to review specific package updates on an as-needed basis.

ficristo:
IMHO using the latest version of jshint to lint itself would be nice...

Issue #2759, opened 2015-11-13
[2.9.1-rc1] Missing 'use strict' issue introduced


hughes:
I've noticed several files with "use strict" in the top-level function have started failing jshint since 2.9.1-rc1 was published, with the message of a missing "use strict" statement. Anyone else noticed this?

nicolo-ribaudo:
Can you provide an example?

benatwork99:
Yes I have noticed this too.

See: https://travis-ci.org/smart-facility/cognicity-reports-powertrack/builds/90874359

nicolo-ribaudo:
@benatwork99 You get different warnings:

> Incompatible values for the 'strict' and 'globalstrict' linting options.

JSHint reports this warning when you use both `strict: global` and `globalstrict` (#2703)

> Read only.

It looks like `-W020` doesn't work anymore. Can you open a new issue?

randspy:
I also have this problem.

nicolo-ribaudo:
@randspy Can you provide an example / a link to a file to reproduce this issue?

hughes:
Might have actually just been a bugfix related to `strict: "global"` that revealed something in my project that was actually not right. The failing files all looked like this:

```
/*globals describe*/
describe('foo', function() {
  'use strict';
  // ...
});
```

In 2.8, this was passing with `"strict": "global"` in my .jshintrc

nicolo-ribaudo:
The strict option requires statements different from function declarations and IIFEs to be in strict mode. `strict: true` doesn't warn about global level code (while `strict: global` does) because it doesn't allow the use of a global "use strict" directive.

In 2.8.0, using `strict: true, globalstrict: true` you didn't get the error because jshint didn't warn when invocations and other expressions were not in strict mode: (fixed in #2678)
```js
describe(); // jshint didn't warn
var a = 2;   // 'Missing "use strict" statement.', ok
```

> In 2.8, this was passing with `"strict": "global"` in my .jshintrc

`strict: global` was introduced in JSHint 2.9.0, so in 2.8.0 it was interpreted as `strict: true`

benatwork99:
@nicolo-ribaudo You are correct re. the different warnings, I should have read more carefully. Thanks for opening the new issue, I only got back to this today.

jugglinmike:
I think this is safe to close, but @lukeapage's recent assignment of the "Regression" label suggests otherwise. Luke: can you explain the regression here?

jugglinmike:
[...]

PR #2810, opened 2015-12-27
Update package.json

paladox:
Update to jshint 2.9.1-rc2 so it lints its self with the very latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "coveralls":                      "2.11.x",
     "istanbul":                       "0.3.x",
     "jscs":                           "1.11.x",
-    "jshint":                         "2.6.x",
+    "jshint":                         "2.9.1-rc2",
     "mock-stdin":                     "0.3.x",
     "nodeunit":                       "0.9.x",
     "phantom":                        "~0.7.2",


jugglinmike:
Looks like we've been relying on a bug in the implementation of `strict: true`.
jshint.js is failing because the `strict` and `node` options are both set to
`true`, but the file does not include a `use strict` directive on a "global"
level.

#### Some background

When set to `true`, the `strict` linting option requires that strict mode be
enabled for every function in the input source. Here's the option's
documentation from version 2.8.0:

> This option requires all functions to run in ECMAScript 5's strict mode.
> [Strict mode](https://developer.mozilla.org/en/JavaScript/Strict_mode)
> is a way to opt in to a restricted variant of JavaScript. Strict mode
> eliminates some JavaScript pitfalls that didn't cause errors by changing
> them to produce errors.  It also fixes mistakes that made it difficult
> for the JavaScript engines to perform certain optimizations.
>
> *Note:* This option enables strict mode for function scope only. It
> *prohibits* the global scoped strict mode because it might break
> third-party widgets on your page. If you really want to use global
> strict mode, see the *globalstrict* option.

In Node.js code, a "global" `use strict` directive is not actually evaluated in
the global scope and therefore not susceptible to the same problems you would
expect in web browser environments. JSHint has long allowed (but not
*required*) the global directive in Node.js code for this reason.

#### Explanation

Here are a few concrete examples:

```js
// jshint strict: true, node: true
'use strict';
module.exports = function() {
};
```

JSHint@2.8.0 and JSHint@2.9.1-rc2 both validate the above input.

```js
// jshint strict: true, node: true
module.exports = function() {
  'use strict';
};
```

- JSHint@2.8.0 validates this input. This is consistent with a close reading of
  the `strict` documentation published for that release ("avoid the global `use
  strict` directive), despite the recommendation being inappropriate for
  Node.js code
- JSHint@2.9.1-rc2 reports, "Missing "use strict" statement. (E007)" for the
  assignment expression. This is more in-line with the intent of users who
  enabled the option: they want *all* their code to be in strict mode, and in
  Node.js, that can safely be the entire file.

While the new behavior is definitely preferable, it also means new errors for
people upgrading from 2.8.0. Although we previously identified this change as a
"bug fix" (see https://github.com/jshint/jshint/issues/2759), seeing it again
here prompted me to review the old documentation. In light of that closer
reading, I'm inclined to be more conservative here and revert this change,
adding it to the growing list of fixes for our next major release.

jugglinmike:
Following up with a fix in gh-2822

jugglinmike:
Superseded by gh-2822.

Issue #1788, opened 2014-08-07
You might be leaking a variable (exports) here (node.js)


ragamufin:
Hi, I'm running into something related to https://github.com/jshint/jshint/issues/480. The following code triggers the `W120` error "You might be leaking a variable (exports) here":

```
var settings = exports = module.exports = ...
```

This happens even if I set:

```
/*global exports:true*/
```

I often write singletons this way declaring it and setting the exports and module.exports to it in one line for simplicity's sake. I can get around this by declaring the variable before setting it to equal exports but since this is perfectly valid node.js code I figured it would be nice if jshint could handle it. Thanks in advance.

vandigroup:
+1

bewam:
This behavior is the same with constants.
```javascript
const myFirstConstant = mySecondConstant = 'what ever';
```

nicolo-ribaudo:
@bewam In your code, `mySecondConstant` isn't a constant. That code is like
```js
mySecondConstant = 'what ever';
const myFirstConstant = mySecondConstant;
```

thetutlage:
Any updates on this ?

superkhau:
+1

jugglinmike:
It turns out this issue is unrelated to the `node` option. I'm proposing a fix in gh-2821; there's more detail there if you're curious.

Pictor13:
No news about this?
Is it something that one should actually pay attention to? Or can just be ignored/suppressed?

PR #2821, opened 2016-01-02
Improve w120

jugglinmike:
I've split the patch into two commits to more clearly demonstrate the effect of
the change. The first commit simply adds tests (some of which "prove" incorrect
functionality). The second commit fixes the implementation, and updates the
test only by removing the invalid assertions and updating the expected message
for the valid assertions.

(Personally, I think this would best be avoided by disallowing variable
shadowing via `shadow: outer`, but that is a subjective decision that doesn't
help users contributing to codebases that disagree.)

This resolves gh-1788.

Commit message:

> Warning 120 (message "You might be leaking a variable ({a}) here.") is
> triggered when a binding statement contains an assignment expression whose left
> hand side is an identifier. For instance, the assignment to `x` triggers the
> warning in each of the following three statements:
> 
>     var v = x = null;
>     let l = x = null;
>     const c = x = null;
> 
> In these contexts, it is possible that the author mistakenly believes that a
> new binding for the `x` identifier is being created. Warning 120 was presumably
> implemented to help developers avoid unintentionally creating global
> bindings.
> 
> As implemented, JSHint does not consider whether the binding in question has
> already been created, so the following code also triggers the warning (despite
> being safe):
> 
>     var y;
>     var v2 = y = null;
> 
> JSHint maintains environment records for each scope, so this problem could be
> resolved by referencing the containing scope(s). Such a solution would not
> address a more fundamental problem: the warning is mostly duplicative of W117
> ("{a} is not defined."). Users are already alerted of the "leakage" when the
> `undef` option is set. JSHint already operates as expected for the following
> input, issuing W117 for line 2 but *not* for line 4:
> 
>     // jshint undef: true, -W120
>     var v = x = null;
>     var y;
>     var v2 = y = null;
> 
> W117 is preferable for two reasons:
> 
> 1. it is much more robust, as it covers a wide array of additional instances
>    where global bindings are being created
> 2. it is controlled by a dedicated linting option that is human-readable and
>    well documented
> 
> Although this makes the warning a good candidate for outright removal, the
> original intention suggests a special case that W120 is uniquely positioned to
> address: re-assignment of bindings in a "higher" scope. For example:
> 
>     let v1 = 0;
>     {
>       let v2 = v1 = 1;
>     }
> 
> Just as in previous examples, the author of this code may have mistakenly
> assumed that the second `let` statement created a new `v1` binding and that
> each instance of the `v1` identifier referenced a unique block-scoped binding.
> This is incorrect for the same reasons, and the mistake has similar
> implications for program correctness.
> 
> However, unlike previous examples, this mistake would *not* be
> identified by `undef`/W117 because `v1` is defined.
> 
> Refactoring W120 to address this concern is backwards compatabile with
> previous versions of JSHint because this is a special case of the
> previously-implemented behavior (the overall effect is a relaxing of
> linting constraints).
> 
> Re-write the conditions that trigger warning W120 to be limited to only
> those cases where the nested assignment expression may be misinterpreted
> as shadowing a binding in a higher scope. Update the warning message to
> describe the more specific issue.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3570,7 +3570,10 @@ var JSHINT = (function() {
 
       if (state.tokens.next.id === "=") {
         advance("=");
-        if (!prefix && peek(0).id === "=" && state.tokens.next.identifier) {
+        if (!prefix && peek(0).id === "=" && state.tokens.next.identifier &&
+          state.funct["(scope)"].has(state.tokens.next.value) &&
+          !state.funct["(scope)"].block.has(state.tokens.next.value)
+          ) {
           warning("W120", state.tokens.next, state.tokens.next.value);
         }
         var id = state.tokens.prev;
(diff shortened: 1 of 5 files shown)


rwaldron:
This is a lot more to read than the others, so I will look tomorrow.

rwaldron:
Unless I don't, which means I opted to play video games instead ;)

Pictor13:
Ping. No news about this?

PR #2822, opened 2016-01-03
[[FIX]] Do not require global USD for any envs

jugglinmike:
In web browsers, enabling ES5 "strict mode" at a global level effects
all scripts executed in the same context. For this reason, the `strict`
option was originally implemented to enforce the directive's use in
function scope only.  However: the Browserify, Node.js, and PhantomJS
environments do not suffer from this problem.

Recent improvements to the option's behavior included making JSHint's
interpretation of `strict: true` to be contextual: in those environments
where a global `use strict` directive is "safe," enabling the option
would require the global directive.

Although this change is an improvement (because it encourages more code
to be written in strict mode), it is backwards incompatible. Moreover,
the option's documentation in JSHint 2.8.0 explicitly described the
sub-optimal behavior:

> *Note:* This option enables strict mode for function scope only. It
> *prohibits* the global scoped strict mode because it might break
> third-party widgets on your page. If you really want to use global
> strict mode, see the *globalstrict* option.

For this reason, `strict: true` should continue to trigger the "legacy"
behavior. The change brings that configuration completely in-line with
the recently-introduced `strict: func`, and because `strict: func` has
not been released in a stable version, it can safely be removed.

@lukeapage @nicolo-ribaudo This effectively reverts gh-2576, so I'd appreciate if either of you could weigh in--I want to make sure I'm not missing something important!

### DIFF
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -1678,12 +1678,10 @@ exports.strict = function (test) {
   var run = TestRun(test)
     .addError(1, 'Missing "use strict" statement.');
   run.test(code, { es3: true, strict: true });
-  run.test(code, { es3: true, strict: "func" });
   run.test(code, { es3: true, strict: "global" });
   TestRun(test).test(code, { es3: true, strict: "implied" });
 
   TestRun(test).test(code1, { es3: true, strict: true });
-  TestRun(test).test(code1, { es3: true, strict: "func" });
   TestRun(test).test(code1, { es3: true, strict: "global" });
   TestRun(test)
     .addError(1, 'Unnecessary directive "use strict".')
@@ -1695,7 +1693,6 @@ exports.strict = function (test) {
     .addError(7, 'Strict violation.')
     .addError(8, 'Strict violation.');
   run.test(src, { es3: true, strict: true });
-  run.test(src, { es3: true, strict: "func" });
   run.test(src, { es3: true, strict: "global" });
 
   run = TestRun(test)
@@ -1709,7 +1706,6 @@ exports.strict = function (test) {
     .test(src3, {es3 : true});
 
   TestRun(test).test(code2, { es3: true, strict: true });
-  TestRun(test).test(code2, { es3: true, strict: "func" });
   TestRun(test)
     .addError(1, 'Missing "use strict" statement.')
     .test(code2, { es3: true, strict: "global" });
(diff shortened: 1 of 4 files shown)


jugglinmike:
This resolves gh-2759.

jugglinmike:
This is blocking the release of version 2.9.1. Although @lukeapage and/or
@nicolo-ribaudo's input would be really valuable here, I'd be willing to move
forward after a review from any of the maintainers. @rwaldron @caitp: do either
of you have time to review and (possibly) merge? My plan is to publish another
release candidate with just this fix and hopefully following that with the
stable release after a few days.

nicolo-ribaudo:
This PR looks good to me (I can't think of a better approach to fix the regression).

Will the stricter behaviour be reintroduced in JSHint 3?

tests/unit/options.js
```diff
@@ -1745,6 +1736,42 @@ exports.strict = function (test) {
...
+    "(function() { 'use strict'; void 0; }());"
+  ];
+  TestRun(test, "")
+    .addError(2, "Missing \"use strict\" statement.")
+    .test(partialStrict, { strict: true, browserify: true });
+  TestRun(test, "")
+    .addError(2, "Missing \"use strict\" statement.")
+    .test(partialStrict, { strict: true, node: true });
+  TestRun(test, "")
+    .addError(2, "Missing \"use strict\" statement.")
```
rwaldron: nit: can you just use single quotes inside the string here to avoid the escaping? Feel free to ignore this

rwaldron:
I'm with @nicolo-ribaudo here, I also cannot think of a better way to resolve the regression.

@jugglinmike I'll leave it up to you to change the quoting or not, otherwise this looks good to me and is ready to merge.

tests/unit/options.js
jugglinmike: I agree with the sentiment, but this is unavoidable because [the assertion uses a strict equality to compare "expected" and "actual" string values](https://github.com/jshint/jshint/blob/cfe6c43548debf11b4f7b4098d79a1316ca420a7/tests/helpers/testhelper.js#L67)

rwaldron: Ah! Ok then nevermind. Thanks for the reminder

jugglinmike:
@nicolo-ribaudo I think it would be good to re-introduce this in version 3.

PR #2827, opened 2016-01-11
fix some sonar alert

gpartenet:


rwaldron:
Can you offer any explanation for these changes? Thanks!

gpartenet:
Sure.
 I use SonarQube to analyse a project where jshint is integrated. And it reports some "Avoid trailing comma in array and object literals" issue due to IE throwing a meaningless error in this case.
Even if not relevant for jshint, this patch turn off those issues

nicolo-ribaudo:
> due to IE throwing a meaningless error in this case

Trailing commas works fine in IE9+. IE <= 8 isn't supported (for example, because of [jshint.js#L5189](https://github.com/jshint/jshint/blob/3fa9ecef7ef1f2f9a40f4d6517b844feb068bc23/src/jshint.js#L5189)), so they aren't a problem.

Anyway, I think you should add [`disallowTrailingComma`](http://jscs.info/rule/disallowTrailingComma) to `.jscsrc`, if this PR gets accepted.

gpartenet:
I may just send a request to sonar js plugin to change this rule if it's outdated since IE9+

jugglinmike:
Sounds good to me!

PR #2829, opened 2016-01-15
[[CHORE]] Pass correct parameter to state.inES6

nicolo-ribaudo:


src/jshint.js
```diff
@@ -3858,7 +3858,7 @@ var JSHINT = (function() {
     var generator = false;
     if (state.tokens.next.value === "*") {
       advance("*");
-      if (state.inES6({ strict: true })) {
+      if (state.inES6(true)) {
```
nicolo-ribaudo: `state.inES6` definition:
```js
inES6: function(strict) {
    if (strict) {
```

jugglinmike:
Good catch! Merged to `master` at commit d089d7d5caaf2c0f1f61dc9d9e471ca4bdc5bee4

PR #2830, opened 2016-01-16
[[CHORE]] Update JSHint

jugglinmike:
The latest stable version of JSHint revealed an erroneous configuration:
when the `console` identifier is listed as a global variable, JSHint
creates an internal "label" to track references to that binding. The
local `console` variable in `jshint.js` shadows this global reference,
meaning the global binding is never referenced.

Lint the in-development codebase with the latest stable version of
JSHint. Update the configuration to avoid the new (valid) warning.

Supersedes gh-2810.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "coveralls":                      "2.11.x",
     "istanbul":                       "0.3.x",
     "jscs":                           "1.11.x",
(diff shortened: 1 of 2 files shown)


jugglinmike:
Closing in favor of gh-2860.

Issue #2838, opened 2016-01-20
Incorrect "used before declared" warning for const value


twhb:
Given the following code:

```javascript
/* jshint esversion: 6, devel: true */

window.foo = function () {
  alert(x);
};

window.bar = function () {
  alert(x);
};

const x = 1;
```

I get a single warning: "15: 'x' was used before it was declared, which is illegal for 'const' variables."

The code runs fine. I'm not sure if some people consider this bad style, but either way it doesn't seem to be something JSHint means to enforce, as you get no warning if you comment out either `foo` or `bar`.

Tested in v2.9.1.

jugglinmike:
Thanks for the report! This has been fixed and will be available in the next release of JSHint (post 2.9.1)

PR #2842, opened 2016-01-21
[[FIX]] Don't throw E056 for vars used in two functions

nicolo-ribaudo:
Fixes gh-2838


jugglinmike:
The fix is good, and your test demonstrates it. The test does not cover the lines that you added, though. I added another commit with two more tests and merged the result. Thanks, @nicolo-ribaudo!

Issue #2836, opened 2016-01-19
Incompatible values for the 'strict' and 'globalstrict' complains after upgrading to jshint@2.9.1


amrish7:
On upgrading to `2.9.1`, jshint has started complaining about `Incompatible values for the 'strict' and 'globalstrict'` for the inline comments I put in to bump up the `maxcomplexity` threshold  

```
foo(data, callback) {
    /*jshint maxcomplexity:11 */
   .
   .
   .
}
```

Wondering what is the right way to do the overrides going forward?

jugglinmike:
Could you share a small example that demonstrates this problem all on its own? Please include the necessary configuration values as in-line directives (e.g. `// jshint strict: true`) to help us reproduce the problem.

amrish7:
I have created `cyclomaticComplexity.js` file in https://github.com/amrish7/jshint-failure explaining the issue that I am running into. After upgrading to jshint@2.9.1, this [line] (https://github.com/amrish7/jshint-failure/blob/master/cyclomaticComplexity.js#L4) does not seem to help with bumping the max complexity

jugglinmike:
Thanks for taking the time to put that together! This is a regression. Here's a simplified test case:

```js
// jshint strict: true, globalstrict: true
// jshint maxcomplexity: 1
```

To reproduce this, the `maxcomplexity` option must be set in a separate directive.

This is a high priority bug--thanks for reporting!

ooskapenaar:
@jugglinmike, any idea when this fix will make it into a version of jshint?

jugglinmike:
@ooskapenaar Good news--there's a patch over at gh-2843. I wrote it myself, so it just needs another maintainer's review, and then it can land in `master`. I plan to follow that up with a patch release promptly.

jishi:
I see this for what I can tell, is any overrides in any file, for example:

/*jshint unused:false */

Reverting to 2.8 fixed it. Just FYI, will test next version when it is released.

ooskapenaar:
@jishi, I am seeing the same.

PR #2843, opened 2016-01-23
[[FIX]] Do not fail on valid configurations

jugglinmike:
In addressing gh-2836, I ran across some inefficient/noisy behavior. Rather
than address that as part of the bug fix, I'm going to follow up with a
separate commit. For now, I'm submitting a separate commit to assert the
undesirable behavior, just to document that it isn't a regression introduced by
my bug fix.

Resolves gh-2836.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -215,9 +215,6 @@ var JSHINT = (function() {
     }
 
     if (state.option.module) {
-      if (state.option.strict === true) {
-        state.option.strict = "global";
-      }
       /**
        * TODO: Extend this restriction to *all* ES6-specific options.
        */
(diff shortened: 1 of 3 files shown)


rwaldron:
LGTM

jmustonen:
Sorry if this is the wrong place to ask this, but do you have any schedule set for a new 2.9.x release, which would include also this patch?

Marshallx:
Ditto, the nodejs package is still on 2.9.1

jugglinmike:
This fix has been released in JSHint 2.9.2 http://jshint.com/blog/2016-04-19/release-2-9-2/

Issue #2852, opened 2016-01-28
No CLI for Rhino


cscott:
The distributed `jshint-rhino.js` used to invoke the CLI, but the CLI has since been factored out.  Further, the CLI now uses the npm `cli` package, which does not apparently support browserify.

PR #2853, opened 2016-01-28
Save rhino global object `this` in `global` variable.

cscott:
This is a rebase of https://github.com/segrey/jshint/commit/1824d740c2d6656e663de43cf6d95fc8a642c796 to latest master.

This fixes the rhino build, but does not replace the CLI -- the only jshint-rhino.js package invoked the CLI, but since 2.6.3 the CLI has been factored out.  Filed #2852 for this issue.

Fixes: #2308

PR #2855, opened 2016-01-29
Restore CLI functionality to `jshint-rhino.js`

cscott:
Fixes: #2852


cscott:
This patch depends on #2853 and #2856.  I've included them in a merge patch at the top of this branch to demonstrate that `travis` passes, but it would probably be best to rebase after merging #2853 and #2856.

package.json
```diff
@@ -34,7 +34,8 @@
     "test-browser": "node tests/browser",
     "test-cli": "nodeunit tests/cli.js",
     "test-node": "npm run test-unit && npm run test-cli && npm run test-regression",
-    "test-regression": "nodeunit tests/regression",
+    "test-regression": "nodeunit tests/regression && node bin/jshint src",
```
nicolo-ribaudo: Why do you run JSHint here?

package.json
```diff
@@ -34,7 +34,8 @@
...
+    "test-rhino": "java -jar tests/regression/libs/rhino-1.7.7.jar -opt -1 dist/jshint-rhino.js src",
```
nicolo-ribaudo: Does this command only run the packaged JSHint file? Is it possible to run the tests from the `tests/` dir?

cscott: Yes, it "only" runs jshint over itself.  Without reimplementing `nodeunit` for rhino, I don't believe that the unit tests can be run.  It would certainly be nicer to run the unit tests (a future patch, perhaps!) but this way we at least have *some* idea of whether the rhino build is still roughly working.  It will catch most problems of the #2853 type, where the rhino build fails to start up at all.

PR #2856, opened 2016-01-29
Run this version of jshint over src during testing; fix a unused def error.

cscott:
We set `exported` and `global` flags for `console` in `src/jshint.js`, but `console` wasn't actually an exported global.  The current version of jshint was therefore correctly flagging it as an erroneous unused variable.  Tweak the flags to correct this, and add a pass to `npm test` which will run the current version of jshint over the `src` directory to prevent such errors from occurring in the future.


nicolo-ribaudo:
Isn't this a duplicate of #2830?

package.json
```diff
@@ -34,7 +34,7 @@
     "test-browser": "node tests/browser",
     "test-cli": "nodeunit tests/cli.js",
     "test-node": "npm run test-unit && npm run test-cli && npm run test-regression",
-    "test-regression": "nodeunit tests/regression",
+    "test-regression": "nodeunit tests/regression && node bin/jshint src",
```
nicolo-ribaudo: Why do you run JSHint here?

jugglinmike:
Hi @cscott--I have an open pull request to fix the warning over at gh-2830. I
prefer that approach because it is slightly more direct (no need to cherry-pick
lines to ignore) and doesn't rely on buggy behavior (which I only just
discovered through reviewing this patch--see gh-2858).

I think I'd like to close this pull request in favor of this one. Will that
work for you?

cscott:
Actually, I think the change to the `package.json` is the more important change here, not fixing the warning.  Routinely running the latest `jshint` over itself will ensure that any future regressions of this sort are found and fixed quickly.  To answer @nicolo-ribaudo's question, I added the test to the `test-regression` clause because we already are running the latest JSHint over a number of third-party codebases to catch regressions.  This is effectively just adding JSHint itself as another "third party" codebase to find regressions in.  (That is, the earlier lint verified that some older "stable" version of JSHint correctly linted the code; if the pass in `test-regression` fails it means that the newer JSHint is failing to pass code which the stable JSHint accepts, which is likely to be a regression.)

jugglinmike:
> Actually, I think the change to the package.json is the more important change here, not fixing the warning.

Fair enough. I'll update that patch accordingly.

> This is effectively just adding JSHint itself as another "third party" codebase to find regressions in.

An important aspect of our current "regression" tests is that they operate on static input. The current state of the JSHint source code is not static, so it can't serve that same purpose. The good news is that we are using JSHint to lint the source with every invocation of `npm test`, so (combined with your previous suggestion to use the edge version of the project), we already have a mechanism to catch these sorts of problems.

cscott:
I think the proposed solution (#2860) isn't quite as good actually, since it doesn't catch *regressions* -- which was the big problem with 2.9.0, as I understand it.  I actually prefer the approach here -- lint the code using *both* the "stable" version of jshint (or last released-to-npm version, if you prefer) *and* the "edge" version (latest code).  This ensures that we are aware when jshint's behavior is different in edge and stable.  In particular, if "edge" inadvertently stops reporting a violation, we don't want to have those violations silently start to pile up in the codebase.  IMO, at least.

PR #2846, opened 2016-01-26
[[FIX]] Report character position for camelcase errors

danielbarela:
Character positions were not being reported when camelcase errors were detected.
This fix will now report the character position correctly.

Fixes #2845

PR #2857, opened 2016-01-30
[[FIX]] Emit correct token value from "module" API

jugglinmike:
I found this bug while reviewing gh-2846. The fix there is valid, but it did
not address an underlying problem with the little-known "module" API.

Unfortunately, neither JSHint nor NodeUnit are built to support testing this
behavior effectively. As noted in the first (test-only) commit, I've factored
the only test for the API into its own file and encapsulated necessary
setup/teardown logic there.

The second commit addresses the bug.

As an aside: I'm not convinced that this API is useful (or used!), but it ought
to at least be correct.

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -1776,7 +1776,7 @@ Lexer.prototype = {
         this.triggerAsync("Identifier", {
           line: this.line,
           char: this.char,
-          from: this.form,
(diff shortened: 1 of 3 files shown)


rwaldron:
Good to get it covered. LGTM

PR #2860, opened 2016-01-31
[[CHORE]] Use project in latest state when linting

jugglinmike:
The latest stable version of JSHint revealed an erroneous configuration:
when the `console` identifier is listed as a global variable, JSHint
creates an internal "label" to track references to that binding. The
local `console` variable in `jshint.js` shadows this global reference,
meaning the global binding is never referenced.

Lint the in-development codebase with the tool in its current state.
Update the configuration to avoid the new (valid) warning.

Supersedes gh-2830

@rwaldron I have some nagging feeling that using master to lint master may be problematic. I can't substantiate it, though.


cscott:
@jugglinmike You need to add `node` to the beginning of the command in order to make appveyor like it -- look how I did it at #2856.

rwaldron:
The appveyor failure doesn't appear to be caused by this patch, but you might want to restart it. Can you ping me here or on slack when that's ready?

jugglinmike:
@rwaldron I took @cscott's advice and updated the invocation pattern to satisfy the Windows build. All green!

Issue #2863, opened 2016-02-02
Expecting syntax error for expression `();`


vilic:
Can be reproduced on current jshint.com.

jugglinmike:
This may come as a surprise, but some of our users are likely *depending* on this behavior. You can read more in the proposed fix over at gh-2865. In the mean time, I am going to label this as low-priority because it represents a small edge case.

Thanks for the report!

PR #2865, opened 2016-02-02
[[FIX]] Disallow () as expression

nicolo-ribaudo:
Fixes gh-2863

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2508,22 +2508,26 @@ var JSHINT = (function() {
       return doFunction({ type: "arrow", parsedOpening: true });
     }
 
+    if (state.tokens.next.id === ")") {
+      error("E030", state.tokens.next, ")");
+      advance(")");
+      return;
+    }
+
     var exprs = [];
+      
+    for (;;) {
+      exprs.push(expression(10));
 
-    if (state.tokens.next.id !== ")") {
-      for (;;) {
-        exprs.push(expression(10));
-
-        if (state.tokens.next.id !== ",") {
-          break;
-        }
-
-        if (state.option.nocomma) {
-          warning("W127");
-        }
-
-        parseComma();
+      if (state.tokens.next.id !== ",") {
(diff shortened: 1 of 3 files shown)


tests/unit/fixtures/gh826.js
```diff
@@ -19,14 +19,14 @@
...
+            /*jshint ignore:start */
 			return (
-				/*jshint ignore:start */
 				<div data-foo-id={this.props.foo}>
 					/*some other comment*/
 					<input type="text" value={this.props.bar} />
 				</div>
-				/*jshint ignore:end */
 			);
+            /*jshint ignore:end */
```
nicolo-ribaudo: This is a breaking change

caitp: is there some way we can flag that some content was ignored in the parser, and skip the warning if it was encountered?

jugglinmike:
Thanks! De-prioritizing until the next major release.

src/jshint.js
```diff
@@ -2508,22 +2508,26 @@ var JSHINT = (function() {
       return doFunction({ type: "arrow", parsedOpening: true });
     }
 
-    var exprs = [];
+    if (state.tokens.next.id === ")") {
```
caitp: So there are a few things here:

first off: lets put this check behind an enforcing flag, so that this is not actually a breaking change. We could allow `()` by default, but error if an option is enabled. This could later be turned into a leniency option in the next version.

second: `() => { /* do stuff */ }` is a valid and useful pattern, so we'll also want to make sure it's actually an empty expression, and not [ArrowFormalParameters](https://tc39.github.io/ecma262/#prod-CoverParenthesizedExpressionAndArrowParameterList).

caitp: ah, after reading the code a bit more, that second point seems to be irrelevant. so, lgtm but I think we still want a flag here to avoid the breaking change.

PR #2875, opened 2016-02-11
[[FEAT]] Added preset variables for protractor

riichard:
With other presets I expected that there would be a protractor preset
available as well. Now I can simply state in the .jshintrc `"protractor":true`
so it won't trigger errors on the default functions anymore.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -283,6 +283,10 @@ var JSHINT = (function() {
       combine(predefined, vars.jasmine);
     }
 
+    if (state.option.protractor) {
+      combine(predefined, vars.protractor);
+    }
+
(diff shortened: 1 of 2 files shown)


jugglinmike:
Accepting library-specific globals seems like a slippery slope of supporting all sorts of lesser-used libraries. It's true that JSHint currently supports a few, but we're trying to avoid committing to any more.

Instead of trying to define which libraries are "popular enough" to justify inclusion, we've been reserving environment settings for true environments. This also alleviates this project of the burden of maintaining information that is defined by another library (which potentially confuses release cycles).

riichard:
I understand that. I'll leave this pull request for what it is then.

Issue #2876, opened 2016-02-12
Redefinition bugs


tmpfs:
I am not sure if this is my mistake but it seems like a bug to me, `jshint v2.9.1`.

I am getting redefinition warnings on:

```
var setup = require('../setup/home');
```

And:

```
var suite = new benchmark.Suite();
```

Feel free to correct me but this appears to be a regexp/comparison that is a little too simple.

nicolo-ribaudo:
You are declaring the `setup` var two times. You have to omit the `var` keyword in the second line.

tmpfs:
@nicolo-ribaudo , I got your comment in my email but the github interface is not showing it right now (probably a caching bug), you said:

> You are declaring the setup var two times. You have to omit the var keyword in the second line.

This is the file:

https://github.com/tmpfs/rlx/blob/master/test/util/teardown/home.js

This is the error message:

```
test/util/teardown/home.js: line 1, col 5, Redefinition of 'setup'.
```

I don't see any redefinition there. The second example I just renamed the variable to avoid it but trust me there was no redefinition ;)

tmpfs:
Please try it, create a file where the require path has the same string as the variable name, ie:

```
var setup = require('../setup/home');
```

Then run `jshint` on it, I get:

```
jshint test/util/teardown/home.js
test/util/teardown/home.js: line 1, col 5, Redefinition of 'setup'.
test/util/teardown/home.js: line 1, col 5, 'setup' is defined but never used.
```

Obvously the second error is irrelevant because I deleted everything else in the file to test it.

tmpfs:
Also please try the second example in a standalone file, I am sure there was no redefinition.

Do you want me to create more files to reproduce the issues for you?

nicolo-ribaudo:
@tmpfs I deleted my comment because I saw I was wrong

tmpfs:
Neither it appears, `.jshintrc` looks like this for the project:

```
{
  "laxcomma"                : true,
  "laxbreak"                : true,
  "node"                    : true,
  "curly"                   : true,
  "bitwise"                 : false,
  "undef"                   : true,
  "eqeqeq"                  : true,
  "noarg"                   : true,
  "mocha"                   : true,
  "unused"                  : true,
  "asi"                     : true
}
```

Any ideas? Do you want me to try to reproduce this in isolation?

tmpfs:
I think it might be a regression because I only started to come across this after upgrading to 2.9.1, sadly I forget which version I was using before.

[...]

PR #2877, opened 2016-02-13
[[FIX]] Account for implied closures

jugglinmike:
@rwaldron I'm not a fan of this patch. But it does what it needs to do.

I started with an approach that I think is superior, but I eventually
discovered that it is incompatible with the legacy behavior of "environmental"
options:

https://github.com/jshint/jshint/compare/master...jugglinmike:implied-closure

I'm sharing that link here because I'm not convinced that the breakage is
meaningful. It might be a small enough edge case that we try it out in an RC.

Commit message:

> Some environments wrap code in an immediately-invoked function
> expression prior to evaluation. This behavior has an effect on global
> `var` declarations. Unlike in typical JavaScript source code, such
statements in these contexts do *not* create new entries in the
> environment record, nor do they resolve to references to the global
> binding (where defined).
>
> Update the logic that tracks global definitions to account for such
> environments, avoiding issuing warning W079 ("Redefinition of '{a}'.")
> which is otherwise appropriate.
>
> Note: a far preferable solution would involve updating JSHint's scope
> tracking mechanism to create a new entry for this implicit function
> scope. Unfortunately the legacy implementation of the environmental
> options allowed for usage scenarios that are incompatible with that
> approach (specifically: enabling the environment options after the
> source text has been partially parsed).

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3624,7 +3624,7 @@ var JSHINT = (function() {
       for (var t in tokens) {
         if (tokens.hasOwnProperty(t)) {
           t = tokens[t];
-          if (!implied && state.funct["(global)"]) {
+          if (!implied && state.funct["(global)"] && !state.impliedClosure()) {
             if (predefined[t.id] === false) {
               warning("W079", t.token, t.id);
             } else if (state.option.futurehostile === false) {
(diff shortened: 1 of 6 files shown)


jugglinmike:
Oh, right: resolves gh-2876

rwaldron:
> Oh, right: resolves gh-2876

If you include "Fixes gh-2876" in the commit message, that issue will automatically close, with a reference link, when your commit is merged.

rwaldron:
Also, LGTM

PR #2879, opened 2016-02-21
[[FIX]] Reserve `await` keyword in ES6 module code

jugglinmike:
(Little noisier than I would like... Use [`?w=1` to see a simplified changeset for the `isReserved` function](https://github.com/jshint/jshint/pull/2879/files?w=1).)

From the ECMAScript 2015 specification:

> The following tokens are reserved for used as keywords in future
> language extensions.
>
>     Syntax
>         FutureReservedWord::
>             enum
>             await
>
> await is only treated as a FutureReservedWord when Module is the goal
> symbol of the syntactic grammar.

http://www.ecma-international.org/ecma-262/6.0/#sec-future-reserved-words

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -134,22 +134,28 @@ var JSHINT = (function() {
     }
     var meta = token.meta;
 
-    if (meta && meta.isFutureReservedWord && state.inES5()) {
-      // ES3 FutureReservedWord in an ES5 environment.
-      if (!meta.es5) {
+    if (meta && meta.isFutureReservedWord) {
+      if (meta.moduleOnly && !state.option.module) {
         return false;
       }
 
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -134,22 +134,28 @@ var JSHINT = (function() {
     }
     var meta = token.meta;
 
-    if (meta && meta.isFutureReservedWord && state.inES5()) {
-      // ES3 FutureReservedWord in an ES5 environment.
-      if (!meta.es5) {
+    if (meta && meta.isFutureReservedWord) {
+      if (meta.moduleOnly && !state.option.module) {
```
rwaldron: Slick

rwaldron:
LGTM

PR #2881, opened 2016-02-27
[[FIX]] Don't attempt to indent errors using data from other files, during extract with multiple files

### LINKED ISSUE #2778: "Warning: Invalid array length" when running jshint
Running jshint on my project gives the following output:

```
Running "jshint:all" (jshint) task

   src/Binding.js
     69 |    for (key in bindings) {
Warning: Invalid array length Use --force to continue.
```

It's possible that this is just an unhelpful message but it looks to me like a JavaScript error.

`npm list jshint` gives:

```
└─┬ grunt-contrib-jshint@0.11.3
  └── jshint@2.8.0 

``` 

The issue can be reproduced by cloning my project from here:
https://github.com/MartinRixham/Datum
and running `grunt`.

AMoo-Miki:
This change will fix the report of `col NaN` and the problem of `NaN` in `error.character` while checking JS code extracted from non-JS files, when more than one file has been passed in and their lengths don't match.

The function `extractOffsets` finds the number of characters a JS block must be shifted to show the correct character column for an error. The result is an array consisting of numbers, for the block it checked and, `undefined` for the lines it didn't have to look at.
`cli.run` executes `extractOffsets` on each input file but wrongly tries to indent all the errors found so far, across files, using the current file's offsets. This commit makes sure `cli.run` only indents errors of a file against the offsets fetched for the same file.
This commit also makes sure that we don't change or ruin the value of `error.character` unnecessarily.

I have chosen to trust that `extractOffsets` will return an array with only numbers (as a result of Array.length) and falsy values, to avoid the unnecessary type conversion.

A testcase named `usingMultipleFiles` has been introduced.

Fixes #2778

### DIFF
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -633,7 +633,7 @@ exports.group = {
 
     test.done();
   },
-  
+
   testMultipleIgnores: function (test) {
     var run = this.sinon.stub(cli, "run");
     var dir = __dirname + "/../examples/";
@@ -642,10 +642,10 @@ exports.group = {
     cli.interpret([
       "node", "jshint", "file.js", "--exclude=foo.js,bar.js"
     ]);
-    
+
     test.equal(run.args[0][0].ignores[0], path.resolve(dir, "foo.js"));
     test.equal(run.args[0][0].ignores[1], path.resolve(dir, "bar.js"));
-    
+
     test.done();
   },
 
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Can you write some tests in `tests/cli.js`?

AMoo-Miki:
@nicolo-ribaudo done :)

tests/cli.js
```diff
@@ -1150,6 +1150,53 @@ exports.extract = {
...
+      "node", "jshint", "indent.html", "another.html", "--extract", "auto", "--reporter=reporter.js"
+    ]);
+    test.equal(cli.exit.args[0][0], 2);
+
+    test.equal(errors.length, 2, "found two errors");
+    var lintError = errors[0].error;
+    test.ok(lintError, "have error object");
+    test.equal(lintError.code, "W033", "found missing semicolon warning");
+    test.equal(lintError.line, 2, "misaligned line");
+    test.ok(!isNaN(lintError.character), "error position is a number");
```
jugglinmike: This assertion is subsumed by the one that follows it.

tests/cli.js
```diff
@@ -1150,6 +1150,53 @@ exports.extract = {
...
+    test.equal(lintError.character, 6, "first misaligned character at column 2");
+
+    lintError = errors[1].error;
+    test.ok(lintError, "have error object");
+    test.equal(lintError.code, "W030", "found an expression warning");
+    test.equal(lintError.line, 3, "misaligned line");
+    test.ok(!isNaN(lintError.character), "error position is a number");
```
jugglinmike: This assertion is subsumed by the one the follows it.

jugglinmike:
Thanks for the patch, @AMoo-Miki! In fixing the problem of sharing errors across files, I *think* you completely avoided the problem of NaN-value-creation. Can you verify that?

If so, then the new check for `offsets[line]` isn't necessary. It's probably
fine to keep the change as you've written it, if only for code safety. I'm
bringing this up because if it *is* still necessary, I'd like to include a
dedicated test for it.

AMoo-Miki:
I was away for a few weeks on account of our new year. Apologies for the late response.

`offsets` is the output of `extractOffsets` which is an array that contains numbers for lines within an identified script block.

If `offsets[line]` is asked for a line which has not been handled by the parser, it would result in `undefined` and will ruin the entire error report with its `NaN`.

That `offsets[line]` at the end of the condition is meant to avoid that.

Also, I hinted some of my bad HTML files and I didn't get NaNs, so I too think they are avoided. Of course I cheated by doing that last `offsets[line]` check!

jugglinmike:
No worries about the delay. Thanks for confirming that the check isn't technically necessary.

Could you remove the two assertions that I indicated are not necessary?

AMoo-Miki:
@coveralls you are annoying!

@jugglinmike, I removed those 2 assertions.

jugglinmike:
Great--I've merged your patch to `master` at commit 47daf76632d70e9dea8fc59a5bf6faf2db9ff41f. Thank you!

PR #2883, opened 2016-03-01
[[FIX]] Add CompositionEvent to the list of browser globals

alexeykuzmin:
https://www.w3.org/TR/uievents/#events-compositionevents


jugglinmike:
Merged to `master` at commit 56515cf5444a06a0f02970536deaac84cf88227e. I had never heard of CompositionEvent before--interesting. Thanks for the patch!

PR #2888, opened 2016-03-07
[[FIX]] Return a non-zero exit code when file not found

### LINKED ISSUE #2884: What is the expected behaviour for non-existent files?
I'm using `jshint` on `npm` and command line. When a file does not exist,  is `jshint` suppose to exit 0?

```sh
   jshint  file-does-not-exist
```

tanishalfelven:
Issue #2884 noted that when a file is not found the CLI returns with a zero exit code. This commit m
akes the program return 1 when a file can't be opened.

Fixes #2884

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2575,7 +2575,9 @@ var JSHINT = (function() {
           // Used to delineate an integer number literal from a dereferencing
           // punctuator (otherwise interpreted as a decimal point)
           (ret.type === "(number)" &&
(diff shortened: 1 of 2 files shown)


caitp:
why closed PR? I think it's a good feature, but it needs a flag

tanishalfelven:
I started working on something else and forgot to branch. SO, had to take a few steps back - but there is a new PR up.

Oops.

PR #2889, opened 2016-03-07
[[FIX]] Return a non-zero exit code when file not found

### LINKED ISSUE #2884: What is the expected behaviour for non-existent files?

tanishalfelven:
Issue #2884 noted that when a file is not found the CLI returns with a zero exit code. This commit makes the program return 1 when a file can't be opened.

Fixes #2884

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -422,7 +422,7 @@ function collect(fp, files, ignores, ext) {
 
   if (!shjs.test("-e", fp)) {
     cli.error("Can't open " + fp);
-    return;
+    exports.exit(1);
   }
 
   if (shjs.test("-d", fp)) {


src/cli.js
```diff
@@ -422,7 +422,7 @@ function collect(fp, files, ignores, ext) {
 
   if (!shjs.test("-e", fp)) {
     cli.error("Can't open " + fp);
-    return;
+    exports.exit(1);
```
caitp: @jugglinmike do you think this should be driven by a flag?

jugglinmike: @caitp I didn't consider that. I think a non-zero exit code for faulty file
references is common behavior for many tools (at least, those that are not
expected to create non-existent files):

    $ cat foobar; echo $?
    cat: foobar: No such file or directory
    1
    $ less foobar; echo $?
    foobar: No such file or directory
    1
    $ ls foobar; echo $?
    ls: cannot access foobar: No such file or directory
    2
    $ d8 foobar; echo $?
    Error reading 'foobar'
    1

There may be a case to be made for backwards compatibility, but the current
behavior is unspecified and (I think) surprising. Were you suggesting this to
avoid disrupting existing build processes?

jugglinmike:
Hi @tanishalfelven. Despite the silence since our initial review, I have been
thinking about this patch. I'm concerned about a potential use case that this
change might disrupt:

    $ jshint existent-file.js non-existent-file.js

In other words: invocations of the JSHint executable that specify both files
that can be found *and* files that cannot be found. I think it's possible that
many active projects are configured along these lines: this kind of pattern
could accrue as developers rename files/directories. In such cases, they might
forget to remove references to the "old" files, and because JSHint is
"permissive," they may be unaware that they retain references to non-existent
files.

So while the accepted convention among Unix utilities is to fail *any* time a
non-existent file is specified, observing that convention may be too aggressive
for JSHint without a major release.

I'm still not sure why @caitp suggested making this behavior "opt-in" with a
dedicated flag, but doing so would address the issue I described above. I have
two objections to such an approach. First, many users will not discover it, so
to the extent that we recognize this behavior as "incorrect," a dedicated flag
will not benefit many users. Second, it commits this project to a flag that we
might call "doomed"--one that would be deprecated from the start, to be removed
once we could implement the conventional behavior in the next major release.

But there's still room for improvement, even within the 2.x release cycle. I
think JSHint should fail with a non-zero exit code if *all* files specified
cannot be found. Unlike above, I cannot think of a use case where users (1)
have mistakenly configured their projects in this way, or (2) *expect* the
executable to exit successfully.

The original report (gh-2884) described a case like this--a single non-existent
file. Could you update this patch to only fail when *zero total* files are
found? Also, could you add a test for this behavior? We can then make a
follow-up issue to observe the more conventional semantics JSHint's next major
release.

tanishalfelven:
Hello @jugglinmike! Thank you for the detailed report. I understand your concerns and am wondering about the implementation of a fix for this.

What you want is for jshint to exit only when *zero files are found*. Obviously, this could be implemented very simply - but what would  be the cleanest way to go about this? I would think that we don't want to bloat cli.js with more global variables.

I will work out a clean fix for this in the meantime, and will definitely add tests for this behavior. Thank you.

jugglinmike:
No problem at all; I'm glad you're still up for the challenge! Although in this case, I don't think it will be too bad. At first glance, it looks like [the length of the array returned by `gather` in the CLI's `run` method](https://github.com/jshint/jshint/blob/56515cf5444a06a0f02970536deaac84cf88227e/src/cli.js#L603) would be a reliable indicator of the condition we are concerned with. Try to push that assumption when writing the tests--are there any cases where that is *not* the right value to detect this error?

tanishalfelven:
@jugglinmike Alright, so the code is returning the correct error codes, BUT, it looks like a bunch of the unit tests are failing now due to a new error code being returned.

I've never written unit tests - although I'd love to learn how, and I understand the purpose. If you could point me in the right direction, I'd love to fix/add to them. I understand that all of the pertinent tests are held within tests/cli.js, but I'm not sure exactly where and how to add to that. If there is some sort of resource you could point me towards (and I'm gonna keep looking for resources also) I that'd be great as I continue work on this.

jugglinmike:
Those tests aren't written against non-existent files; they are written against
using the standard input stream *instead* of files. We'll definitely want that
behavior to continue to work as specified, so I think you may have to change
your approach.

Here's an idea: take a look at the `run` method. That signals success and
failure by invoking a callback with `true`/`false`. We should be able to
implement the functionality we've been discussing by altering the conditions
for returning there. Also: I agree with your point about global variable
"bloat." With this new approach, I think you'll be able to avoid defining a new
top-level variable.

As for tests, [the test named `testIgnores`](https://github.com/jshint/jshint/blob/93818f3f3b35cbdb0a66416db62bd03b4e02bf83/tests/cli.js#L612-L635) might be a good place to look for
inspiration. That test asserts an unrelated functionality, so you won't need to
replicate everything there (no need to specify the `--excludes` flag, for
instance), but it should give you an idea for how to express our expectations
for this new functionality. Include tests for a few combinations--just one
non-existent file, multiple non-existent files, and a mix of existent *and*
non-existent files.

The CLI tests are admittedly a little awkward to write. If you have any
questions about them or the implementation, don't hesitate to ask!

PR #2890, opened 2016-03-07
[[FIX]] `freeze` now errors when an ECMA6 object's prototype is edited.

### LINKED ISSUE #2859: 'freeze' doesn't protect ES6 native objects
```js
// jshint freeze: true, esversion: 6
Array.prototype.foo = 2;
Map.prototype.foo = 2;
``` 

Expected: two `Extending prototype of native object` warings.
Actual: only `Extending prototype of native object: 'Array'` at line 2

tanishalfelven:
Issue #2859 noted that when jshint `freeze: true` and `esversion: 6`, it did not give errors when extending the prototype of a ECMA6 object. This commit adds all of the natives to the list of natives in `jshint.js` so that these errors can be correctly reported.

Fixes #2859

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1284,10 +1284,11 @@ var JSHINT = (function() {
       "Array", "ArrayBuffer", "Boolean", "Collator", "DataView", "Date",
       "DateTimeFormat", "Error", "EvalError", "Float32Array", "Float64Array",
       "Function", "Infinity", "Intl", "Int16Array", "Int32Array", "Int8Array",
-      "Iterator", "Number", "NumberFormat", "Object", "RangeError",
-      "ReferenceError", "RegExp", "StopIteration", "String", "SyntaxError",
+      "Iterator", "Map", "Number", "NumberFormat", "Object", "Promise", "Proxy", "RangeError",
+      "Symbol", "Set",
+      "ReferenceError", "Reflect", "RegExp", "StopIteration", "String", "SyntaxError",
       "TypeError", "Uint16Array", "Uint32Array", "Uint8Array", "Uint8ClampedArray",
-      "URIError"
+      "URIError", "WeakMap", "WeakSet"
     ];
 
     function walkPrototype(obj) {


nicolo-ribaudo:
JSHint shouldn't warn if you are extending the prototype of an es6 object but `esversion` is 3 or 5.

jugglinmike:
That's right--we only want to warn about modification of ECMAScript 6 globals
in ECMAScript 6 environments. Please see [my suggestion on
gh-2859](https://github.com/jshint/jshint/issues/2859#issuecomment-184933597)
for notes on how to do this. Also, could you add some unit tests to demonstrate
this fix?

src/jshint.js
```diff
@@ -1284,10 +1284,11 @@ var JSHINT = (function() {
...
       "DateTimeFormat", "Error", "EvalError", "Float32Array", "Float64Array",
       "Function", "Infinity", "Intl", "Int16Array", "Int32Array", "Int8Array",
-      "Iterator", "Number", "NumberFormat", "Object", "RangeError",
-      "ReferenceError", "RegExp", "StopIteration", "String", "SyntaxError",
+      "Iterator", "Map", "Number", "NumberFormat", "Object", "Promise", "Proxy", "RangeError",
+      "Symbol", "Set",
+      "ReferenceError", "Reflect", "RegExp", "StopIteration", "String", "SyntaxError",
       "TypeError", "Uint16Array", "Uint32Array", "Uint8Array", "Uint8ClampedArray",
-      "URIError"
+      "URIError", "WeakMap", "WeakSet"
```
caitp: How about this:

```js
var _natives = null;
var _nativesVer = NaN;
var baseNatives = [
      "Array", "ArrayBuffer", "Boolean", "Collator", "DataView", "Date",
      "DateTimeFormat", "Error", "EvalError",
      "Function", "Infinity", "Intl",
      "Number", "NumberFormat", "Object", "RangeError",
      "ReferenceError", "RegExp", "StopIteration", /* what even is this? Intl thing? */ "String", "SyntaxError",
      "TypeError",
      "URIError"
];
var ecma2015Natives = [
  "Iterator", "Float32Array", "Float64Array", "Int16Array", "Int32Array", "Int8Array",
  "Uint16Array", "Uint32Array", "Uint8Array", "Uint8ClampedArray",
  "WeakMap", "WeakSet", "Symbol", "Map", "Set", "Proxy", "Promise",
];

function getNativesNames() {
  if (_nativesVer === state.option.esversion) return _natives;
  
  _nativesVer = state.option.esversion;
  _natives = _.assign([], baseNatives);
  if (_nativesVer > 6) _.assign(_natives, ecma2015Natives);
  return _natives;
}
```

this gives allows inline `esversion` directives to be honoured within the program, and lets us avoid checking modern builtins in legacy environments

Issue #2518, opened 2015-07-02
I003 isn't reported for the first `es5: true`


nicolo-ribaudo:
```js
// jshint es5: true
// jshint es5: true
// jshint es5: true
```

Result:
```
$ bin/jshint test.js
test.js: line 2, col 1, ES5 option is now set per default
test.js: line 3, col 1, ES5 option is now set per default

2 errors
```

Expected:
```
$ bin/jshint test.js
test.js: line 1, col 1, ES5 option is now set per default
test.js: line 2, col 1, ES5 option is now set per default
test.js: line 3, col 1, ES5 option is now set per default

3 errors
```

lukeapage:
and to clarify that means the info message is not given at all if you only set it once..
```
// jshint es5: true
```

caitp:
This is a stupid error anyways, best fix is to just get rid of it.

nicolo-ribaudo:
get rid :+1:

lukeapage:
No disagreement from me

shaunstanislaus:
how do you get rid of it?

jugglinmike:
We've resolved this by removing the "info" message entirely--see gh-2900.

Issue #2874, opened 2016-02-10
"ES5 option is now set per default" exits code 2


emaiax:
I'm getting "ES5 option is now set per default" when using `esversion: 5` on my `.jshintrc`. After that, `jshint` exits code 2.

Is that normal since it's an Info Message? I can't get rid of that message.

```
$ jshint --verbose test.js
test.js: line 0, col 0, ES5 option is now set per default (I003)

1 error
$ echo $?
2
$ jshint --version
jshint v2.9.1
```

jugglinmike:
This has been the case for a long while (at least going back to 2.5.0), but I can see no reason to keep the behavior. Wondering what @rwaldron thinks

wjdenny:
Having the same problem. Downgrading to 2.8.0 and commenting out the esversion: 5 option in .jshintrc seems to fix it enough for me to keep working.

jugglinmike:
I've opened https://github.com/jshint/jshint/pull/2900 to address this.

jugglinmike:
We've resolved this by removing the "info" message entirely--see gh-2900.

Issue #2766, opened 2015-11-16
I003 (ES5 option is now set per default) should not be displayed when esversion=5 overrides  esversion=6


vcarel:
I'm using the following project structure:
```
client/
        somefile.js
        .jshintrc
server/
        another_file.js
gulpfile.js
.jshintrc
```

I'm writing all my JS files in ES6 *except* my client sources which are in ES5.
So I defined a global .jshintrc file with the property (among others):
```
{
  ...
  esversion: 6
  ...
}
```

And in my client directory, I have another .jshintrc file which contains:
```
{
    "extends": "../.jshintrc",
    "esversion": 5,
}
```

But in all my client files I'm getting the warning message: ```I003 ES5 option is now set per default```
In this case, this message is wrong. My default is ES6 as per the global .jshintrc file and if I don't specify esversion=5, I will be in ES6 in my client app.

jugglinmike:
We've resolved this by removing the "info" message entirely--see gh-2900.

PR #2900, opened 2016-03-26
[[CHORE]] Remove I003

jugglinmike:
This warning is inappropriate in cases where `esversion` has been
globally set to a value other than `5` (e.g. in a project's `.jshintrc`
file) and a specific file overrides this value with an in-line
directive.

Although it is possible to detect this case while interpreting option
values, the warning itself has minimal utility. Some users may set
options to their default values in the interest of explicitness.
Preventing this practice (and doing so inconsistently--only for the
`esversion` option) adds little value.

Remove I003 completely.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -524,15 +524,6 @@ var JSHINT = (function() {
           return;
         }
 
-        /**
-         * TODO: Remove in JSHint 3
-         */
-        if (key === "es5") {
-          if (val === "true" && state.option.es5) {
-            warning("I003");
-          }
-        }
-
         if (key === "validthis") {
           // `validthis` is valid only within a function scope.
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
@rwaldron thoughts?

jugglinmike:
Hey looks who's back

jugglinmike:
This resolves gh-2518,  gh-2766, and gh-2874

rwaldron:
> Although it is possible to detect this case while interpreting option
values, the warning itself has minimal utility. Some users may set
options to their default values in the interest of explicitness.
Preventing this practice (and doing so inconsistently--only for the
esversion option) adds little value

This is compelling.

LGTM

jugglinmike:
Strange, the commit message for the commit in `master` has a whole bunch of extra white space. Bug in your tooling, @rwaldron?

PR #2921, opened 2016-05-02
Always resolve ignored paths

shanesmith:
Having issue with the inconsistent handling of the paths in .jshintignore between using the jshint cli and tools such as [mocha-jshint](https://github.com/ebdrup/mocha-jshint) where the paths aren't always resolved.

For example, a folder with a trailing slash in the .jshintignore, such as `node_modules/`, will successfully be ignored when using the jshint cli but will not with mocha-jshint.

This has been [previously found](https://github.com/jshint/jshint/pull/1837/files#diff-d9d9214113dea1c364abd4ba3c70be11L563) but the pull request (in which this issue was only a small part) was eventually abandoned. I've gone ahead and pulled the fix out into this quick patch in the hopes it goes through easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "phantomjs":                      "1.9.13",
     "regenerate":                     "1.2.x",
     "sinon":                          "1.12.x",
-    "unicode-6.3.0":                  "0.1.x"
+    "unicode-6.3.0":                  "0.1.x",
+    "recursive-readdir":              "2.1.x"
   },
 
   "license": "(MIT AND JSON)",
--- a/src/cli.js
+++ b/src/cli.js
@@ -571,10 +571,11 @@ var exports = {
       (!opts.extensions ? "" : "|" +
         opts.extensions.replace(/,/g, "|").replace(/[\. ]/g, "")) + ")$");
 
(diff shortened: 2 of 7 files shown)


jugglinmike:
Thanks for the patch! Would you mind adding a test? It's been a while since you
submitted this; if you're otherwise occupied, I can step in and add the test on
your behalf. Let me know!

shanesmith:
Sure, I'll give it a shot within the next few days.

shanesmith:
Unit tests added and commit message updated.... sorry about taking more than "a few days". ^^;;

jugglinmike:
No worries about the delay; I know how things get! The testing approach needs a
little work, though.

Instead of defining actual fixture files, lets take a cue from the existing
tests and define the new input via a CLI "stub." Beyond being more consistent,
this will save us from having to increase the complexity of the browser testing
infrastructure (where this test isn't relevant, anyway).

Because that kind of indirection could potentially lead to false positives,
lets also define a corresponding test--one where the "stub" files are *not*
ignored and the errors are reported as expected.

Finally, we should define the new test in addition to the existing test
(instead of replacing it).

Are you up for it, Shane?

shanesmith:
Alright, I'll take another stab at it.. just give me a few more days ;)

Issue #2777, opened 2015-11-21
New error: Literal values in right-hand side of `instanceof`


jugglinmike:
According to [ES5 11.8.6](https://es5.github.io/#x11.8.6):

> 11.8.6 The instanceof operator
> 
> The production RelationalExpression: RelationalExpression instanceof ShiftExpression is evaluated as follows:
> 
> 1. Let lref be the result of evaluating RelationalExpression.
> 2. Let lval be GetValue(lref).
> 3. Let rref be the result of evaluating ShiftExpression.
> 4. Let rval be GetValue(rref).
> 5. If Type(rval) is not Object, throw a TypeError exception.
> 6. If rval does not have a [[HasInstance]] internal method, throw a TypeError
>    exception.
> 7. Return the result of calling the [[HasInstance]] internal method of rval
>    with argument lval.

We can't completely enforce this statically, but there are some forms which we can reject
outright. The following `instanceof` applications should trigger an error:

```js
var x;
x instanceof 0;
x instanceof '';
x instanceof null;
x instanceof undefined;
x instanceof {};
x instanceof [];
x instanceof /./;
```

And it probably wouldn't hurt to warn on the following cases, too:

```js
x instanceof function() {};
x instanceof function MyUnusableFunction() {};
```

Issue #2773, opened 2015-11-19
RegExp after instanceof


nicolo-ribaudo:
A stupid example:
```js
var a = new Boolean() instanceof /a/.test("a").constructor;
```

JSHint doesn't like the `/` after `instanceof`, even if this is valid js.

jugglinmike:
Sounds good to me! See also gh-2777.

PR #2927, opened 2016-05-18
[[FEAT]] Error for literals on rhs of `instanceof`

ehjay:
This patch adds a new parse error that is used when a literal or function
declaration is found on the right-hand side of the `instanceof` infix operator.
Specifically, it checks for number, string, object, array and regular expression
literals, as well as the null and undefined keywords and function declarations.

The following generate errors:

    x instanceof 0;
    x instanceof '';
    x instanceof null;
    x instanceof undefined;
    x instanceof {};
    x instanceof [];
    x instanceof /./;
    x instanceof function() {};
    x instanceof function MyUnusableFunction() {};

But this is okay:

    /x/ instanceof /x/.constructor;

as per GH-2773

Closes GH-2777

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2141,14 +2141,13 @@ var JSHINT = (function() {
   infix("instanceof", function(left, token) {
     var right;
     var scope = state.funct["(scope)"];
-    var redefinedUndefined = scope.has("undefined");
     token.left = left;
     token.right = right = expression(120);
 
     if (right.id === "(number)" ||
         right.id === "(string)" ||
         right.value === "null" ||
-        (right.value === "undefined" && !redefinedUndefined) ||
+        (right.value === "undefined" && !scope.has("undefined")) ||
         right.arity === "unary" ||
         right.id === "{" ||
         (right.id === "[" && !right.right) ||
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2138,7 +2138,31 @@ var JSHINT = (function() {
   bitwise(">>", "shiftright", 120);
   bitwise(">>>", "shiftrightunsigned", 120);
   infix("in", "in", 120);
-  infix("instanceof", "instanceof", 120);
+  infix("instanceof", function(left, token) {
+    var right;
+    var scope = state.funct["(scope)"];
+    var redefinedUndefined = scope.has("undefined");
+    token.left = left;
+    token.right = right = expression(120);
+
+    if (right.id === "(number)" ||
+        right.id === "(string)" ||
+        right.value === "null" ||
+        (right.value === "undefined" && !redefinedUndefined) ||
+        right.arity === "unary" ||
+        right.id === "{" ||
+        (right.id === "[" && !right.right) ||
+        right.id === "(regexp)" ||
+        (right.id === "(template)" && !right.tag)) {
+      error("E060");
+    }
+
+    if (right.id === "function") {
+      warning("W139");
+    }
+
+    return token;
+  }, 120);
   infix("+", function(left, that) {
     var right;
     that.left = left;
--- a/src/messages.js
+++ b/src/messages.js
@@ -74,7 +74,8 @@ var errors = {
   E056: "'{a}' was used before it was declared, which is illegal for '{b}' variables.",
   E057: "Invalid meta property: '{a}.{b}'.",
   E058: "Missing semicolon.",
-  E059: "Incompatible values for the '{a}' and '{b}' linting options."
+  E059: "Incompatible values for the '{a}' and '{b}' linting options.",
+  E060: "Literals and unary operators cannot be used after the `instanceof` operator."
 };
 
 var warnings = {
@@ -219,7 +220,8 @@ var warnings = {
   W135: "{a} may not be supported by non-browser environments.",
   W136: "'{a}' must be in function scope.",
   W137: "Empty destructuring.",
-  W138: "Regular parameters should not come after default parameters."
+  W138: "Regular parameters should not come after default parameters.",
+  W139: "Function declarations should not be used after the `instanceof` operator."
 };
 
 var info = {
--- a/src/messages.js
+++ b/src/messages.js
@@ -75,7 +75,7 @@ var errors = {
   E057: "Invalid meta property: '{a}.{b}'.",
   E058: "Missing semicolon.",
   E059: "Incompatible values for the '{a}' and '{b}' linting options.",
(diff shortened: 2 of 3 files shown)


nicolo-ribaudo:
- Do these codes generate errors?

  ```js
  new String() instanceof "".constructor;
  ```
  ```js
  function X() {}
  new X() instanceof function() { return X; }();
  ```
- Are template literals reported?

  ```js
  x instanceof ``;
  x instanceof `${x}`;
  x instanceof tag``; // Should this generate an error? Note that a tag can return a function
  ```
- I think the check should happen in `"instanceof"`'s `led` method (then you just have to check `token.right` instead of the next token, to avoid false positives as the `"".constructor` one)
  To define the `led` method, just replace

  ```js
  infix("instanceof", "instanceof", 120);
  ```
  with

  ```js
  infix("instanceof", function (left, token) {
    token.left = left;
    token.right = expression(120);
  
    // Check token.right
  
    return token;
  }, 120);
  ```

 

*(Now I'm just thinking loud)*
- Should this be a waring or an error? `instanceof` followed by a function is *valid* javascript (even if JSHint should report *something* because it always\* returns `false`) and, if it is followed by an object literal, it can even return `true`:

  ```js
  0 instanceof { [Symbol.hasInstance](x) { return x < 3; } }; // This is a stupid thing anyway, so JSHint should report *something*
  ```

- It would be great if JSHint also reports unary operators, e.g.:

  ```js
  x instanceof +y;
  x instanceof ~y;
  // ...
  ```

ehjay:
Thanks for the feedback @nicolo-ribaudo! I made the changes you suggested:

* moved error checking to the `led` function
* added cases for template literals
* function declarations on the right-hand side now cause warnings instead of errors

jugglinmike:
Thanks for the patch; it's looking good so far! It's just a little too
aggressive--it produces warnings for code that, while ugly, is technically
valid:

```js
x instanceof [y][0];
x instanceof {}[constructor];
{
  let undefined = function() {};
  x instanceof undefined;
}
```

As for the tests: could you use "esverion: 6" and remove the grouping operator?

src/messages.js
```diff
@@ -74,7 +74,8 @@ var errors = {
   E056: "'{a}' was used before it was declared, which is illegal for '{b}' variables.",
   E057: "Invalid meta property: '{a}.{b}'.",
   E058: "Missing semicolon.",
-  E059: "Incompatible values for the '{a}' and '{b}' linting options."
+  E059: "Incompatible values for the '{a}' and '{b}' linting options.",
+  E060: "Literals and unary operators cannot be used after the `instanceof` operator."
```
nicolo-ribaudo: The other messages use `'` instead of `\`` (Just for consistency)

ehjay:
@jugglinmike most of that makes sense to me, but I'm still trying to understand the case where `undefined` is assigned to a function. I thought that was only possible in ES3 and no longer valid in ES5 and ES6.

jugglinmike:
Nope, that remains valid to this day. `undefined` is simply a global identifier. As such, it can be shadowed by other scopes. And because so much code on the web uses a defensive pattern like

```js
(function(undefined) {

}());
```

Changing this behavior would be problematic.

ehjay:
@jugglinmike that makes sense. I was trying to redefine the global `undefined` and I couldn't. But your example defines `undefined` within block scope.

Thanks! I'll try to get an updated pull request today

ehjay:
@jugglinmike from the `led` function, how can I get the value of `undefined` in the current `scope`? I'm trying to find an example of this in the codebase but haven't seen it yet.

ehjay:
@jugglinmike nevermind, I think I see it :smile:

ehjay:
Update the PR with the following features:

* adds three new valid test cases suggested by @jugglinmike 
* removes the grouping operators in the tests
* uses `esversion: 6`

src/jshint.js
```diff
@@ -2138,7 +2138,36 @@ var JSHINT = (function() {
   bitwise(">>", "shiftright", 120);
   bitwise(">>>", "shiftrightunsigned", 120);
   infix("in", "in", 120);
-  infix("instanceof", "instanceof", 120);
+  infix("instanceof", function(left, token) {
+    var right;
+    var redefinedUndefined = false;
+    var scope = state.funct["(scope)"] ? state.funct["(scope)"] : null;
```
nicolo-ribaudo: Isn't `state.funct.(scope)` always defined?

ehjay:
@nicolo-ribaudo yep, you're right. I cleaned it up it the latest push

src/messages.js
jugglinmike: We should say, "Non-callable values" to more precisely describe the error.
Also, let's say "as the second operand to instanceof".

src/messages.js
```diff
@@ -219,7 +220,8 @@ var warnings = {
   W135: "{a} may not be supported by non-browser environments.",
   W136: "'{a}' must be in function scope.",
   W137: "Empty destructuring.",
-  W138: "Regular parameters should not come after default parameters."
+  W138: "Regular parameters should not come after default parameters.",
+  W139: "Function declarations should not be used after the `instanceof` operator."
```
jugglinmike: It's actually, "Function expressions" that we are warning about. Also, let's
say "as the second operand to instanceof".

src/jshint.js
```diff
@@ -2138,7 +2138,31 @@ var JSHINT = (function() {
   bitwise(">>", "shiftright", 120);
   bitwise(">>>", "shiftrightunsigned", 120);
   infix("in", "in", 120);
-  infix("instanceof", "instanceof", 120);
+  infix("instanceof", function(left, token) {
+    var right;
+    var scope = state.funct["(scope)"];
+    var redefinedUndefined = scope.has("undefined");
```
jugglinmike: Considering how rarely this value will matter, this operation is a little
expensive. I think `scope.has("undefined")` is readable enough to warrant
invoking it in-line (without a named binding like `redefinedUndefined`) from
the conditional.

jugglinmike:
I've left a few more inline comments for you. Some of them apply to code I
reviewed earlier--sorry about that. It's tough to see everything on the first
pass.

Also: no need to squash your work as you go. I'll be happy to do that for you
once this is ready for `master`.

ehjay:
@jugglinmike thanks for the feedback, I made changes based on the in-line comments

jugglinmike:
Merged to `master` at commit e3e745be9408409b099a069959ad305955970ff0. Thanks, Andrew! This is great work.

PR #2928, opened 2016-05-18
[[CHORE]] Ignore vim temp files

ehjay:
Ignore files with .swp extension so that contributers using Vim can safely run

    git add --all

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ demo.js
 .idea
 
 coverage/
+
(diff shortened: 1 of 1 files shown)


ehjay:
I'm not sure why the appveyor build failed. Checking the log it looks like a connectivity issue?

fatal: unable to access 'https://github.com/jshint/jshint.git/': Failed to connect to github.com port 443: Timed out

PR #2930, opened 2016-05-24
Remove preferGlobal.  Fixes #2642

### LINKED ISSUE #2642: Remove preferGlobal
```
➜ npm install
npm WARN prefer global jshint@2.8.0 should be installed with -g

├─┬ grunt-contrib-jshint@0.11.2
│ └─┬ jshint@2.8.0
```

Could you remove [`"preferGlobal": true`](https://github.com/jshint/jshint/blob/master/package.json#L74)?
It's no any sense and bad practice as well.

asselin:
npm still emits a warning because of the preferGlobal flag.  i.e.: `npm WARN prefer global jshint@2.5.11 should be installed with -g`.  See for example npm/npm#8517 and npm/npm#11652.

An easy way to repro is `npm i swagger-js-codegen`, which includes jshint as a dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,8 +71,6 @@
 
   "license": "(MIT AND JSON)",
 
-  "preferGlobal": true,
-
   "files": [
     "bin",
     "data",


jugglinmike:
I am unable to reproduce the issue using npm version 3.3.12. In any event, it's
clear from the referenced discussions that the npm team considers this behavior
a bug in their project's code. Hopefully it gets resolved soon, but in the mean
time: is the warning having some effect on your project other than producing a
message during a successful installation process?

jugglinmike:
In the absence of additional information, I'm going to close this. We can re-open if anyone provides that info.

Issue #2932, opened 2016-06-02
Unclosed comment crashes JSHint


fix-fix:
JSHint version: 2.9.2 (I tested it on http://jshint.com/)
Environment: Chrome v51

When there is a open comment block above the property in object, JSHint crashes with error:
```
Uncaught TypeError: Cannot read property 'id' of null
e.nud @ jshint.js:6
C @ jshint.js:5
(anonymous function) @ jshint.js:6
C @ jshint.js:5
fe @ jshint.js:6
pe @ jshint.js:6
gt @ jshint.js:7
self.onmessage @ worker.js:7
```
I tested in on http://jshint.com/ so the message is a bit cryptic, but you can reproduce the bug with the following snippet
```js

({
  /*
  prop: 0
})

```

JSHint should give "Unrecoverable syntax error" error instead of crashing.

jugglinmike:
Thanks for the report! I just submitted a patch for this at gh-2937. Hopefully we'll have the fix out to you soon.

Arcanemagus:
Just so it's a little easier searching for this for anyone in the future, here is the full stack trace from running into this on the CLI:

```text
C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:5309
        throw err;
        ^

TypeError: Cannot read property 'id' of null
    at Object.x.nud (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:3228:26)
    at expression (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:884:34)
    at Object.fud (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:3666:17)
    at expression (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:881:32)
    at statement (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:1632:9)
    at statements (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:1677:16)
    at itself (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\jshint.js:5288:9)
    at lint (C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\cli.js:495:8)
    at C:\Temp\github\linter-jshint_GH292\node_modules\jshint\src\cli.js:656:7
    at Array.forEach (native)
```

jugglinmike:
[...]

PR #2937, opened 2016-06-04
[[FIX]] Avoid crash when peeking past end of prog

jugglinmike:
The lexer returns `null` values when tokens are requested after all
input has been consumed. Previously, the `peek` function would correctly
produce the special "(end)" token in situations where the parser
attempted to look beyond the end of the program. In such cases, however,
it would also pollute the lookahead buffer with an invalid entry--the
`null` value returned by the lexer. Future calls to `peek` could receive
this buffered value.

Because JSHint's internals are written with the assumption that `peek`
always returns a token object, the `null` value would trigger a
TypeError and subsequent program crash.

Re-factor the `peek` function to only insert valid token objects into
the lookahead buffer.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -716,16 +716,16 @@ var JSHINT = (function() {
     }
 
     while (j <= i) {
-      t = lookahead[j];
-      if (!t) {
-        t = lookahead[j] = lex.token();
-      }
-      j += 1;
-    }
+      t = lex.token();
 
-    // Peeking past the end of the program should produce the "(end)" token.
-    if (!t && state.tokens.next.id === "(end)") {
-      return state.tokens.next;
+      // Peeking past the end of the program should produce the "(end)" token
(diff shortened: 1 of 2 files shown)


jugglinmike:
This ought to resolve gh-2932.

@rwaldron and/or @caitp: would you mind merging this if it looks good to you?

src/jshint.js
```diff
@@ -716,16 +716,16 @@ var JSHINT = (function() {
     }
 
     while (j <= i) {
-      t = lookahead[j];
-      if (!t) {
-        t = lookahead[j] = lex.token();
+      t = lex.token();
+
+      // Peeking past the end of the program should produce the "(end)" token
+      // and should not extend the lookahead buffer.
```
rwaldron: Appreciate the explanation here

jugglinmike: :)

Issue #2923, opened 2016-05-05
jshint doesn't see SyntaxError about yield precedence


norlin:
jshint doesn't see SyntaxError about yield precedence
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

Example:
```javascript
'use strict';
/* jshint esnext:true */

function*x(){
    return 2 + yield 3;
}
```

```
/Users/norlin/Work/hola.org/test.js:5
    return 2 + yield 3;
               ^^^^^
SyntaxError: Unexpected strict mode reserved word
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:139:18)
    at node.js:968:3
```

jshint v2.9.2
nodejs v4.4.3

xderry:
ESLint detects this correctly: http://eslint.org/demo/

'use strict';
/* eslint-env es6 */

function*x(){
    return 2 + yield 3;
}

jugglinmike:
Thanks for the report! You're right--JSHint is currently very wrong about its
interpretation of the `yield` expression. I was hoping to have a patch for you
last weekend, but the fix is more involved than I anticipated. Still nothing to
show for my efforts yet, but I'm on the case!

PR #2951, opened 2016-06-16
[[FIX]] Correct behavior of singleGroups

jugglinmike:
The extension to `nud` is a fairly low-level change for such a fringe feature
as "singleGroups", but I believe it is technically necessary to expose correct
behavior. It doesn't seem too dirty--the change is functional and does not rely
on additional global state.

And believe it or not, this is actually blocking my patch for gh-2923 (`yield`'s
binding power is a little tricky).

I'm curious to hear what others think! Especially @caitp and @rwaldron :)

---

The grouping operator may be necessary if the right-binding power of the
operator preceeding the "(" token is greater than the left-binding power
of the operator immediately following it.

The initial implementation approximated this behavior by using the
readily-accessible left-binding power of both operators. This heuristic
is not always accurate, however as in the following case:

    typeof (a + b);

Here, the necessity of the grouping operator is determined by the
right-binding power of the `typeof` operator.

Extend the Pratt parser to make the right-binding power available to all
null denotation ("nud") functions. Remove an invalid unit test exposed
by this improvement.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2929,12 +2929,17 @@ var JSHINT = (function() {
     return funct["(global)"] && !funct["(verb)"];
   }
 
-  function doTemplateLiteral(left) {
+  /**
+   * This function is used as both a null-denotation method *and* a
+   * left-denotation method, meaning the first parameter is overloaded.
+   */
+  function doTemplateLiteral(leftOrRbp) {
     // ASSERT: this.type === "(template)"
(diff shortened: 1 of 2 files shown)


jugglinmike:
@rwaldron Mind reviewing and merging this?

rwaldron:
Nice work—I can't see any immediately obvious issues with this and all tests are green: LGTM.

PR #2965, opened 2016-06-23
allow custom devel's functions

comdeng:
Sometimes we need `alert` and don't need `console`.


nicolo-ribaudo:
Why just don't use the `global` option?

PR #2977, opened 2016-07-09
[[FIX]] Correct interpretation of ASI

jugglinmike:
@rwaldron this is the latest installment of the ongoing saga to resolve gh-2923. It's the last one, though--I have already written the follow-up patch that will finish this up.

Commit message:

> Ensure JSHint observes correct automatic semicolon insertion semantics.
> This change also modifies parsing logic for the Mozilla "JavaScript 1.7"
> implementation of YieldExpression. There is no formal specificaion for
> that version of the language feature, so the validity of this change was
> experimentally confirmed.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -815,7 +815,8 @@ var JSHINT = (function() {
     if (next.id === ";" || next.id === "}" || next.id === ":") {
       return true;
     }
-    if (isInfix(next) === isInfix(curr) || (curr.id === "yield" && state.inMoz())) {
+    if (isInfix(next) === isInfix(curr) || curr.ltBoundary === "after" ||
+      next.ltBoundary === "before") {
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -815,7 +815,8 @@ var JSHINT = (function() {
     if (next.id === ";" || next.id === "}" || next.id === ":") {
       return true;
     }
-    if (isInfix(next) === isInfix(curr) || (curr.id === "yield" && state.inMoz())) {
+    if (isInfix(next) === isInfix(curr) || curr.ltBoundary === "after" ||
+      next.ltBoundary === "before") {
```
rwaldron: This got a little weird for readability, but otherwise not a blocking issue

jugglinmike:
Thanks, Rick!

PR #2978, opened 2016-07-10
Correct parsing of YieldExpression

jugglinmike:
At long last, here is a fix for gh-2923! You'll note the very first thing this
patch does is factor out the Mozilla-JavaScript-1.7-specific logic into a
standalone function. I think we 'd all like to remove `moz` one day, but until
then, this seems like the safest/most readable way to change the ECMAScript
standard version. (Since we we have no means to verify conformance with the
Mozilla-specific extension and no bug reports for that implementation, I've
elected to maintain behavior there.)

@rwaldron Would you mind reviewing this for me?

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -806,7 +806,11 @@ var JSHINT = (function() {
   }
 
   function isInfix(token) {
-    return token.infix || (!token.identifier && !token.template && !!token.led);
+    return token.infix ||
+      (!token.identifier && !token.template && !!token.led) ||
+      // Although implemented as an Identifier, the `yield` keyword behaves as
+      // an operator when it appears in the body of a generator function.
+      (token.id === "yield" && !!state.funct["(generator)"]);
   }
 
   function isEndOfExpr() {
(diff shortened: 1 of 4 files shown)


jugglinmike:
@rwaldron Bugging you again because you mentioned that you didn't see this one.

rwaldron:
Are you ok with the coveralls fail?

jugglinmike:
I was okay with the decrease in test coverage, but that's because I was being
lazy. When I went to address it, I found I could simplify the new `mozYield`
function slightly--when we're taking that branch, it's impossible for
`state.inES6()` to return `false`, so one branch can removed.

Thanks for keeping me honest! I'm happy to squash and force-push if you'd
like--let me know.

AttilaSATAN:
When this going to be added on releases?

jugglinmike:
@rwaldron Can you take another look?

rwaldron:
Sorry for the delay in response :\

This looks good. I like that my tangentially related question about coverage lead to removal and simplification of a code path—that's rare!

rwaldron:
LGTM!

Issue #2975, opened 2016-07-07
Module Declarations allowed in invalid positions


jugglinmike:
```js
// jshint esversion: 6, module: true
{
  import x from './m.js';
}
```

Expected: error (ImportDeclaration is not allowed outside of a ModuleItem)
Actual: no warning

sstern6:
@jugglinmike I am very interested in creating a PR for this bug.  I am brand new to the code base and was wondering if you could point me in the direction of a folder or file that would be a good start?

Thanks

jugglinmike:
Sure thing! Here are the relevant parts of the codebase:

- [Logic for `import` declarations](https://github.com/jshint/jshint/blob/3ef7a0395cea5ec963a51f69b55a404f3f8e4578/src/jshint.js#L4500)
- [Logic for `export` declarations](https://github.com/jshint/jshint/blob/3ef7a0395cea5ec963a51f69b55a404f3f8e4578/src/jshint.js#L4588)
- [Utility method to detect the global scope](https://github.com/jshint/jshint/blob/3ef7a0395cea5ec963a51f69b55a404f3f8e4578/src/scope-manager.js#L771-L777)

If you have any more questions, don't hesitate to ask.

sstern6:
Thank you!  Ill get cranking and will probably have questions

sstern6:
@jugglinmike having some issues setting up local env.  I followed the guides, ran `npm install`, then `bin/build`. finally `npm test` and I am getting 1 error from the cli.js.  Attaches is the output, any idea why this could be happeneing? I am on a mac FYI.

```
> jshint@2.9.2 pretest /Users/scottstern/Desktop/JSHINT/jshint
> jshint src && jscs src

No code style errors found.

> jshint@2.9.2 test /Users/scottstern/Desktop/JSHINT/jshint
> npm run test-node && npm run test-browser


> jshint@2.9.2 test-node /Users/scottstern/Desktop/JSHINT/jshint
> npm run test-unit && npm run test-cli && npm run test-regression


> jshint@2.9.2 test-unit /Users/scottstern/Desktop/JSHINT/jshint
> nodeunit tests/unit


core
✔ testCustomGlobals
✔ testUnusedDefinedGlobals
✔ testImplieds
✔ testExportedDefinedGlobals
✔ testGlobalVarDeclarations
✔ globalDeclarations
✔ multilineGlobalDeclarations
✔ testNewArray
✔ testNewNonNativeArray
✔ testNonNativeArray
[...]

PR #2982, opened 2016-07-18
[[FIX]] Only allow Import statemnets in ES6 to be at top level of scope

sstern6:
Added a conditional within stmt("import" that checks if the scope of the import statement is not global using the isGlobal function call.  If it is not global I created an error message in messages.js named E061 that lets the user know that their import statement needs to be at the top level of the module scope.

This issue fixes https://github.com/jshint/jshint/issues/2975.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4478,7 +4478,7 @@ var JSHINT = (function() {
 
   stmt("import", function() {
     if (!state.funct["(scope)"].block.isGlobal()) {
-      error("E053", state.tokens.curr, state.tokens.curr.value);
+      error("E053", state.tokens.curr, "Import");
     }
 
     if (!state.inES6()) {
@@ -4579,7 +4579,7 @@ var JSHINT = (function() {
     }
 
     if (!state.funct["(scope)"].block.isGlobal()) {
-      error("E053", state.tokens.curr, state.tokens.curr.value);
+      error("E053", state.tokens.curr, "Export");
       ok = false;
     }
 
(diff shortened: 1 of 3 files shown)


jugglinmike:
First off, thanks for the patch! There are a few things I think we could
improve before we merge this to `master`.

I'd like to avoid creating new error codes where possible, so do you think you
could re-use `E053` for this? You'll need to parameterize the message first.
You can use `E048` as a guide:

- Here's how a parameterized error message is defined:
  https://github.com/jshint/jshint/blob/039ee2e7f26b5d4b6f503d21b66664387185d940/src/messages.js#L66
- Here's how a parameterized error message is used:
  https://github.com/jshint/jshint/blob/039ee2e7f26b5d4b6f503d21b66664387185d940/src/jshint.js#L3513

I prefer the content of your error message to what we currently have in `E053` since it does not conflate module scope and global scope. The message content is not public API, so it's safe to update `E053` with your text (plus the parameterization).

Also, could you also add a few more declarations to your tests? I have four in mind: importing a namespace, importing a binding (with and without `as`), and importing just using a module specifier (e.g. `import 'm.js';`).



Sound good?

sstern6:
Perfect, thanks @jugglinmike for the feedback, will update and sign the CLA the when I repush!

sstern6:
@jugglinmike pushed code with your recommendations and comments above, let me know if there is anything else that should be done before merging.  

Thank you

jugglinmike:
Squashed and merged to `master` at commit d800e44715e85803602667d8f353f83b0e8f7f7b. Thanks for your help!

PR #2988, opened 2016-07-26
Improve the license information in package.json

reiz:
I separated the license information into an Array. That way it's easier to handle for [VersionEye](https://www.versioneye.com) and other tools which are dealing with license information.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "unicode-6.3.0":                  "0.1.x"
   },
 
-  "license": "(MIT AND JSON)",
+  "licenses": ["MIT", "JSON"],
 
   "preferGlobal": true,
 


reiz:
I just accepted the CLA at CLAHub.

jugglinmike:
The current value was implemented in gh-2420 to conform to the SPDX syntax. [This is the format that is explicitly endorsed by npm](https://docs.npmjs.com/files/package.json):

> ### license
>
> You should specify a license for your package so that people know how they
> are permitted to use it, and any restrictions you're placing on it.
>
> If you're using a common license such as BSD-2-Clause or MIT, add a current
> SPDX license identifier for the license you're using, like this:
>
>     { "license" : "BSD-3-Clause" }

...and a bit later, it discourages the format used in this patch:

> Some old packages used license objects or a "licenses" property containing an
> array of license objects:
>
>     // Not valid metadata
>     { "license" :
>       { "type" : "ISC"
>       , "url" : "http://opensource.org/licenses/ISC"
>       }
>     }
>
>     // Not valid metadata
>     { "licenses" :
>       [
>         { "type": "MIT"
>         , "url": "http://www.opensource.org/licenses/mit-license.php"
>         }
>       , { "type": "Apache-2.0"
>         , "url": "http://opensource.org/licenses/apache2.0.php"
>         }
>       ]
>     }
>
> Those styles are now deprecated. Instead, use SPDX expressions, like this:
>
>     { "license": "ISC" }
>
>     { "license": "(MIT OR Apache-2.0)" }

The best solution in this case may be for VersionEye to improve its support for the SPDX syntax.

reiz:
OK. Fair enough.

PR #2992, opened 2016-07-30
[[CHORE]] Refactor directive parsing logic

jugglinmike:
Extend the recently-improved `isEndOfExpr` function so that it may be
re-used during directive parsing, limiting duplication.

---

@nicolo-ribaudo this refactors some of your work from gh-2716 to re-use the
recent general improvement to ASI interpretation from gh-2977. What do you
think?


rwaldron:
Nice work here

PR #2999, opened 2016-08-03
[[FIX]] Support `y` RegExp flag in ES2015 code

jugglinmike:


src/lex.js
```diff
@@ -1394,10 +1394,24 @@ Lexer.prototype = {
...
+        if (!state.inES6(true)) {
+          this.trigger("warning", {
+            code: "W119",
+            line: this.line,
+            character: this.char,
+            data: [ "Sticky RegExp flag", "6" ]
+          });
+        }
+        if (value.indexOf("y") > -1) {
+          malformedDesc = "Duplicate RegExp flag";
```
rwaldron: Is this necessary? Duplicate flags produce syntax errors

jugglinmike: This is a bit tricky. We're maintaining two variables for the flags: a string
named `value` and an array named `flags`. The `flags` array is eventually used
to create a RegExp instance in order to off-load the difficult work of RegExp
validation to whatever engine happens to be running JSHint itself.

If we include the `y` flag in this array, then running JSHint with `esversion:
6` would produce different results according to whether the current engine
supported "sticky" RegExps. So instead, this patch "skips" the `y` flag when it
is encountered.

This could be an invalid transformation for other flags. For example, `/{/` is
valid JavaScript but `/{/u` is not. Indiscriminately removing `u` would
therefor effect parsing correctness.

As far as I know, the only time the `y` flag could impact parsing is if it were
duplicated. (Runtime behavior is another matter, of course.) So this patch
explicitly verifies that the flag is *not* duplicated (using the separate
`value` string mentioned earlier) and otherwise parses `y`-containing RegExps
as though they did not have the `y` flag at all.

jugglinmike:
@rwaldron This should be good to go.

rwaldron:
Thanks for the explanation!

LGTM

Issue #3002, opened 2016-08-06
Syntax error when a function expression with a nonempty body is a default parameter in an arrow function


McBrainy:
I'm using JSHint version 2.9.2. The following code is valid es6 that compiles fine with Babel:

```JavaScript
// jshint esversion: 6
const foo = (x = function () { return; }) => x;
```

But JSHint gives this output:

```
jshint-error.js: line 2, col 46, Expected '(' and instead saw 'x'.
jshint-error.js: line 2, col 47, Expected an identifier and instead saw ';'.
jshint-error.js: line 3, col 1, Unrecoverable syntax error. (100% scanned).

3 errors
```

Notably, there are no errors if the inner function body is empty:
```JavaScript
const foo = (x = function () {}) => x;
```
Nor if the outer function is a function declaration:
```JavaScript
function foo(x = function () {}) {}
```
Nor even if the outer function is a function expression using the old syntax:
```JavaScript
const foo = function(x = function () {}) {};
```

jugglinmike:
Thanks for the report, Brian! I've submitted a fix at gh-3003.

In case you're curious: the `return` statement is a bit of a red herring. The
actual source of the bug is the semicolon token. Weird!

McBrainy:
Wow, that was quick. In case you're wondering what kind of a crazy use case I could have for this, it's [here](https://gist.github.com/McBrainy/9affcb2a0c991613806dda1e9fee45ab), specifically the factorial example, where I first wrote `factAcc` as a long-form function before I realized that recursive arrow functions are possible in default params.

PR #3003, opened 2016-08-07
[[FIX]] Support semicolons within arrow fn params

jugglinmike:
When JSHint encounters an "open parenthesis" token, it must determine if
the token marks the beginning of an arrow function or a grouping
operator before parsing can proceed.

To make this determination, it inspects the token stream to find the
corresponding "close parenthesis" character.

Previously, this lookahead behavior halted whenever a semicolon token
was encountered. This was presumably an optimization to allow the
algorithm to exit early in the event of an unmatched "open parenthesis"
token.

Because a semicolon token may appear in the parameter list of an arrow
function (e.g. within a default parameter value that is a function
expression), the presence of such a token does *not* necessarily
indicate that the "open parenthesis" is unmatched.

Remove the condition concerning the semicolon character from the
lookahead algorithm.

Resolves gh-3002.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2498,7 +2498,7 @@ var JSHINT = (function() {
       i += 1;
       pn1 = pn;
       pn = peek(i);
-    } while (!(parens === 0 && pn1.value === ")") && pn.value !== ";" && pn.type !== "(end)");
+    } while (!(parens === 0 && pn1.value === ")") && pn.type !== "(end)");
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
@rwaldron Mind taking a look?

rwaldron:
My only issue with this is the commit message's short description, and only because my first read of it was "alarming". I'm not going to block on it, but I would've used the word "nested" somewhere in there ;)

PR #3010, opened 2016-08-16
[[CHORE]] Update dependency: cli

jugglinmike:
The `cli` module at this version range was recently found to have a
security vulnerability in a feature that is not used in JSHint. Update
to the latest version which resolves the vulnerability by removing the
affected functionality.


rwaldron:
Nice!

PR #3011, opened 2016-08-16
[[FIX]] Allow RegExp literal as `yield` operand

jugglinmike:
Extend the lexer to recognize RegExp literals which immediately follow
the `yield` operator.


rwaldron:
Nice work!

Issue #3013, opened 2016-08-19
Disabling warning for W100 not working in 2.9.3


damyanpetev:
See https://github.com/IgniteUI/ignite-ui/pull/243:

First build passed with 2.9.2, consecutive builds with 2.9.3 update fail, with a line wrapped with /*jshint -W100 */

Don't see anything related in the [update blog](http://jshint.com/blog/2016-08-18/release-2-9-3/), but has something changed in the way those are handled?

jugglinmike:
Thanks for the report! This is indeed a regression.

Not that it helps you much, but it looks like the new release has surfaced a previously-existing bug--in the process of reducing the input you shared, I was able to experience this problem in version 2.9.2, as well.

I need to step away for a bit, but I hope to have a patch ready this evening (EST).

jugglinmike:
Here's the story:

Upon encountering an open parenthesis character (`(`) in an expression
position, a ES2015-aware JavaScript parser like JSHint's has to determine
whether it marks the beginning of a grouping operator or an arrow function.

JSHint does this by tokenizing all the input that follows until it finds a
"matching" close parenthesis character (`)`). At that point, it can make a
determination about the nature of the production in question--if it's followed
by the `=>` token, it's an arrow function; otherwise, it's a grouping operator.

While it is looking ahead, it does not apply in-line directives it encounters
such as `-W100`. However, the lexer is responsible for emitting certain linting
errors *including* `W100`. So as soon as JSHint begins "looking ahead", its
behavior in response to these potentially-dangerous characters is "locked in."
(That suggests a short-term fix for anyone suffering from this regression: disable
`W100` before the IIFE. Not ideal, I know, but note that [the relevance of that
warning is disputed](https://github.com/jshint/jshint/issues/2741).)

This regression came about because of a seemingly-unrelated patch: gh-3003.
Previously, the lookahead would be cancelled if a semicolon token were
encountered, as in:

```js
(function() {
;//<-- lookahead ends at this semicolon token
}());
```

...however, that was found to be an invalid heuristic because arrow function
parameters may themselves contain that token:

```js
(x = function() {
;//<-- lookahead should *not* end at this semicolon token
}) => x;
```

So we removed it in order to properly recognize arrow functions in those cases.

With that in mind, the fully-reduced test case may be a little more clear (that
is an unprintable `\u200f` character in the string literal):

```js
(function() {
	;

	/*jshint -W100 */
	"‏";
})();
```

[...]

Issue #2741, opened 2015-10-31
W100 is signaled in Node.js sources


acarstoiu:
This warning
`"This character may get silently deleted by one or more browsers."`
is reported by JSHint even if there is a `"node": true` in the configuration file. Can be obtained with odd Unicode characters (written directly, UTF-8) on the current _**master**_ version (post _2.9.0_).

jugglinmike:
Thanks for the report! I'm unable to find the regression point based on your description. Could you share a small example that demonstrates the warning and the version of JSHint that does *not* warn for that input?

acarstoiu:
Sorry, I do not read regularly my emails, if you can believe that.
The fact that this issue is a regression was my assumption given [another Node.js-related problem](https://github.com/jshint/jshint/issues/2657) discovered shortly after the launch of 2.9.0. To avoid a misunderstanding, I removed this detail.

But the problem is *perfectly reproduceable* :bug:
Just use this `.jshintrc`
```js
{
	"node": true,
	"latedef": "nofunc",
	"laxbreak": true,
	"sub": true,
	"funcscope": true
}
```
on [a file as this](http://pastebin.com/V93fFb6s) (remove the JSHint comment to see the bug, of course).

jugglinmike:
I've started on a fix for this, but in all honesty, I don't understand W100 to begin with. [Warning W100 reads "This character may get silently deleted by one or more browsers."](https://github.com/jshint/jshint/blob/29dd071d66ee16d250ce63c0c4e061b74f7423ac/src/messages.js#L184) It's emitted when a character satisfies [the follow regular expression](https://github.com/jshint/jshint/blob/b554ffe61eabf6e3e3a2876a3a377271da308811/src/reg.js#L13-L15):

      /[\u0000-\u001f\u007f-\u009f\u00ad\u0600-\u0604\u070f\u17b4\u17b5\u200c-\u200f\u2028-\u202f\u2060-\u206f\ufeff\ufff0-\uffff]/

One fix that I am considering is to limit emitting this warning to when JSHint's `browser` setting has been enabled. [This behavior has existed from the very first commit of JSLint](https://github.com/jshint/jshint/blob/ca120a731db548c0014320fa0c196edc613536ae/fulljslint.js#L895), and like many warnings from that era, the motivation is not clear to me. I don't actually know which browsers would exhibit this behavior, and I can't find justification online. So another fix would be to just remove the warning entirely. 

[...]

PR #3016, opened 2016-08-20
Fix for gh-3013

jugglinmike:
@rwaldron This is a fix for gh-3013 in two parts. You can see my response to
that issue for more detail about the problem. Please review carefully because
I'm messing with some relatively fundamental behavior here.

One specific question I have for you: I do not think it's possible to test for
this behavior at every call site of `nextLine`. (The situation we're interested
in here is when the lexer is tokenizing input and may pass over a `-W100`
directive.) I've still updated the call sites for consistency, and in case some
future change *can* use `checks` from different call sites. Does that seem
right to you?


jugglinmike:
@rwaldron Also note that a much simpler fix would be to remove `W100` altogether--I still don't know exactly what that warning is for. Check out gh-2741 for an unresolved discussion along these lines.

jugglinmike:
This needs work--W101 is also suffering the same problem.

jugglinmike:
@rwaldron Thanks to gh-3017, I have a better understanding of the scope of this problem.
The bug concerns all warnings that are detected by the lexer, namely W048,
W049, W100, W101, W113, and W119. An unrelated bug fix that was released with
version 2.9.3 exposed the problem to a wider category of input. (I suppose that
still qualifies as a "regression," but we may be splitting hairs there.)

The context that triggers this error is the lookahead operations JSHint
performs to disambiguate between multiple valid parsing strategies:

- the `{` token in an expression position (either an object literal or a
  destructuring assignment)
- the `[` token (either an array literal or a destructuring assignment)
- the `(` token (either a grouping operator or an arrow function parameter
  list)

This latest version of the patch addresses all lexer-triggered warnings in all
relevant contexts. In the interest of transparency, I've pushed up the new
version of the patch as a `fixup!` commit. Before this lands, I'd like to
squash with a new commit message that correctly describes the expanded scope of
the fix.

> [[FIX]] Honor "ignore" directives during lookahead
>
> When the lexer emits a token, it may not be aware of the complete set of
> in-line linting directives because these are activated by the parser. This
> is notably the case during parsing lookahead operations (e.g. when selecting
> between multiple parsing strategies that may follow the `{`, `[`, and `(`
> tokens).
>
> In these contexts, the lexer should not immediately report token-related
> warnings. Reporting must be deferred until the parser has visited all preceding
> comments and fully applied all linting directives present in the source
> code.

I've also included the above text as a comment within the test's fixture file.
That file necessarily contains a few instances of non-printable characters,
which has the unfortunate side-effect of suppressing its output in patch
summaries from git and GitHub.com.

@damyanpetev @batista would either of you mind testing this branch of JSHint
against your code base and verifying that it resolves your issue?

batista:
Sure, I'm unable to do it now, but I'll test it and get back to you
tomorrow.

On Sep 10, 2016 21:03, "jugglinmike" <notifications@github.com> wrote:

@rwaldron <https://github.com/rwaldron> Thanks to gh-3017
<https://github.com/jshint/jshint/issues/3017>, I have a better
understanding of the scope of this problem.
The bug concerns all warnings that are detected by the lexer, namely W048,
W049, W100, W101, W113, and W119. An unrelated bug fix that was released
with
version 2.9.3 exposed the problem to a wider category of input. (I suppose
that
still qualifies as a "regression," but we may be splitting hairs there.)

The context that triggers this error is the lookahead operations JSHint
performs to disambiguate between multiple valid parsing strategies:

   - the { token in an expression position (either an object literal or a
   destructuring assignment)
   - the [ token (either an array literal or a destructuring assignment)
   - the ( token (either a grouping operator or an arrow function parameter
   list)

This latest version of the patch addresses all lexer-triggered warnings in
all
relevant contexts. In the interest of transparency, I've pushed up the new
version of the patch as a fixup! commit. Before this lands, I'd like to
squash with a new commit message that correctly describes the expanded
scope of
the fix.

[[FIX]] Honor "ignore" directives during lookahead

When the lexer emits a token, it may not be aware of the complete set of
in-line linting directives because these are activated by the parser. This
is notably the case during parsing lookahead operations (e.g. when selecting
between multiple parsing strategies that may follow the {, [, and (
tokens).

In these contexts, the lexer should not immediately report token-related
warnings. Reporting must be deferred until the parser has visited all
preceding
comments and fully applied all linting directives present in the source
code.

I've also included the above text as a comment within the test's fixture
file.
That file necessarily contains a few instances of non-printable characters,
which has the unfortunate side-effect of suppressing its output in patch
summaries from git and GitHub.com.

@damyanpetev <https://github.com/damyanpetev> @batista
<https://github.com/batista> would either of you mind testing this branch
of JSHint
against your code base and verifying that it resolves your issue?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jshint/jshint/pull/3016#issuecomment-246126241>, or mute
the thread

jugglinmike:
Thanks!

rwaldron:
I can review tomorrow—please nudge me if I haven't reported by noon

batista:
@jugglinmike can confirm that 72209ba95dde23bb2b19912a36c713f5e475d9e6 fixes #3017 for me

jugglinmike:
Thanks, Sérgio!

rwaldron:
@jugglinmike I'm confused about the relationship to https://github.com/jshint/jshint/issues/2741

jugglinmike:
@rwaldron I made that comment when I believed that the bug reported in gh-3013 was limited to W100. From that perspective, another viable fix was to remove W100 altogether--I suggested that as a possibility in gh-2741.

While I still don't understand the intent of W100, it's clear that the underlying problem is more far-reaching, so I'm happy to let the conversation about removing W100 proceed separately.

damyanpetev:
@jugglinmike A bit late, but I can also verify 72209ba95dde23bb2b19912a36c713f5e475d9e6 honored the ignore for W100 warning with as initially logged in #3013 (massive IIFE). Also tested with this bit:
```js
(function() {
	var test;
    /*jshint -W100 */
    test = "‏س";
})();
```
And it worked out okay as well.

rwaldron:
Thanks @damyanpetev and @batista, you're testing and confirmation is helpful here

rwaldron:
Create work here, thanks @jugglinmike

PR #3019, opened 2016-08-28
Replace Private Use Area character

### LINKED ISSUE #3018: A comment contains a Private Use Area character
(originally added by by commit 87378cca3c98e1ff4d5be11f6c16a9f24a212e3e)

https://github.com/jshint/jshint/blob/master/src/jshint.js#L4664

> ```
>      //      export default [lookahead  { function, class }] AssignmentExpression[In] ;
> ```

After the word `lookahead`.

Characters from the Private Use Area don't have any inherent meaning or appearance associated with them.

In the [Symbol font](https://en.wikipedia.org/wiki/Symbol_(typeface)), U+F0CF is ∉ (U+2209, 'not an element of'). Maybe this is the intended character?

LMLB:
fixes #3018

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4661,7 +4661,7 @@ var JSHINT = (function() {
 
     if (state.tokens.next.type === "default") {
       // ExportDeclaration ::
-      //      export default [lookahead  { function, class }] AssignmentExpression[In] ;
+      //      export default [lookahead ∉ { function, class }] AssignmentExpression[In] ;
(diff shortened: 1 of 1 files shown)


jugglinmike:
Merged to `master` at commit 8accbaed73ebaf8f2baab2ee3132bc83428b0a8c (I added `[[DOCS]]` to the commit message as per our guidelines.) Thanks!

PR #3023, opened 2016-08-30
[[FEAT]] Remove fdescribe and fit from Jasmine's globals

### LINKED ISSUE #3022: Remove global definitions for Jasmine focused specs
Jasmine's focused spec globals (`fdescribe`, `fit`) were added when updated with the Jasmine 2.1 globals:
https://github.com/mrjoelkemp/jshint/commit/343c45e990e7f6b93a3374adb3d077894960af01

This isn't consistent with the Jasmine 1.3 globals which do not include them (`ddescribe`, `iit`).

sgreen:
Remove Jasmine 2.1 focused spec globals to be consistent with the Jasmine 1.3 globals

Fixes #3022

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -730,7 +730,5 @@ exports.jasmine = {
   beforeAll   : false,
   afterAll    : false,
   fail        : false,
-  fdescribe   : false,
-  fit         : false,
   pending     : false
 };


jugglinmike:
The current implementation isn't optimal, but this change is backwards-incompatible, so we can't accept it. I've written up some thoughts on a more robust solution in your original bug report at gh-3022.

PR #3030, opened 2016-09-10
Rephrase message for W014

jugglinmike:
W014 was previously implemented with the following message:

> Bad line breaking before '{ token }'.

This message does little to explain the rationale for the warning or the
means to correct it. Re-phrase the message to better reflect the
intention.

### DIFF
--- a/src/messages.js
+++ b/src/messages.js
@@ -93,7 +93,7 @@ var warnings = {
   W011: null,
   W012: null,
   W013: null,
-  W014: "Bad line breaking before '{a}'.",
+  W014: "Misleading line break before '{a}'; readers may interpret this as an expression boundary.",
   W015: null,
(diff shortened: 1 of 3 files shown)


jugglinmike:
@rwaldron What do you think?

rwaldron:
I can review tomorrow—please nudge me if I haven't reported by noon

src/messages.js
```diff
@@ -93,7 +93,7 @@ var warnings = {
   W011: null,
   W012: null,
   W013: null,
-  W014: "Bad line breaking before '{a}'.",
+  W014: "Misleading line break before '{a}'; readers may interpret this as an expression boundary.",
```
rwaldron: So much clearer.

Issue #3037, opened 2016-09-28
weird behaviour (yield related)


alinalexa:
![Image](https://ibin.co/w800/2wcvWbKdlS3S.png)
vs 
![Image](https://ibin.co/2wcwGUt4C6Yg.png)
vs
![Image](https://ibin.co/2wcxavGvzHv5.png)

`function a() {}` can be anything

`.jshintrc`
```
{
    "maxerr": 10,
    "curly": false,
    "eqeqeq": true,
    "latedef": true,
    "undef": true,
    "unused": "last",
    "maxlen": false,
    "varstmt": true,
    "asi": true,
    "esversion": 6,
    "lastsemic": true,
    "noyield": true,
    "node": true
}
```

jugglinmike:
Confirmed. Here's a furher reduced test case:

```js
// jshint esversion: 6, asi: true
function* g() {
  void 0
  yield;
}
```

Thanks for the report

jugglinmike:
This bug extends beyond the `yield` operator--it can also be reproduced using the increment/decrement operators, as in:

```js
// jshint asi: true
x++
void 0
```

jugglinmike:
I've submitted a fix here: https://github.com/jshint/jshint/pull/3045. This was much trickier than it looked!

jugglinmike:
Resolved via gh-3045. You can expect this fix to be available in the next release of JSHint (post 2.9.3).

PR #3045, opened 2016-10-17
[[FIX]] Correct interpretation of ASI

jugglinmike:
Previously, the following operators were incorrectly interpreted as
"infix" operators:

- `++`
- `--`
- `yield`

The internal `isInfix` function tended to obscure this error because it
identified "infix" tokens via a boolean logic expression that acted as a
blacklist. This expression was difficult to understand and hostile to
future extensions for new language features (because future tokens could
be mistakenly implemented as "infix" operators).

Update all "infix" operator token definitions to explicitly define the
`infix` property, obviating the need for the internal `isInfix`
function. Ensure that the aforementioned tokens are not interpreted as
"infix" operators, This corrects the behavior of the internal
`isEndOfExpr` function, which is appreciable when linting code that
relies on automatic semicolon insertion.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -805,12 +805,15 @@ var JSHINT = (function() {
     }
   }
 
-  function isInfix(token) {
-    return token.infix ||
-      (!token.identifier && !token.template && !!token.led) ||
-      // Although implemented as an Identifier, the `yield` keyword behaves as
-      // an operator when it appears in the body of a generator function.
-      (token.id === "yield" && !!state.funct["(generator)"]);
+  /**
+   * Determine whether a given token is an operator.
+   *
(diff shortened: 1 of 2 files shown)


jugglinmike:
@rwaldron Mind taking a look? This should resolve gh-3037.

src/jshint.js
```diff
@@ -805,12 +805,15 @@ var JSHINT = (function() {
...
-      // an operator when it appears in the body of a generator function.
-      (token.id === "yield" && !!state.funct["(generator)"]);
+  /**
+   * Determine whether a given token is an operator.
+   *
+   * @param {token} token
+   *
+   * @returns {boolean}
+   */
+  function isOperator(token) {
```

Issue #3039, opened 2016-10-01
Incomplete instanceof crashes


Guichaguri:
Writing an incomplete instanceof throws `Cannot read property 'id' of undefined`.

The following input code should reproduce the problem:
```js
var something = [];
if(something instanceof) {
    // Notice the instanceof is missing the class name
}
```

I'm currently using JSHint 2.9.3

emileber:
I'm having this problem through the atom package `linter-jshint`.

```
Error: C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:5386
        throw err;
        ^

TypeError: Cannot read property 'id' of undefined
    at C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:2131:14
    at Object.x.led (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:1155:16)
    at expression (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:926:36)
    at C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:2051:18
    at Object.x.led (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:1155:16)
    at expression (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:926:36)
    at Object.fud (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:4423:22)
    at expression (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:885:32)
    at statement (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:1636:9)
    at statements (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\jshint\src\jshint.js:1681:16)
    at ChildProcess.<anonymous> (C:\Users\emile.bergeron.CAMINO\.atom\packages\linter-jshint\node_modules\atom-linter\node_modules\sb-exec\lib\index.js:56:20)
[...]

PR #3046, opened 2016-10-17
[[FIX]] Avoid crashing on invalid input

jugglinmike:
A recent extension to the `instanceof` operator introduced logic that
performed property access on the result of parsing the operator's
right-hand expression operand. Because invalid code may not specify such
an expression, this logic led to program crashes in response to the
invalid syntax.

Ensure that when the `instanceof` operator appears without a right-hand
expression operand, JSHint reports a syntax error and exits gracefully.

This should resolve gh-3039. @rwaldron mind taking a look?

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2128,6 +2128,12 @@ var JSHINT = (function() {
     token.left = left;
     token.right = right = expression(120);
 
+    // This condition reflects a syntax error which will be reported by the
+    // `expression` function.
+    if (!right) {
+      return token;
+    }
+
     if (right.id === "(number)" ||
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -2128,6 +2128,12 @@ var JSHINT = (function() {
     token.left = left;
     token.right = right = expression(120);
 
+    // This condition reflects a syntax error which will be reported by the
+    // `expression` function.
+    if (!right) {
+      return token;
+    }
```
rwaldron: Nice and simple!

jugglinmike:
Thanks, Rick!

PR #3047, opened 2016-10-17
[[FIX]] Consistently lint evaluated code

jugglinmike:
Previously, upon encountering certain forms of dynamic code evaluation
(namely the browser's `setTimeout` and `setInterval` functions), JSHint
would attempt to lint the code being evaluated. Due to what appears to
to be an unintentional typo on the JSHint source code, this behavior did
not extend to code evaluated via the `Function` and `eval` functions.

Extend this behavior to all forms of evaluated code.

---

@rwaldron: This is about 3 bugs separated from a more important issue, but I
want to tackle each part on its own. What do you think about this? It's
backwards-breaking, but the current behavior is clearly inconsistent and clearly
the result of a typo.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -2454,7 +2454,7 @@ var JSHINT = (function() {
             left.value === "execScript") {
           warning("W061", left);
 
-          if (p[0] && [0].id === "(string)") {
+          if (p[0] && p[0].id === "(string)") {
             addInternalSrc(left, p[0].value);
           }
         } else if (p[0] && p[0].id === "(string)" &&
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -452,7 +452,23 @@ exports.escapedEvil = function (test) {
 
   TestRun(test)
     .addError(1, "eval can be harmful.")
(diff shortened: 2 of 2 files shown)


src/jshint.js
```diff
@@ -2454,7 +2454,7 @@ var JSHINT = (function() {
             left.value === "execScript") {
           warning("W061", left);
 
-          if (p[0] && [0].id === "(string)") {
```
rwaldron: That's AMAZING.

jugglinmike:
Rick and I have discussed this offline, and we've decided that unfortunately,
we cannot include this fix outside of a major release of JSHint.

While it is a bug fix, and Semantic Versioning supports fixing bugs (even those
that are backwards-incompatible) with patch release, the important detail here
is that the underlying behavior is not documented. As far as I can see, there
is no wording in JSHint's documentation which explains how JSHint is expected
to lint any dynamically-modified code it encounters. Because of that, enabling
it in any new context (even though doing so would be more logical/consistent
overall) is effectively introducing a new feature.

In light of that, I'm going to label this fix for inclusion in the next major
release.

PR #3050, opened 2016-10-18
Do not duplicate reported warnings/errors

jugglinmike:
The final commit is where it's at. The first two (along with https://github.com/jshint/jshint/pull/3049 before them) address more minor bugs in service of that change.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -408,7 +408,7 @@ var JSHINT = (function() {
   }
 
   // Tracking of "internal" scripts, like eval containing a static string
-  function addInternalSrc(elem, token) {
+  function addEvalCode(elem, token) {
     JSHINT.internals.push({
       id: "(internal)",
       elem: elem,
@@ -2465,13 +2465,13 @@ var JSHINT = (function() {
           // to the fact that the behavior was never formally documented). This
           // branch should be enabled as part of a major release.
           //if (p[0] && p[0].id === "(string)") {
-          //  addInternalSrc(left, p[0]);
+          //  addEvalCode(left, p[0]);
           //}
         } else if (p[0] && p[0].id === "(string)" &&
              (left.value === "setTimeout" ||
             left.value === "setInterval")) {
           warning("W066", left);
-          addInternalSrc(left, p[0]);
+          addEvalCode(left, p[0]);
 
         // window.setTimeout/setInterval
         } else if (p[0] && p[0].id === "(string)" &&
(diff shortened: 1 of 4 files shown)


jugglinmike:
As near as I can tell, the decrease in code coverage is due to the fact that
this patch reduces the total number of overall branches. It introduces only one
new branch (the nested `for` loop within the new `lintEvaled` function), and
that is definitely under test here.

src/jshint.js
```diff
@@ -409,15 +409,13 @@ var JSHINT = (function() {
   }
 
   // Tracking of "internal" scripts, like eval containing a static string
-  function addInternalSrc(elem, src) {
-    var i;
-    i = {
+  function addInternalSrc(elem, token) {
```
rwaldron: Since you're refactoring, it makes sense to rename this more appropriate, perhaps: `addEvalCode`. https://tc39.github.io/ecma262/#sec-types-of-source-code

src/jshint.js
```diff
@@ -5152,13 +5143,38 @@ var JSHINT = (function() {
...
+  /**
+   * Lint dynamically-evaluated code, appending any resulting errors/warnings
+   * into the global `errors` array.
+   *
+   * @param {array} internals - collection of "internals" objects describing
+   *                            string tokens that contain evaluated code
+   * @param {object} options - linting options to apply
+   * @param {object} globals - globally-defined bindings for the evaluated code
+   */
+  function lintEvaled(internals, options, globals) {
```
rwaldron: Maybe `lintEvalCode()`? To match the earlier suggestion.

tests/helpers/testhelper.js
```diff
@@ -126,7 +132,8 @@ exports.setup.testRun = function (test, name) {
 
       test.ok(
         undefinedErrors.length === 0
-          && unthrownErrors.length === 0 && wrongLineNumbers.length === 0,
+          && unthrownErrors.length === 0 && wrongLineNumbers.length === 0
+          && duplicateErrors.length === 0,
```
rwaldron: A quick ctrl-f of this file shows that this is the only place where the logical operator is at the beginning of the line.

src/jshint.js
jugglinmike: Sounds good to me!

src/jshint.js
jugglinmike: Perfect

tests/helpers/testhelper.js
jugglinmike: Nice catch--updated

rwaldron:
Awaiting green light from CI

jugglinmike:
Squashed intermediary commits in preparation for merge to `master`.

rwaldron:
Nice work!

PR #3051, opened 2016-10-18
Warning/error refactoring

jugglinmike:
@rwaldron As a public API, I don't recommend that we change these message
identifiers lightly. Here is why I believe this change is safe:
- Changing errors to warnings: users cannot disable errors by identifier (e.g.
  `-E012` is reported as a "bad option") so this makes JSHint more configurable
  without effecting any existing installations
- Changing warnings to errors: this may introduce new errors--users that had
  previously disabled warning `W094` will now receive error `E069` in its
  place. This should not be a problem because each of the warnings "escalated"
  to errors concern invalid JavaScript, making it highly unlikely that any
  consumer is actually disabling them.

What do you think?


rwaldron:
@jugglinmike Needs rebasing

jugglinmike:
@rwaldron Rebase complete!

rwaldron:
@jugglinmike thanks!

jugglinmike:
I've rebased this again. Because we'd like to trial this change via a release candidate, we're going to defer to version 2.9.10. I've created a branch to contain new features for that version and updated this pull request to target that new branch.

rwaldron:
Sorry Mike, can you resolve and rebase this patch one last time? :|

rwaldron:
I attempted to rebase this today and was unable to successfully do so. I want to try one more time, but I need to come back to it later.

rwaldron:
I've attempted to rebase this again today and I couldn't even resolve the first change.

Issue #3055, opened 2016-10-26
Labels Break JSHint


iamlikeme-x:
The following produces the error `E053 - Export declarations are only allowed at the top level of module scope.`

Whilst I appreciate that `const` and `export` shouldn't be used in production code (unless controlled environments, eg. Electron - the use case I have) this still shouldn't be happening AFAIK.

``` javascript
testLabel: {
    break testLabel;
}
export const test = 'test';
```

It appears JSHint isn't matching the closing brace properly, as the following supresses the error despite being syntactically incorrect:

``` javascript
testLabel: {
    break testLabel;
}
}
export const test = 'test';
```

jugglinmike:
Confirmed. Here's a further-reduced test case:

```js
// jshint esversion: 6, module: true
x: {}
export default null;
```

You're right to suspect improper scope tracking. It would appear that the same underlying bug is effecting the binding tracking mechanism. Note how `undef: true` ought to trigger a warning in the following code but currently does not:

```js
// jshint esversion: 6, module: true, undef: true
{ let y; }
void y;
```

Thanks for the report!

jugglinmike:
Submitted a patch here: https://github.com/jshint/jshint/pull/3057

jugglinmike:
This was resolved in gh-3057 and [subsequently released as part of JSHint version 2.9.5](https://github.com/jshint/jshint/releases/tag/2.9.5).

PR #3057, opened 2016-10-30
[[FIX]] Close synthetic scope for labeled blocks

jugglinmike:
@rwaldron Honestly, I'm not such a big fan of the way labeled statements have
been implemented here. Creating a synthetic scope is just asking for bugs
exactly like gh-3055.

That said, it seems kind of onerous to implement more accurate semantics. I
think we'd need to extend the `expression` function to accept an optional label
identifier. When it encountered block-creating statements, it would assign the
label correctly.

The approach I'm proposing here is much simpler, and it is probably "safer" for
it. Maybe we can address the synthetic scope in some future patch. What do you
think?

Commit message:

> [[FIX]] Close synthetic scope for labeled blocks
> 
> JSHint creates a synthetic block scope in order to track labeled
> statements prior to parsing the statements themselves. Previously, this
> scope was not closed (or "unstacked") for labeled Block statements.
> 
> Ensure that the synthetic scope is closed in these cases as well. Assert
> correct behavior via various parsing/linting mechanisms that are aware
> of scope.

Resolves gh-3055.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1625,6 +1625,11 @@ var JSHINT = (function() {
       //  }
       var iscase = (state.funct["(verb)"] === "case" && state.tokens.curr.value === ":");
       block(true, true, false, false, iscase);
+
+      if (hasOwnScope) {
+        state.funct["(scope)"].unstack();
+      }
+
       return;
     }
 
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -867,6 +867,14 @@ exports.testES6Modules = function (test) {
   TestRun(test)
     .test(src2, {});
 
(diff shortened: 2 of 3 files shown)


rwaldron:
> The approach I'm proposing here is much simpler, and it is probably "safer" for
it. Maybe we can address the synthetic scope in some future patch. What do you
think?

I definitely support and prefer the cautious path.

rwaldron:
I'm not sure why merging this locally didn't close the PR, but the code landed here: https://github.com/jshint/jshint/commit/5f0f78926f3a7f40ada6f4b3bcd8f5f35d5b0d38

PR #3054, opened 2016-10-20
Esversion normalization

jugglinmike:
This patch contains two refactoring commits followed by a bug fix that should
resolve both gh-2909 and ~~gh-3051~~ gh-2902.

PR #3060, opened 2016-11-07
esversion: 2016

jugglinmike:
@rwaldron It's about time for this, wouldn't you say?

Because it's a new feature, I'm submitting this patch as pull request against the `v2.10.0` branch. This is a slight complication to our traditional "master-only" workflow, but I want to keep 2.9 available for bug fixes. If this works for 2.10.0, then I'll update the `CONTRIBUTING.md` file with new instructions.

If possible, I'd like to see gh-3054 landed in `master` first, though, since it concerns the interpretation of the `esversion` option's values. After that, this will need to be rebased.


src/jshint.js
```diff
@@ -653,12 +653,15 @@ var JSHINT = (function() {
...
+          case "7":
             state.option.moz = false;
             state.option.esversion = +val;
             break;
           case "2015":
+          case "2016":
             state.option.moz = false;
-            state.option.esversion = 6;
+            // Translate specification publication year to version number.
+            state.option.esversion = +val - 2009;
```
rwaldron: lol

nicolo-ribaudo:
I just tested this pr with the `singleGroups` option, but it is buggy:
```js
// jshint esversion: 2016, expr: true, singleGroups: true

2 ** 3 ** 4 === 2.4178516392292583e+24;
2 ** (3 ** 4) === 2.4178516392292583e+24;
(2 ** 3) ** 4 === 4096;
```
```
$ bin/jshint demo.js
demo.js: line 4, col 6, Unnecessary grouping operator.
demo.js: line 5, col 1, Unnecessary grouping operator.

2 errors
```

Line 5 should be ok.

nicolo-ribaudo:
This also doesn't work:
```js
// jshint esversion: 2016, expr: true

(-2) ** 3 === -8;
```
```
$ bin/jshint demo.js
demo.js: line 3, col 6, Unexpected '**'.

1 error
```

src/lex.js
```diff
@@ -341,6 +341,14 @@ Lexer.prototype = {
...
+    if (ch1 === "*" && ch2 === "*") {
+      return {
+        type: Token.Punctuator,
+        value: "**"
+      };
+    }
+
     // 2-character punctuators: <= >= == != ++ -- << >> && ||
     // += -= *= %= &= |= ^= /=
     if (ch1 === ch2 && ("+-<>&|".indexOf(ch1) >= 0)) {
```
jugglinmike: Totally! I missed this opportunity because the surrounding comments are a
little inaccurate. I've updated those as well.

jugglinmike:
@nicolo-ribaudo Thanks for the review! It looks like I underestimated the
complexity of this operator. I think I've addressed both issues, although the
code is no prettier for my efforts.

jugglinmike:
Hey @caitp: I've rebased this branch over `master` and (thanks to Test262) found it lacking. Specifically: it did not implement support for compound exponentiation assignment, `**=`. Other than that, though, it was just dandy. I've corrected that, squashed all the `fixup!` commits, and force- ([the original branch is available here](https://github.com/jshint/jshint/compare/master...jugglinmike:esversion2016-orig) in case anyone is interested). Would you mind reviewing this?

src/jshint.js
```diff
@@ -2002,6 +2035,19 @@ var JSHINT = (function() {
     error("E014");
   };
   assignop("%=", "assignmod", 20);
+  assignop("**=", function(left, that) {
+    if (!state.inES7()) {
```
caitp: Seems like you might consider renaming `inES7()` to match `esversion` values, I don't think maintaining api compat is a big deal for jshint. It's awkward to have two conventions in the same project.

src/jshint.js
```diff
@@ -2049,6 +2095,26 @@ var JSHINT = (function() {
...
   }, orPrecendence);
   infix("&&", "and", 50);
+  // The Exponentiation operator, introduced in ECMAScript 2016
+  //
+  // ExponentiationExpression[Yield] :
+  //   UnaryExpression[?Yield]
+  //   UpdateExpression[?Yield] ** ExponentiationExpression[?Yield]
+  infix("**", function(left, that) {
+    if (!state.inES7()) {
+      warning("W119", that, "Exponentiation operator", "7");
```
caitp: Again, conventions should match

src/jshint.js
```diff
@@ -2581,6 +2649,9 @@ var JSHINT = (function() {
           (isFunctor(ret) && !isEndOfExpr()) ||
           // Used as the return value of a single-statement arrow function
           (ret.id === "{" && preceeding.id === "=>") ||
+          // Used to cover a unary expression as the left-hand side of the
+          // exponentiation operator
+          (beginsUnaryExpression(ret) && state.tokens.next.id === "**") ||
```
caitp: I'd prefer it if we rewrote this whole block of conditionals, but that's not a blocker for this.

src/jshint.js
```diff
@@ -2598,7 +2669,7 @@ var JSHINT = (function() {
         isNecessary =
           (rbp > first.lbp) ||
           (rbp > 0 && rbp === first.lbp) ||
-          (!isEndOfExpr() && last.lbp < state.tokens.next.lbp);
+          (!isEndOfExpr() && last.rbp < state.tokens.next.lbp);
```
caitp: bugfix?

src/jshint.js
```diff
@@ -653,12 +653,15 @@ var JSHINT = (function() {
           case "3":
           case "5":
           case "6":
+          case "7":
```
caitp: hmmm --- maybe for some kind of design aesthetic, should save the esversion option as it was written, and print it rather than "7" when converting it to text. That way, there won't be a mismatch if `esversion=2015`

jugglinmike: I like this idea, but the variability in warning messages may fragment Google/StackOverflow search results. What do you think about extending the message to suggest both alternatives, e.g.

> 'Exponentiation operator' is only available in ES7 (use 'esversion: 7' or 'esversion: 2016').

...or (maybe better) omitting the suggested value altogether:

> 'Exponentiation operator' is only available in ES7. Enable it with the 'esversion' option.

src/jshint.js
jugglinmike: The `esversion` option supports both "traditional" values (e.g. 5, 6, 7) and "year-based" values (though it's currently a little broken\*). I think we're internally consistent, though we're using the "traditional" scheme. So method names like `inES7` seem logical enough to me.

\* I've had a pull request open to fix that for a little while now: https://github.com/jshint/jshint/pull/3054. If you get a chance, could you take a look at that, as well?

src/jshint.js
jugglinmike: I'm down for that! I wrote this originally, though, so I'm at a disadvantage when it comes to identifying improvements. Maybe you can get the ball rolling in a follow-up pull request?

src/jshint.js
jugglinmike: Arrgg wrong place for that response. Re-posting...

src/jshint.js
jugglinmike: The `esversion` option supports both "traditional" values (e.g. 5, 6, 7) and "year-based" values (though it's currently a little broken\*). I think we're internally consistent, though we're using the "traditional" scheme. So method names like `inES7` seem logical enough to me.

\* I've had a pull request open to fix that for a little while now: https://github.com/jshint/jshint/pull/3054. If you get a chance, could you take a look at that, as well?

src/jshint.js
jugglinmike: Up until now, all the `rbp` and `lbp` were interchangeable. The exponentiation operator is the first feature to need distinct values, and that matters right here. This is what the somewhat lengthy comment in the `symbol` function is intended to explain.

src/jshint.js
rwaldron: Follow up: #3054 is merged (no action needed)

src/jshint.js
rwaldron: @caitp can you write an issue for this? Thanks 😄

rwaldron:
Nice work here, sorry it took me so long to review. I promise I will do better for ES2017

PR #3061, opened 2016-11-09
[[DOCS]] Add Codacy to "Who uses JSHint" list

funkylexoo:


jugglinmike:
It took some digging, but I found a primary source for Codacy's use of JSHint

> ### Open Source Tools
>
> In addition to in-house built rules, we use open source tools for many of our
> static analysis. We want to express our gratitude to those tools and to
> everyone who contributed.
>
> [...]
>
> - [jshint](https://github.com/jshint/jshint)

https://support.codacy.com/hc/en-us/articles/207995005-Special-Thanks

So I merged your patch (ac4f7c064db590e1de6b87d1772c18b198804e57) and added
this in a follow-up commit (12afbcdeb38edcbe379b1e9eb9b388e3741ecc75).

If anyone reading this would like to submit a similar patch, please include a
reference in your patch!

Issue #3004, opened 2016-08-10
Starting jsdoc comment causes 'Unclosed regular expression' error


dylansteele8:
When I begin to type `/*` to start a jsdoc comment, I get a jshint error: `E015 - Unclosed regular expression`. Is there a way to turn this error off for jsdoc comments?

rwaldron:
Close the comment: `*/`

dylansteele8:
Sorry, the main issue is with Atom; it throws type errors and other exceptions:
```
Error: /Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:5309 throw err; ^

TypeError: Cannot read property 'id' of null at Object.x.nud (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:3228:26) at expression (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:884:34) at /Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:2427:23 at Object.x.led (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:1151:16) at expression (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:922:36) at Object.fud (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:4599:7) at expression (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:881:32) at statement (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:1632:9) at statements (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:1677:16) at itself (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:5288:9)
Error: /Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:5309
        throw err;
        ^
TypeError: Cannot read property 'id' of null
    at Object.x.nud (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:3228:26)
    at expression (/Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:884:34)
    at /Users/<user>/.atom/packages/linter-jshint/node_modules/jshint/src/jshint.js:2427:23
[...]

PR #3063, opened 2016-11-12
[[FIX]] Avoid crash when peeking past end of prog

jugglinmike:
The internal `peek` function may be used to request tokens beyond the
end of the program. Because multiple call sites expect the function to
consistently return an object value, the "end of program" token is be
returned in these cases.

However, this situation may occur in cases where the "end of program"
token is not the next token. The previous implementation of the function
would produce the `null` value in this event, violating callers'
expectation and resulting in runtime reference errors.

Update the internal `peek` function to consistently return the "end of
program" token when input has been exhausted--even when this is not the
next token.

This should resolve gh-3004. @rwaldron What do you think?

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -699,12 +699,23 @@ var JSHINT = (function() {
     }
   }
 
-  // We need a peek function. If it has an argument, it peeks that much farther
-  // ahead. It is used to distinguish
-  //     for ( var i in ...
-  // from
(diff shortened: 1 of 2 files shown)


rwaldron:
Again, not sure why merging this locally didn't result in closing the PR. https://github.com/jshint/jshint/commit/c0838666dd53ac40106208d3a3edf9a9f77e9b18

PR #3066, opened 2016-11-23
[[FIX]] Throw W033 instead of E058 when the ; after a do-while stmt is missing (continued)

jugglinmike:
This extends gh-2797 with a few minor alterations.

@nicolo-ribaudo This review has been a long time coming--I felt obliged to implement my feedback directly. Do these changes look good to you?


nicolo-ribaudo:
LGTM 👍

PR #3083, opened 2017-01-03
Fixes a typo and missing space

hlapp:
I also _think_ that "prerequisite" should be plural, but I may not understand enough what that term means here, so left that as is. I'd be happy to include that though 😄

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -17,7 +17,7 @@ var OPTIONS = {
   "reporter": ["reporter", "Custom reporter (<PATH>|jslint|checkstyle|unix)", "string", undefined ],
   "prereq": [
     "prereq",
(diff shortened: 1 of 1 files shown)


jugglinmike:
Hey Hilmar, thanks for the patch! I think you're right--"prerequisite" should be plural here. Mind including that as well?

hlapp:
Done 😃

jugglinmike:
Wow, so fast! I squashed these commits, updated the message to satisfy our [contribution
guidelines](https://github.com/jshint/jshint/blob/540ed85f075f7ae1317a37447c9fe9644e0d62e4/CONTRIBUTING.md#header) and applied that to `master` at e7abce2dfb8a37addb49d4cb17ffffdda44f639f.

Thanks again :)

PR #3084, opened 2017-01-04
[[FEAT]] Allow HTML-like comments

### LINKED ISSUE #1256: jshint crashes with certain character sequences (html-style comments)
The following code crashes jshint:

``` js
var a = 2;
<!-- a = 3;
--> a = 4;
console.log(a);
```

Those character sequences are actually [HTML-style comment markers in the JS](http://javascript.spec.whatwg.org/#comment-syntax), which (crazy enough) are actually valid in all JS engines.

They are not officially in the ES spec yet, but [apparently they will eventually be standardized in ES6](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-B.1.3).

It would seem that all JS tools should align with how JS engines (and ES6 eventually) allow these, even if we think they're a bad idea.

Perhaps these should be identifiable and properly handled by jshint, but ALSO have a new jshint rule that can complain about finding them and warning against using them as a bad idea. But then again, this is perfectly valid ES:

``` js
var a = 1, b = 1; a <!--b;
```

Would such a rule assume you were doing the bad comment style, or doing the (admittedly weird/strange) comparison-negation-decrement operator dance (which works)?

I personally think jshint should handle these and have a warning (not a failure rule) if it finds either one (suggesting you just add some whitespace to disambiguate).

But regardless of how that shakes out, jshint definitely shouldn't hard-crash. ;-)

### LINKED ISSUE #3082: <!-- single line comments not lexed properly
`<!--` is a HTML-like comment starts a single line comment (According to <http://www.ecma-international.org/ecma-262/6.0/#sec-html-like-comments>), but jshint throws around 5 warnings when it's used as one:

```
1	Expected an identifier and instead saw '<'.
1	Expected an operator and instead saw '!'.
1	Expected an operator and instead saw '--'.
1	Expected an assignment or function call and instead saw an expression.
1	Missing semicolon.
```
(This was from http://jshint.com/ with just the line `<!--`. It should throw no errors.)

They work like regular `//` comments: The line is ignored from the `<!--` upto, but not including, a line terminator (So it's just another single-line comment).

And for non-browser environments, in Node.js v7.3.0, the following file when run:

```javascript
<!-- console.log('Fail');
console.log('Pass');
```

only outputs `Pass`.

---

A fix to this would be adding these lines at line 492 in [/src/lex.js](https://github.com/jshint/jshint/blob/master/src/lex.js#L492):

```javascript
    // ...
      this.skip(2);
      return null;
    }

    // HTML-like comments start with <!-- and cause the line to be ignored upto the EOL
    if (ch1 === "<" && ch2 === "!" && this.peek(2) === "-" && this.peek(3) === "-") {
      this.skip(this.input.length);  // Skip to the EOL
      return commentToken("<!--", rest.substr(2));  // Remove the two -- characters in rest
    }

    // Comments must start either with // or /*
    if (ch1 !== "/" || (ch2 !== "*" && ch2 !== "/")) {
      return null;
    }
    // ...
```
This has the side effect that `<!--` will allow members and globals comments without the qualifying `<!-- jshint: `.

([Current revision of lex.js](https://github.com/jshint/jshint/blob/b646aeae63f49a24e649a86dbc8d94c20d57f38c/src/lex.js#L492))

MitalAshok:
HTML-like comments were standardised in ECMAScript 2015 (ES5),
so this raises a warning when esversion < 5.

The HTML-like comments will also allow `globals` and the (deprecated) `members` directive where
regular single line comments can't. They also allow the regular `js[hl]int` directives.

A test was added and a new warning.
W140: "HTML-like comments (`<!--`) may not be supported before ES5."

BREAKING CHANGE:

`<!--` will no longer be interpreted as "less than not pre-decrement", for example in:

    for(var i = 1; 0<!--i;) { /* This will run once */ }

This can be fixed by putting a space between the `<` and the `!`.

Closes #3082
Closes #1256

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -490,7 +490,26 @@ Lexer.prototype = {
       return null;
     }
 
-    // Comments must start either with // or /*
+    // HTML-like comments start with <!-- and cause the line to be ignored upto the EOL
+    if (!state.option.htmlcomments && ch1 === "<" && ch2 === "!" &&
+      this.peek(2) === "-" && this.peek(3) === "-") {
+      // Warn if HTML-like comments are used before ES5, where they were
+      // standardised, but still widely implemented.
+      if (!state.inES5()) {
+        this.triggerAsync("warning", {
+            code: "W119",
+            line: this.line,
+            character: this.char,
+            data: ["<!--", "5"]
+          }, checks, function() { return true; }
+        );
+      }
+      this.skip(this.input.length);  // Skip to the EOL
+      // Remove the two -- characters in rest
+      return commentToken("<!--", rest.substr(2));
+    }
+
+    // Comments must start with either // or /* if they aren't HTML-like
     if (ch1 !== "/" || (ch2 !== "*" && ch2 !== "/")) {
       return null;
     }
@@ -1613,7 +1632,8 @@ Lexer.prototype = {
       state.option.maxlen < this.input.length) {
       var inComment = this.inComment ||
         startsWith.call(inputTrimmed, "//") ||
-        startsWith.call(inputTrimmed, "/*");
+        startsWith.call(inputTrimmed, "/*") ||
+        (!state.option.htmlcomments && startsWith.call(inputTrimmed, "<!--"));
 
       var shouldTriggerError = !inComment || !reg.maxlenException.test(inputTrimmed);
 
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -2191,3 +2191,19 @@ exports["TDZ within for in/of head"] = function(test) {
 
   test.done();
 };
+
+exports["HTML-like comments (<!--)"] = function(test) {
+  var code = [
+    "<!--global c",
+    // --> does not end a HTML-like comment
+    "<!--c-->c",
+    "c.c = 1;"
+  ];
+
+  TestRun(test).test(code, { esversion: 5, htmlcomments: false });
+  TestRun(test)
+    .addError(2, "'<!--' is only available in ES5 (use 'esversion: 5').")
+    .test(code, { esversion: 3, htmlcomments: false });
+
+  test.done();
+};
--- a/examples/.jshintrc
+++ b/examples/.jshintrc
@@ -11,6 +11,7 @@
     "eqeqeq"        : true,     // true: Require triple equals (===) for comparison
     "forin"         : true,     // true: Require filtering for..in loops with obj.hasOwnProperty()
     "freeze"        : true,     // true: prohibits overwriting prototypes of native objects such as Array, Date etc.
+    "htmlcomments"  : true,     // true: Prohibit use of HTML-like comments (`<!--`)
     "immed"         : false,    // true: Require immediate invocations to be wrapped in parens e.g. `(function () { } ());`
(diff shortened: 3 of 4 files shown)


tests/unit/core.js
```diff
@@ -1333,7 +1333,7 @@ exports.testES6TemplateLiteralReturnValue = function (test) {
 
   TestRun(test).test(src, { esnext: true });
 
-  var src = [
+  src = [
```
nicolo-ribaudo: This correction is not related to this pr. (Also the other corrections)

src/messages.js
```diff
@@ -223,7 +223,8 @@ var warnings = {
   W136: "'{a}' must be in function scope.",
   W137: "Empty destructuring.",
   W138: "Regular parameters should not come after default parameters.",
-  W139: "Function expressions should not be used as the second operand to instanceof."
+  W139: "Function expressions should not be used as the second operand to instanceof.",
+  W140: "HTML-like comments (`<!--`) may not be supported before ES5."
```
nicolo-ribaudo: You could use W119

MitalAshok:
@nicolo-ribaudo I have commited some more and I added a `htmlcomments` option, but it is inverted right now (true = prohibit, false = allow), should it be in the relaxing options, or should it be renamed?

Also, right now it warns even if `htmlcomments` is true if you are on pre-es5. Should it not warn if you have the option to allow it?

And I couldn't get it to be conditionally true if isES5. Where do I put that? In `src/jshint.js`?

I also reverted the changes to the tests, and am using W119.

nicolo-ribaudo:
I think `htmlcomments: false` -> `HTML-like comments are enabled` can be confusing, it should be the other way (`htmlcomments: true` enables HTML-like comments).

src/lex.js
```diff
@@ -490,7 +490,26 @@ Lexer.prototype = {
...
+      // standardised, but still widely implemented.
+      if (!state.inES5()) {
+        this.triggerAsync("warning", {
+            code: "W119",
+            line: this.line,
+            character: this.char,
+            data: ["<!--", "5"]
+          }, checks, function() { return true; }
+        );
+      }
```
nicolo-ribaudo: Since this is an opt-in option, is this warning necessary?

jugglinmike:
Since setting the option to `true` actually *disables* support for HTML-like
comments, I think a name like `nohtmlcomments` would be more appropriate.
However, to your point about the "breaking change," this is indeed technically
backwards-incompatible. Given the realities of today's engines, we could
probably justify calling this a "bug fix," but for a fringe feature like this,
the conservative path might be better. In other words--keeping the option name
`htmlcomments`, but making the new behavior opt-in.

Separately: as an informally evolved feature, HTML-like comments are
surprisingly tricky to get right. For instance, did you know that the following
is valid according to the latest specification:

    --> foo bar baz?

So the parsing strategy proposed here is not quite complete.

The complete grammar is defined here:
https://tc39.github.io/ecma262/#sec-html-like-comments, and there are some
tests in Test262:
https://github.com/tc39/test262/tree/4ff7e4342e1a15ce7af893ba9af9bd55c7d95e51/test/annexB/language/comments

Would you be up to the challenge of getting this fully compliant with the
latest specification text?

PR #3088, opened 2017-01-08
Update options.js

nadgerz:
JSCS project merged with ESLint.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -28,8 +28,8 @@ exports.bool = {
      *
      * @deprecated JSHint is limiting its scope to issues of code correctness.
      *             If you would like to enforce rules relating to code style,
-     *             check out [the JSCS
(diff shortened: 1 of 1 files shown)


jugglinmike:
I'm not so sure about this modification. Despite not being maintained, JSCS is
still a well-tested and complete tool. It's "deprecated" status is
well-documented in the project itself, and may change in the future.

@rwaldron thoughts?

nadgerz:
Very good point!

I guess I clicked on the link and saw the 'We Have Moved!' comment and
(naively) decided to suggest the change.

Perhaps I should change it to show a link to both?

I fully realise after getting to the JSCS site people might be able to work
out what they should do though :)

s.

rwaldron:
I don't think this is a necessary change.

PR #3090, opened 2017-01-15
[[FEAT]] Add new "enforcing" option: `trailingcomma` (updated)

jugglinmike:
This changeset continues the work first submitted by @jacksonrayhamilton in
gh-2100. As discussed in that pull request, a few changes to `master` in the
years since he submitted have introduced conflicts in the patch. I've resolved
those with a merge commit because I believe that's the most transparent way to
document this collaboration.


src/jshint.js
```diff
@@ -2743,6 +2743,9 @@ var JSHINT = (function() {
           break;
         }
       } else {
+        if (state.option.trailingcomma && state.inES5()) {
```
jacksonrayhamilton: Do you think the ES5 check should be removed? (Continuing from here: https://github.com/jshint/jshint/pull/2100#discussion_r89679455.)

From a language purist perspective, it should be. Although for practical reasons it is probably still useful for any sorry soul who still has to support old IE. I don't see an option specifically for detecting old IE bugs, otherwise I'd suggest that instead.

src/options.js
```diff
@@ -239,7 +239,47 @@ exports.bool = {
      *             warnings/errors in when upgrading between minor versions of
      *             JSHint.
      */
-    enforceall : false
+    enforceall : false,
+
+    /**
+     * This option warns when a comma is not placed after the last element in an
+     * array or object literal. Due to bugs in old versions of IE, trailing
+     * commas used to be discouraged, but since ES5 their semantics were
```
jacksonrayhamilton: As you mentioned [here](https://github.com/jshint/jshint/pull/2100#discussion_r89679455), the syntax was actually specified before ES5. If we went the "practical" route I suggest [here](https://github.com/jshint/jshint/pull/3090/files#r96153612), we could reword this to, "but since IE implementations of ES5, their semantics became uniform across most JS environments."

src/jshint.js
jugglinmike: JSHint has a few IE-specific warnings (e.g. `W070`, "Extra comma. (it breaks
older versions of IE)"), and rather than referencing an option dedicated to
that condition, it applies these a little inconsistently--when the environment
does not support ES5. So this behavior respects precedence.

But like you, I'm not sure the precedent is a good one. I'm thinking that a new
"enforcing" option like `oldie` might be good, especially from a language
purist's perspective. I've written up a proposal here: gh-3094

rwaldron:
This is great work :)

PR #3093, opened 2017-01-19
[[FEAT]] jest testing framework support

koresar:
The list of variable names was copied from here: https://github.com/sindresorhus/globals/blob/4159cd067369b8242131551beb09fafb52afc289/globals.json

**UPD:**
Hello maintainers.
My apologies for not following your guidelines strictly. All the changes were done in the GitHub online code editor. I hope that's fine due to the new "Squash and Merge" button GitHub has recently introduced.

The build fails due to coverage decrease. Here is the line which causes the build failure:
image
I hope that's fine too because most of those options are not covered with tests anyway.

Could you please accept the PR? It looks tidy as to my taste. :)
Thanks!


jugglinmike:
Thanks for the patch! We appreciate your taking the time to contribute back to JSHint. Unfortunately, we're trying to move away from supporting library-specific "environment" options in JSHint "core". JavaScript libraries come and go, and even the well-established ones like Jest are subject to change over time. When their API changes, it puts JSHint in an awkward position of having to support multiple versions concurrently. The other libraries you've referenced are effectively "grandfathered in,"--we're taking the maintenance burden to honor commitments made before this change in policy.

koresar:
@jugglinmike no worries. Can you help me then?
What's the preferred new way? (The link is enough.) 
Thank you

jugglinmike:
I haven't used the Jest library before, but based on your patch, I assume you'll just want to add the same keys to JSHint's `globals` configuration. So, in your `.jshintrc` file, you would write:

```json
{
  "globals": {
    "afterAll": false,
    "afterEach": false,
    "beforeAll": false,
    "beforeEach": false,
    "check": false,
    "describe": false,
    "expect": false,
    "gen": false,
    "it": false,
    "fit": false,
    "jest": false,
    "pit": false,
    "require": false,
    "test": false,
    "xdescribe": false,
    "xit": false,
    "xtest": false
  }
}
```

Later, if a new version of Jest changes the set of globally-defined bindings, you can update this file in your project at the same time you update to that new version--no coordination with the JSHint project will be necessary :)

For more on configuring JSHint, check out [the project documentation on jshint.com](http://jshint.com/docs/).

Hope this helps!

koresar:
@jugglinmike that's very helpful. Thanks mate! I appreciate your time and effort.
Cheers

PR #3098, opened 2017-02-11
[[TEST]] Run Test262 tests

jugglinmike:
Include the Test262 project as a submodule and introduce a new test
script that attempts to parse every test file it contains. Because this
project currently fails a number of valid tests with both type I and
type II errors, include a text file temporarily white-listing those
tests.

Managing the Test262 project repository as a git submodule is somewhat
cumbersome for general development, so do not integrate this test in the
default `npm test` command.

Hey @caitp, would you mind taking a look? I'm really excited about this--it
should give us a much clearer understanding of what needs to be done
(especially as more features reach stage 3).


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tests/test262/test262"]
+	path = tests/test262/test262
+	url = https://github.com/tc39/test262
```
caitp: I prefer to add scripts for checking out submodules to package.json (or similar) to make it easier for people to make use of them, since most people don't do recursive checkouts and not everyone is familiar with the git submodule commands.

It would be nice to have something like that here to make this more accessible to people.

tests/test262/interpret-results.js
```diff
@@ -0,0 +1,64 @@
...
+        if (result.parseFailure) {
+          expected.failure.push(result.name);
+        } else {
+          expected.success.push(result.name);
+        }
+      }
+    } else {
+      if (isAllowed) {
+        if (result.parseFailure) {
+          expected.typeI.push(result.name);
```
caitp: Instead of this `typeI` / `typeII` stuff, how about names which indicate what these are? Like, `passes` and `failures` maybe? Much easier to grok what it's saying that way

tests/test262/interpret-results.js
```diff
@@ -0,0 +1,64 @@
...
+    typeII: [],
+    unrecognized: null
+  };
+  var totalUnexpected;
+
+  results.forEach(function(result) {
+    var isAllowed = allowed[result.name];
+    delete allowed[result.name];
+
+    if (!!result.parseFailure === result.expected) {
```
caitp: I feel like the logic of these if statements could be made more concise and easier to read, but it looks okay

tests/test262/interpret-results.js
jugglinmike: Those aren't lazy programmer terms--they come from statistical testing. But
there are plenty of people who aren't familiar with that, so your point is
well-taken. The terms "passes" and "failures" are a little too vague here
because both conditions describe failures. A common alternative is "false
positives" and "false negatives." I probably should have used those from the
start!

.gitmodules
jugglinmike: Great idea! You can now `npm run fetch-test262`.

jugglinmike:
Thanks for the review, @caitp! If my latest changes look good to you, then I can
squash if you wouldn't mind merging for me. Or maybe you can do both from the
GitHub.com UI? I'm not sure.

caitp:
looks ok to me

jugglinmike:
Alrighty. I've squashed and force-pushed (original commits available [here](https://github.com/jshint/jshint/compare/master...jugglinmike:test262-orig))--would you do the honors?

jugglinmike:
@caitp Mind pushing that button for me?

PR #3111, opened 2017-03-14
Greasemonkey/Tampermonkey vars.js

galeksandrp:


jugglinmike:
Thanks for the patch! We appreciate your taking the time to contribute back to
JSHint. Unfortunately, we're trying to move away from supporting
library-specific "environment" options in JSHint "core". JavaScript libraries
come and go, and even the well-established ones are subject to change over
time. When their API changes, it puts JSHint in an awkward position of having
to support multiple versions concurrently. (The libraries we currently support
are effectively "grandfathered in,"--we're taking the maintenance burden to
honor commitments made before this change in policy.)

Issue #3109, opened 2017-03-12
Comparison operators `==`/`!=` reported too aggressively


jugglinmike:
The `eqeqeq` option's description states, "This options prohibits the use of ==
and != in favor of `===` and `!==`." Enabling it will cause JSHint to issue
warning W118 for offending code. However, JSHint will issue a separate warning
(W041) if either side of the comparison is a "falsey" literal value,
*regardless of the value of the `eqeqeq` option*.

**If you are a contributor looking to submit a fix for this bug**, remove W041
altogether. This is a shame because it makes JSHint less strict by default, but
the current behavior is inconsistent and confusing. And while the warning
message for W041 is far more descriptive (and therefore a better choice to keep
around), we can't change W118 in-place outside of a major version release. This
should be relatively straightforward, so I'm marking this issue as a "good
first bug."

**If you are a user struggling with this bug today**, disable W041 in your
project configuration, either with an in-line directive:

```js
// jshint -W041
```

Or within a JSHint configuration file:

```js
{
  "-W041": true
}
```

sunilkumarc:
@jugglinmike I would like to work on this bug.

jugglinmike:
It's all yours, @sunilkumarc!

sunilkumarc:
Hi @jugglinmike. Which version of Node.js should I use to test the project? I have version 6.9.5 installed and the below test is failing when I do `npm test`

![image](https://cloud.githubusercontent.com/assets/3911000/24075216/249f9f2e-0c3d-11e7-93bc-2043e29856dd.png)

jugglinmike:
It looks like you've found a bug in JSHint's browser testing code when run on
Windows. I believe updating to the latest Node packages will resolve the
issue, so I've done that. Could you fetch the latest version of the `master`
branch and try again?

(We're not currently running browser tests in the Windows continuous
integration environment but only because doing so would not surface any new
bugs in JSHint itself. That said, it would help us catch development problems
like this, so I've enabled that as well.)

sunilkumarc:
@jugglinmike It is working now. Thanks a lot for taking care of it.

Btw, this is what I'm thinking of to fix this issue. Search for all the occurrences of *W041* in the project and remove all the code related to this message. Is this fine? Or there will be some issue with this fix?

jugglinmike:
> Thanks a lot for taking care of it.

My pleasure!

> Is this fine? Or there will be some issue with this fix?

That sounds about right to me.

[...]

PR #3114, opened 2017-03-18
Fix Windows issue and improve CI configuration

jugglinmike:
Contributors using Windows and Node.js version 6 experienced runtime errors when attempting to run JSHint's browser tests (reference: https://github.com/jshint/jshint/issues/3109#issuecomment-287568133). This was due to an outdated Node.js dependency.

Update the dependencies to their latest versions, and update the configuration for both Linux and Windows continuous integration environments to run these tests in Node.js version 6.


jugglinmike:
I've merged this myself because it is a low-risk change that's necessary to unblock a new contributor. It's been validated by both continuous integration environments, so I don't expect there to be a problem here.

Issue #2305, opened 2015-04-13
JSHint v2.7.0: ./dist/jshint.js bundles lodash@2.4.1 instead of lodash@3.6.x


segrey:
According to package.json, expected lodash version is 3.6.x:
https://github.com/jshint/jshint/blob/2.7.0/package.json#L44

Actual bundled lodash version is 2.4.1:
https://github.com/jshint/jshint/blob/2.7.0/dist/jshint.js#L1518

Using different lodash versions for node and non-node (browser, rhino) versions might result in bugs that can be reproduced in one environment only and can't be reproduced in others.

segrey:
I think this issue could be closed, because jshint@2.7.0 has already been released. Just wanted to let you know about such an inconsistency.

rwaldron:
Ugh. I forgot to npm install before doing the build. Thanks for the heads up.

segrey:
Great, thanks!

rwaldron:
@segrey this will be fixed shortly.

PR #3118, opened 2017-04-02
[[CHORE]] ./bin/build now runs npm {prune, test, install} prior to bu…

gyaneshanand:
…ilding. Fixes gh-2305

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -3,36 +3,85 @@
 
 "use strict";
 
-var browserify = require("browserify");
-var bundle     = browserify();
-var path       = require("path");
-var version    = require("../package.json").version;
 require("shelljs/make");
 
-var distDir    = path.join(__dirname, "../dist");
-var srcDir     = path.join(__dirname, "../src");
-
-if (!test("-e", distDir))
-  mkdir(distDir);
-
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch! I can see how a more formal automation solution would
help avoid deployment errors like the one you've referenced in the pull request
description. That said, I'm not convinced this is the best way to achieve that.

This logic makes the `build` script far more complex, and the new behavior may
surprise callers who expect "build" to mean "just build." In the past, I have
had to do some modification of dependencies within my local `node_modules`
folder directly during debugging. This would make that workflow impossible.

I think there's a place for this kind of safety logic but that it should be
defined at a higher level. By creating a separate "deploy" script (one that
consumes "build"), we can support the distinct aspects of the workflow.

And although implementing with Node.js is certainly more portable than a shell
script, I would still prefer to see that functionality implemented via Bash.
Reason being: the ratio of maintainers-to-contributors is very low, and I don't
think it makes sense for all contributors to be installing a heavy dependency
like `npm` (50 MB of source code) when they will never need it.

I'd be happy to work with you to build out a solution based on those goals, but
I don't think this particular patch is heading in the right direction.

PR #3123, opened 2017-04-23
Import hoisting

jugglinmike:
Hey @rwaldron: if this looks good to you, can you please merge as individual
commits? I want to make sure we stick to the contribution guidelines for commit
messages, but squashing this pull request via the GitHub.com interface will
create a commit that is invalid by those standards.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4587,7 +4587,7 @@ var JSHINT = (function() {
       this.name = identifier();
       // Import bindings are immutable (see ES6 8.1.1.5.5)
       state.funct["(scope)"].addlabel(this.name, {
-        type: "const",
+        type: "import",
         initialized: true,
         token: state.tokens.curr });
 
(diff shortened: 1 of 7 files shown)


tests/unit/fixtures/latedef-esnext.js
```diff
@@ -67,3 +67,11 @@ export class df {}
 export var dg;
 export let dh;
 export const di = {};
+
+void importedName;
+void importedModule;
+void importedNamespace;
```
rwaldron: 👍

src/jshint.js
```diff
@@ -4640,7 +4640,7 @@ var JSHINT = (function() {
 
         // Import bindings are immutable (see ES6 8.1.1.5.5)
         state.funct["(scope)"].addlabel(importName, {
-          type: "const",
+          type: "import",
```
rwaldron: So much nicer to be able to be explicit like this

jugglinmike: Right on

PR #3124, opened 2017-04-23
[[FIX]] Allow MemberExpression in LHS of `for`

jugglinmike:
With thanks to Test262

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4274,11 +4274,23 @@ var JSHINT = (function() {
         letscope = true;
         state.funct["(scope)"].stack();
         state.tokens.curr.fud({ prefix: true });
-      } else {
-        // Parse as a var statement, with implied bindings. Ignore errors if an error
-        // was already reported
+
+      // Parse the set of tokens before the head's `in` or `of` token as though
+      // they part of a `var` declaration if they satisfy any of the following
+      // conditions:
+      //
+      // - It contains an invalid initializer or comma operator (these errors
+      //   are common enough to warrant dedicated messages which are emitted
+      //   above)
+      // - It contains a single token
+      // - It describes a destructuring binding pattern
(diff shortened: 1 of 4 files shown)


src/jshint.js
```diff
@@ -4274,11 +4274,23 @@ var JSHINT = (function() {
...
+      // they part of a `var` declaration if they satisfy any of the following
+      // conditions:
+      //
+      // - It contains an invalid initializer or comma operator (these errors
+      //   are common enough to warrant dedicated messages which are emitted
+      //   above)
+      // - It contains a single token
+      // - It describes a destructuring binding pattern
+      } else if (initializer || comma || peek() === nextop ||
+        checkPunctuators(state.tokens.next, ["{", "["])) {
```
rwaldron: I just read this wrong several times before I realized that `checkPunctuators(state.tokens.next, ["{", "["])` was part of the condition expression

jugglinmike:
> I just read this wrong several times before I realized that
> `checkPunctuators(state.tokens.next, ["{", "["])` was part of the condition
> expression

I'm glad you said something about this. The quick fix was to assign the result
of this expression in a variable and then reference that variable in the
conditional.

...But I have to admit that I've always hated this particular hack. Re-using
the `var` statement's parsing logic requires bizarre extensions to the
declaration and even then, it produces incorrect results.

In its current form, this patch avoids that strange code path for some limited
cases. But as long as we use that code path at all (i.e. as long as we invoke
`varstatement.fud`), then the `var` parsing logic still has a lot of baggage.

More specifically, this patch adds support for

    for (x.y in {}) {}

...but in keeping the "old and busted" path for destructuring (as implemented
via the oddly-formatted condition you called out), it is *still* incorrect for
this input:

    for ([x.y] in {}) {}

So I've opened a new version of this patch that does away with this completely:
gh-3126

PR #3126, opened 2017-04-30
[[FIX]] Parse `for-in/of` head LHS as asnmt target

jugglinmike:
Previously, the left-hand side of the `for-in/of` head re-used the code
path for `var` declaration lists. This was undesirable for a number of
reasons:

- it made the `var` statement's parsing logic more complex
- it tended to produce inaccurate error messages, e.g. "destructuring
  binding is available in ES6"
- it disallowed syntactically valid code, e.g. `for (a.b in {}) {}` or
  `for ([x[y]] of []) {}`

Refactor the parsing logic of `for-in/of` heads to more faithfully match
the actual language grammar. Update the expected errors in the project's
unit test suite according to the more accurate messages that this change
produces.  Simplify `var` statement parsing logic by removing the code
intended to support this condition. Maintain support for initializers
and the comma operator in order to improve error messages in the case of
common programming mistakes.


rwaldron:
Thanks for walking me through this in person :D

PR #3128, opened 2017-05-06
Report invalid syntax as an error (not a warning)

jugglinmike:
@rwaldron I had to work to convince myself that this change was valid--that the
condition previously reported with a "warning" really *is* an error. This was
partially because the `advance` function is so poorly documented. So this patch
begins with a documentation commit which is unrelated to the change, but that I
hope will make the source more clear. It's the second commit that actually
implements the change.


jugglinmike:
This change may actually be somewhat controversial. We should discuss this before landing--it might be better to vet it with an RC for the next minor release. @rwaldron let's talk when you have a few minutes.

jugglinmike:
Rick and I didn't get to talk face-to-face, which is probably for the best--I
ought to explain myself more publicly. Here goes:

My concern revolved around a detail that I noted in the commit message:

> This change triggers failures in four Test262 tests that were previously
> interpreted as "valid." These tests concern the "object spread/rest"
> proposal, but support for this proposal has not yet been implemented in
> JSHint. Therefor, their previous "passing" status was circumstantial,
> and the newly-identified error should be marked as "expected."

It is possible that some users are relying on the buggy behavior, believing it
to be intentional support of the "object spread/rest" proposal. Although we are
technically justified in removing this "psuedo-support" for ES6 code, I have
sympathy for users that aren't aware of the various proposals' place within the
standardization process.

My current thinking is that we ought to correct the bug, but we should only do
so if we can also provide a way for users in this position to continue to use
JSHint with minimal interruption. If we defer this to the next minor version of
JSHint, we can release the fix alongside an experimental linting option for the
"object spread/rest" proposal. This approach may still catch users in this
position off guard, but at least we will be able to offer a viable path
forward: enable the new option.

To that end, I'm closing this pull request and opening a new one against the
`v2.10.0` branch.

PR #3140, opened 2017-06-03
Report errors

jugglinmike:
This is a rebased version of gh-3128, where I wrote:

> @rwaldron I had to work to convince myself that this change was valid--that the
> condition previously reported with a "warning" really is an error. This was
> partially because the advance function is so poorly documented. So this patch
> begins with a documentation commit which is unrelated to the change, but that I
> hope will make the source more clear. It's the second commit that actually
> implements the change.

...though I ultimately decided that this was better to defer until the next minor release of JSHint (see gh-3128 for my rationale).


rwaldron:
Excellent

PR #3142, opened 2017-06-10
Fix spread/rest

jugglinmike:
@rwaldron This builds on https://github.com/jshint/jshint/pull/3140. Like that
patch, it introduces new syntax errors that are valid according to the "object
spread/rest proposal." In the same vein (and as noted in the relevant commit
message), I don't think this is technically a breaking change, but it may
surprise users who relied on the behavior without realizing that it was a bug.
So I think this is ready for the `v2.10.0` branch, but to be considerate, we
should avoid releasing this without exposing opt-in support for that proposal.

Resolves gh-2525.


jugglinmike:
This update does not account for that development because this is limited to the ES2015 feature. I intend to introduce parsing support for "object rest/spread" in another patch; getting this correction in place is (in my mind) a necessary precursor.

tests/unit/parser.js
```diff
@@ -7807,21 +7808,72 @@ exports.restOperatorWithoutIdentifier = function (test) {
...
 
   test.done();
 };
 
+exports.invalidSpread = function (test) {
+  TestRun(test)
+    .addError(1, "Expected an identifier and instead saw '...'.")
+    .addError(1, "Missing semicolon.")
+    .addError(1, "Expected an assignment or function call and instead saw an expression.")
+    .test('void ...x;', { esversion: 6 });
```
rwaldron: 👍

rwaldron:
I don't know... I'm concerned about that -0.03% coverage loss...

;)

PR #3144, opened 2017-06-12
        modified:   src/jshint.js

FuDesign2008:
qq20170612-181935 2x

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -529,7 +529,7 @@ var JSHINT = (function() {
             val = +val;
 
             if (typeof val !== "number" || !isFinite(val) || val <= 0 || Math.floor(val) !== val) {
-              error("E032", directiveToken, g[1].trim());
+              error("E032", directiveToken, (g[1] || "").trim());
               return;
             }
 


rwaldron:
Can you provide more insight into how this occurred?

jugglinmike:
Working backwards from the modified statement, I've identified the problem this patch addresses. Example input:

    // jshint maxstatements

It's certainly invalid to declare a value-accepting option without a value, but users with JSHint editor integrations are likely to transition through this state as they type. JSHint shouldn't crash here, for sure, and this patch addressed the problem.

A few things, though:

- We should include a test to demonstrate that this addresses the problem
- We'll need your signature on the project's CLA. This might sound like a
  hassle, but it only takes a minute or so.

We still haven't heard back from @FuDesign2008, but I'm hoping that with this more specific feedback, they might be able to finish up this patch. Otherwise, I'll submit a version of my own in a few days.

caitp:
@FuDesign2008 ping, can you write a unit test and sign the CLA?

PR #3158, opened 2017-07-16
[[CHORE]] Fix linting error and enable lint in CI

jugglinmike:
A previous commit removed all references to an internal function,
introducing a linting error. This was not reported by the project's
continuous integration system because that system was not configured to
run the project's linting check.

Correct the linting error and enable linting in the continuous
integration environment.


rwaldron:
LGTM

PR #3150, opened 2017-06-23
Unstable opts: object spread/rest

jugglinmike:
As the first "unstable" option, this changeset includes a commit that implements the internals necessary to support options of that type (resolving gh-2601). The second commit introduces the first unstable option--experimental support for [the object spread/rest](https://github.com/tc39/proposal-object-rest-spread) proposal (resolving gh-2991).

PR #3159, opened 2017-07-16
[[FIX]] Update parsing of object "rest" property

jugglinmike:
The latest version of the proposal disallows BindingPattern as the
right-hand side of a "rest" property.

https://github.com/tc39/proposal-object-rest-spread/pull/50
http://tc39.github.io/tc39-notes/2017-05_may-23.html#16ih-why-allow-bindingpattern-for-bindingrestparameter-for-object-rest-maybe-we-should-just-allow-identifiers

---

@rwaldron I wrote this in response to your initial review of gh-3150, but I
stopped short of pushing it up because the correction uncovered an unrelated
bug. Instead, I submitted gh-3157 to fix that bug. I was planning on updating
the first patch after the second landed, but the ordering of these changes
isn't all that important. (It's probably for the best, anyway--I only just
today realized that my original attempt to incorporate your feedback was
flawed.)


rwaldron:
Regarding the list of files that starts here: https://travis-ci.org/jshint/jshint/jobs/254043034#L1029...

Those tests have been removed from test262, for the same reason that this PR exists. So it's actually valuable to see that linting the source in those files with JSHint will now result in errors. I don't think that those failures should block this PR, but maybe we should discuss that further?

jugglinmike:
I've fixed the build, Rick!

rwaldron:
Great work, thank you!

tests/test262/expectations.txt
```diff
@@ -4067,3 +4067,137 @@ language/statements/class/meth-static-dflt-params-ref-self.js
...
+language/statements/class/dstr-meth-obj-ptrn-rest-nested-obj.js
+language/statements/class/dstr-meth-obj-ptrn-rest-obj-nested-rest.js
+language/statements/class/dstr-meth-obj-ptrn-rest-obj-own-property.js
+language/statements/class/dstr-meth-static-dflt-obj-ptrn-rest-nested-obj.js
+language/statements/class/dstr-meth-static-dflt-obj-ptrn-rest-obj-nested-rest.js
+language/statements/class/dstr-meth-static-dflt-obj-ptrn-rest-obj-own-property.js
+language/statements/class/dstr-meth-static-obj-ptrn-rest-nested-obj.js
+language/statements/class/dstr-meth-static-obj-ptrn-rest-obj-nested-rest.js
+language/statements/class/dstr-meth-static-obj-ptrn-rest-obj-own-property.js
+# End invalid "object spread/rest property" tests
```
rwaldron: 👍

PR #3160, opened 2017-07-16
[[FIX]] Allow computed method names in obj literal

jugglinmike:
A quick fix for a bug identified by Test262.


rwaldron:
Excellent!

PR #3162, opened 2017-07-19
Fix parsing of 'in' operator

jugglinmike:
@rwaldron I know this is a lot of change for what seems like a pretty small
edge case, but (beyond making the codebase more true to the specification) it
helps clear the way for proper implementation of async functions (and
improvements to support for `yield`).

It's not a new feature by any stretch, so it could technically land in `master`
for our next patch release. But given all the change, I figured it would be
better to merge it here, since we will (presumably) vet 2.10.0 with a release
candidate. That's a good opportunity to verify that this churn doesn't have any
unintended side effects.


jugglinmike:
@rwaldron and I discussed this patch in person, and he recommended that we
re-order the arguments to `led` and `nud` methods to accept the new "context"
bitfield as the first parameter. His case was focused around consistency, and I
agreed with him.

This concerned two separate-but-related commits, and as such, I couldn't find a
way to avoid rebasing. To maintain some level of transparency, I've introduced
the required changes as `fixup!` commits. Hopefully this is helpful during
review.

This has been a positive change for the codebase overall, as it forces the new
parameter to be specified consistently. Previously, many call sites got away
with leaving it out altogether, and the `undefined` value was subsequently
interpreted as `0`, or an empty bitfield.

Unfortunately, this has also caused the `block` function's signature to grow
from five to six parameters. Rick was already reluctant about the extra
parameter, and I doubt this new development excites him very much. We could
mitigate this by specifying some of that function's boolean parameters as
further members of the "context" bitfield, but if there's any interest in doing
so, I would request that we consider such a refactoring in a follow-up patch.

Issue #3148, opened 2017-06-22
Expected a 'break' statement before 'default' is not symmetric


thomas658:
The warning "Expected a 'break' statement before 'default'" is emitted for line 6 in
```
var x;
switch (x) {
  case 1:
    x++;
    break;
  case 2:
  default:
    x--;
    break;
}
```
but not for
```
var x;
switch (x) {
  case 1:
    x++;
    break;
  default:
  case 2:
    x--;
    break;
}
```
In my opinion these two snippets have the same functionality, the order of `case 2:` and `default:` do not matter.

jugglinmike:
Thanks for the report! This is a tricky one.

If the argument were solely about consistency, then the resolution would be to
disable the warning in your first example. That's because JSHint currently
allows

    switch (0) {
      case 0:
      case 1:
        break;
    }

...so why should the following be any different?

    switch (0) {
      case 0:
      default:
        break;
    }

However, I think there's good reason to warn about the second usage. There, the
`default` clause renders the empty `case` clause completely superfluous. There
isn't much reason to include it at all; it's more likely an indication of an
erroneous refactoring.

What that means is that your second example, simplified here:

    switch (0) {
      default:
      case 0:
        break;
    }

...really ought to produce a warning. That `case` statement is no less
superfluous here.

This is a new warning for a condition that is not well documented, so I'm
reluctant to implement it outside of a major release. Then again, the message,
"Expected a `break` statement before `case`" is not really describing the
problem in enough detail, anyway.

Here's what I'd like to propose:

- We disable the existing warning (code W086) for your first example
- We introduce a new warning (e.g. "Superfluous case clause.") and a new
  linting option (e.g. `leanswitch`) to control it (disabled by default)
- We enable the new option by default, in the next major release of JSHint

What do you think? (also @rwaldron)

thomas658:
That sounds very sensible.

Btw, the reason for the construct
```
switch (0) {
  case MY_SYMBOL:
  default:
    break;
}
```
was not refactorcing but documenting the default behaviour in the (minimized) code.

jugglinmike:
Excellent. I've opened https://github.com/jshint/jshint/pull/3165 to implement the new feature in the next minor release.

jugglinmike:
Good news! The `leanswitch` option is available in [JSHint version 2.10.0](https://jshint.com/blog/2019-02-05/release-2-10-0/) released earlier this week.

PR #3165, opened 2017-07-22
[[FEAT]] Introduce linting option `leanswitch`

jugglinmike:


jugglinmike:
This should resolve gh-3148.

rwaldron:
Seems reasonable and looks good to me

PR #3168, opened 2017-08-06
[[DOCS]] Improve documentation of internals

jugglinmike:
Extend in-line documentation to help future contributors orient
themselves to the code base. Avoid any modification to executable code
in order to ensure that program behavior is unchanged.

---

Hey @rwaldron,

I've been meaning to do this for quite some time now; I'm hoping it will help
more folks get involved in the project. It's probably not as complete as it
could be, but it's a start. As noted in the commit message, I held myself back
from making any changes to the code itself. Rest assured, though: this exercise
has given me plenty of ideas about refactorings. I just figured it would be
better to do that with a follow-up patch.


rwaldron:
This excellent; a lot of reading, but worth every line.

jugglinmike:
Thanks a lot Rick!

PR #3169, opened 2017-08-12
[[FIX]] Do not create binding for illegal syntax

jugglinmike:
When a function declaration without an identifier is encountered, JSHint
issues a warning and attempts to continue parsing the declaration.

Previously, JSHint created a new binding for the "function" keyword, and
because such a binding can never be referenced, this would generate
spurious warnings when the "unused" option was enabled.

Do not create a faulty binding in the presence of this syntax error.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3909,14 +3909,16 @@ var JSHINT = (function() {
     }
     var i = optionalidentifier();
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks :)

PR #3171, opened 2017-08-17
Update generate-identifier-data.js

aqibbak:


jugglinmike:
Can you explain the motivation for this change?

PR #3174, opened 2017-08-20
[[CHORE]] Disallow ambiguity in test suite

jugglinmike:
Some tests cause the same error to be reported for different characters
on the same line. Previously, a single expectation that did not specify
a character number would satisfy all errors with the same message on the
given line.

Update the test harness to disallow the specification of identical
expected errors and to report when errors are triggered at a character
other than what is expected. Update all tests that previously relied on
this ambiguity to formally encode the expected character number and
remove the expectations that were superfluous.

---

This depends on gh-3173.


tests/helpers/testhelper.js
```diff
@@ -41,6 +41,27 @@ exports.setup.testRun = function (test, name) {
...
+          if (err.line !== line) {
+            return false;
+          }
+
+          if (extras && err.character !== extras.character) {
+            return false;
+          }
+
+          return true;
+        });
```
rwaldron: nit: looks like this is indented two spaces too far?

jugglinmike:
Incorporated your feedback (that was my personal coding style creeping in) and rebased as requested.

PR #3175, opened 2017-08-20
[[FIX]] Allow closing over immutable bindings

jugglinmike:
The risk of confusion for functions declared within loops only pertains
to references to mutable bindings. Relax the wanting regarding
"confusing semantics" to only occur in the presence of one or more
references to mutable bindings.


rwaldron:
Nice work!

Issue #3151, opened 2017-06-23
Tab indentation breaks multiple rules


Arcanemagus:
It seems that indentation using tabs breaks the position reported by at least 3 rules, likely more.

All of the following was linted using this `.jshintrc` file:
```json
{
  "strict": "global",
  "unused": true
}
```

The JSHint version tested is v2.9.5.

If you want me to refile these as separate issues I can do that as well.

## W032

If you lint this code:
```js
'use strict';

	function foo() {
	};

foo();
```

There is an error reported on line 4, col 6, but line 4 only has 4 characters in it.

## W098

If you lint this code:
```js
'use strict';

	function foobar(
		$foo
	) {
		return 'foo';
	}

foobar();
```

There is an error reported on line 4, col 9, but line 4 only has 7 characters in it.

## W117

If you lint this code:
```js
'use strict';

	function foobar() {
		if (true) {
			fun1();
		}
	}

foobar();
```

There is an error reported on line 5, col 13, but line 4 only has 11 characters in it.

----

The character position reported seems to go further and further out the more tabs there are at the beginning of the line in question.

Raw files can be found here: [linter-jshint_GH416.zip](https://github.com/jshint/jshint/files/1098846/linter-jshint_GH416.zip)

Originally discovered while investigating https://github.com/AtomLinter/linter-jshint/issues/416.

jugglinmike:
Thanks for the report! The problem stems from [line 1610 of `lex.js`](https://github.com/jshint/jshint/blob/d3d84ae1695359aef077ddb143f4be98001343b4/src/lex.js#L1610):

    this.input = this.input.replace(/\t/g, state.tab);

This seems pretty fundamental to how many of the style-related linting rules are implemented. Those have been deprecated for some time now, but we're still not ready to move on a new major version. So in the mean time, we'll need to design a solution that preserves that behavior.

My initial thought is to track this "effective offset" as a new property of the token object named `column`. This can be used for those style-related warnings. Then, the `character` attribute could be re-implemented to describe the "true" character offset, where each code point (tab or otherwise) increments the count by exactly 1. This value would be used to issue warnings. Reporter plugins would then be in control of how they rendered the tab character, and they could interpret the reported "column" number accordingly.

Does that sound good to you, @Arcanemagus?

Arcanemagus:
That sounds perfect to me, at least on the surface with no real knowledge of JSHint's internals 😛. Ideally I just need some method of getting the true column count (where each code point counts as one column).

[...]

PR #3176, opened 2017-08-26
[[TEST]] Require all tests to specify char no.

jugglinmike:
The character of reported warnings/errors is of critical importance to
tooling built on top of JSHint. These values should be accurate, and
they should be stable between releases of the project.

Update the project's "test" API to require explicit specification of
expected character number. Update each test to provide this value.

Note: due to a deficiency in the test harness, lines that emitted
multiple instances of the same error/warning (with distinct line
numbers) would be satisfied by a single assertion for the relevant
error/warning. This change exposed that deficiency and necessitated the
introduction of additional assertions for such cases, e.g. the following
modification to `tests/unit/options.js`:

    -    .addError(11, "'j' used out of scope.") // 3x
    +    .addError(11, 14, "'j' used out of scope.")
    +    .addError(11, 21, "'j' used out of scope.")
    +    .addError(11, 28, "'j' used out of scope.")

---

@rwaldron: this isn't just pedantry :P The additional rigour really is
important for integration with code editors; see gh-3151. This patch is a
follow-up to gh-3174, and both should help us avoid regressions, which will be
especially if/when we try to address the problem reported in gh-3151.

To be clear: this will break every existing pull requests that introduces new
tests. I think this is acceptable because most of them already require
additional work before they can be landed (either due to review feedback or,
unfortunately, due to bit rot).


rwaldron:
This is excellent!

jugglinmike:
Thanks, Rick!

PR #2584, opened 2015-07-25
increase code coverage

lukeapage:
I'm splitting into seperate commits, feel free to review seperately.

PR #3177, opened 2017-08-27
Fixes: inc/dec and bindings

jugglinmike:
@rwaldron since this updates a few tests, it's going to suffer from conflicts
if gh-3176 is merged first. I'm happy to rebase at that time, but I'm holding
off for now because I don't want to presume that other patch is going to land
at all (since it's a little controversial).

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3019,6 +3019,10 @@ var JSHINT = (function() {
         classExprBinding ? "class" : "function", state.tokens.curr, false);
     }
 
+    if (!isArrow) {
+      state.funct["(scope)"].funct.add("arguments", "var", token, false);
+    }
+
     // create the param scope (params added in functionparams)
     state.funct["(scope)"].stack("functionparams");
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1365,7 +1365,8 @@ var JSHINT = (function() {
       }
 
       return true;
-    } else if (left.identifier && !isReserved(left) && !left.isMetaProperty) {
+    } else if (left.identifier && !isReserved(left) && !left.isMetaProperty &&
+      left.value !== "eval" && left.value !== "arguments") {
       if (state.funct["(scope)"].labeltype(left.value) === "exception") {
         warning("W022", left);
       }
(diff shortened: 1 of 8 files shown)


jugglinmike:
Hm, looks like this fix introduces a kind of regression: it makes JSHint warn that `o["eval"]` is "better written in dot notation." That's true--`o.eval` is valid even in ES3 environments where `o.if` is a SyntaxError. `o.arguments` is likewise valid in all environments.

Still, the conservative thing to do would be to allow that particular form. I'll add a `fixup!` patch to allow those forms (along with a note to remove the allowance in JSHint 3).

jugglinmike:
Alrighty, that "fixup" commit is up!

rwaldron:
And of course I reviewed that one first... Sorry about that

jugglinmike:
No worries; that's the sequence I had in mind. I'll get to rebasing this ASAP.

jugglinmike:
Alrighty, I've rebased, squashed, and force-pushed the result. The original version of this branch is available here:

https://github.com/jshint/jshint/compare/master...jugglinmike:built-ins-3-prebase

jugglinmike:
I've rebased one more time to fix conflicts resulting from today's merge of https://github.com/jshint/jshint/pull/2584

jugglinmike:
Well that was quite a bit trickier than I expected. @rwaldron: to properly satisfy the new tests, I had to re-implement `arguments`. It's for the best, though, since the prior implementation defined `arguments` as a global variable. So the implementation is more technically correct, but there's an observable difference, too: we now catch problems relating to `arguments` being referenced within arrow functions. I've pushed those changes in a "fixup" commit; happy to rebase once you've had a chance to review.

jugglinmike:
This depends on gh-3181, so I'm going to close it until that lands to avoid confusion.

rwaldron:
Hi @jugglinmike! I've landed the blocking PR and re-opened this PR.

jugglinmike:
Okay @rwaldron, I'll admit it. Ever since we saw a string of false positives
from Coveralls, I haven't been paying as much attention the its reports as I
should.  Many of the "uncovered" lines originally reported were actually
covered by Test262 (though I'm thinking about correcting that in the future).
That said, one of the lines was a condition within the `optionalidentifier`
function that this patch made unreachable. By removing it, we're able to remove
a parameter from both `optionalidentifier` and `identifier`. That's done in the
latest `fixup!` commit. An unexpected simplification, and a lesson to pay more
attention to our CI.

jugglinmike:
@rwaldron This should be good to land.

PR #3178, opened 2017-08-30
Unused variables deletion

jshaptic:
Noticed a lot of variables and properties, that are initialized, but not used anywhere in the code. Can be safely deleted.

PR #3180, opened 2017-09-01
[[FIX]] Fixing typos and some dead old code

jshaptic:
Removed old "functionicity" variable, that was used only in useless "for" cycles, which did nothing but adding and removing the same properties to the "fu" object over and over again.

Also, noticed several typing errors like "tokens.current" instead of "tokens.curr"; "v.raw_text" instead of "v.token.raw_text" etc.

Also, property "tokens[0].first" is not used anywhere in the code and can be removed.

And the last thing, function classbody doesn't return anything, so there is no point to assign undefined value to the "c.body" property.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -5043,7 +5043,7 @@ var JSHINT = (function() {
         unstack: function() {
           _current.variables.filter(function(v) {
             if (v.unused)
-              warning("W098", v.token, v.raw_text || v.value);
+              warning("W098", v.token, v.token.raw_text || v.value);
             if (v.undef)
               state.funct["(scope)"].block.use(v.value, v.token);
           });
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3549,9 +3549,7 @@ var JSHINT = (function() {
         if (!prefix && value && value.type === "undefined") {
           warning("W080", id, id.value);
         }
-        if (lone) {
-          tokens[0].first = value;
-        } else {
+        if (!lone) {
           destructuringPatternMatch(names, value);
         }
       }
@@ -3661,9 +3659,7 @@ var JSHINT = (function() {
         if (value && !prefix && !state.funct["(loopage)"] && value.type === "undefined") {
           warning("W080", id, id.value);
         }
-        if (lone) {
-          tokens[0].first = value;
-        } else {
+        if (!lone) {
           destructuringPatternMatch(names, value);
         }
       }
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3724,7 +3724,7 @@ var JSHINT = (function() {
     state.inClassBody = true;
     advance("{");
     // ClassBody(opt)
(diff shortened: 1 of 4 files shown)


jugglinmike:
These changes look good, but there's a lot packed in to a single commit. I appreciated the granularity of on your recent contribution in gh-3178. Here, I think that same granularity is even more important because part of this changeset affects observable behavior. Would you mind factoring this patch into a few independent commits?

jshaptic:
I have split these changes into several independent commits. I hope I did it right (not a git expert), please let me know if something is wrong.

However for some reason Travis build failed - log is just empty and I cannot understand why. Is it me or Travis just failed for no apparent reason?

src/jshint.js
```diff
@@ -3014,17 +3009,17 @@ var JSHINT = (function() {
...
+   * @param {token} [options.statement] The statement that triggered creation
    *                                      of the current function.
    * @param {string} [options.type] If specified, either "generator" or "arrow"
    * @param {token} [options.loneArg] The argument to the function in cases
    *                                  where it was defined using the
    *                                  single-argument shorthand
    * @param {bool} [options.parsedOpening] Whether the opening parenthesis has
    *                                       already been parsed
-   * @param {token} [options.classExprBinding] Define a function with this
+   * @param {string} [options.classExprBinding] Define a function with this
```
jugglinmike: The changes to this comment are good, but they subvert the formatting convention. Could you also update the indentation of the text in the subsequent lines?

src/jshint.js
```diff
@@ -5550,16 +5541,6 @@ var JSHINT = (function() {
...
-      for (j = 0; j < functionicity.length; j += 1) {
-        fu[functionicity[j]] = [];
-      }
-
-      for (j = 0; j < functionicity.length; j += 1) {
-        if (fu[functionicity[j]].length === 0) {
-          delete fu[functionicity[j]];
-        }
-      }
-
```
jugglinmike: Quite silly.

src/jshint.js
```diff
@@ -3789,7 +3780,7 @@ var JSHINT = (function() {
     var props = Object.create(null);
     var staticProps = Object.create(null);
     var computed;
-    for (var i = 0; state.tokens.next.id !== "}"; ++i) {
+    while (state.tokens.next.id !== "}") {
```
jugglinmike: Nice catch!

src/jshint.js
jugglinmike: (That is: the code you removed is silly. The change itself is great!)

jshaptic:
Looks like I accidently deleted previous commits, while playing with git... Anyway I've made new commits with corrected [[FIX]] messages and also fixed indentation to follow formatting conventions.

jugglinmike:
Verified that the reported decrease in coverage is due to the removal of superfluous code. Thanks for this patch--you've done a great job finding subtle mistakes!

PR #3185, opened 2017-09-10
[[DOCS]] Expand warning message

jugglinmike:
`export` declarations with empty export export clauses are not
necessarily superfluous: if they also specify a "from" clause, they will
cause the referenced module to be evaluated. Because of this, the
warning produced in these cases should be contextual--it should
recommend removal only if no "from" clause is present. In the presence
of such a clause, the warning should instead suggest use of a
binding-free `import` declaration (as this clearly communicates the
intention to evaluate a module without requesting any bindings).

Notes on compatibility:

- The previous warning behavior had not yet been published in a stable
  release (it was implemented at the same time as the parser was
  corrected to support `export` declarations without bindings). This
  makes this change in warning behavior backward-compatible with prior
  releases.
- This warning is relevant for `import` declarations as well. However,
  because prior releases of JSHint included correct parsing support for
  binding-free `import` declarations, introducing an equivalent warning
  would constitute a backwards-breaking change.

@rwaldron This is a follow-up to gh-2569 which I just landed today.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3412,7 +3412,7 @@ var JSHINT = (function() {
         advance("[");
       }
       if (checkPunctuator(state.tokens.next, "]")) {
-        warning("W137", state.tokens.curr, "destructuring");
+        warning("W137", state.tokens.curr);
       }
       var element_after_rest = false;
       while (!checkPunctuator(state.tokens.next, "]")) {
@@ -3444,7 +3444,7 @@ var JSHINT = (function() {
         advance("{");
       }
       if (checkPunctuator(state.tokens.next, "}")) {
-        warning("W137", state.tokens.curr, "destructuring");
+        warning("W137", state.tokens.curr);
       }
       while (!checkPunctuator(state.tokens.next, "}")) {
         assignmentProperty();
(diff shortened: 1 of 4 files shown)


src/jshint.js
```diff
@@ -4676,13 +4672,23 @@ var JSHINT = (function() {
...
     advance("from");
     advance("(string)");
+
+    // Support for ES2015 modules was released without warning for `import`
+    // declarations that lack bindings. Issuing a warning would therefor
+    // constitute a breaking change.
+    // TODO: enable this warning in JSHint 3
+    // if (hasBindings) {
+    //   warning("W142", this, "import", moduleSpecifier);
+    // }
```

src/messages.js
```diff
@@ -222,10 +222,12 @@ var warnings = {
...
   W135: "{a} may not be supported by non-browser environments.",
   W136: "'{a}' must be in function scope.",
-  W137: "Empty {a}: this is unnecessary and can be removed.",
+  W137: "Empty destructuring: this is unnecessary and can be removed.",
   W138: "Regular parameters should not come after default parameters.",
   W139: "Function expressions should not be used as the second operand to instanceof.",
-  W140: "Missing comma."
+  W140: "Missing comma.",
+  W141: "Empty {a}: this is unnecessary and can be removed.",
+  W142: "Empty {a}: consider replacing with `import '{b}';`."
```
rwaldron: 👍

PR #3191, opened 2017-09-16
[[FIX]] Consistently ignore dot-prefixed dirs

jugglinmike:
When the Node.js CLI is invoked with a path to an input file, that file
should not be linted if it matches any pattern in a "ignore list"
configuration file.

Prior to this patch, JSHint failed to ignore such files if one of their
parent directories was prefixed with a period character.

Update the CLI's usage of the `minimatch` Node.js library to also
consider such "dot-prefixed" directories when interpreting ignore lists.

Note that this change is backwards-compatible because it limits the
situations in which the CLI reports an error.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -230,7 +230,7 @@ function loadIgnores(params) {
  */
 function isIgnored(fp, patterns) {
   return patterns.some(function(ip) {
-    if (minimatch(path.resolve(fp), ip, { nocase: true })) {
+    if (minimatch(path.resolve(fp), ip, { nocase: true, dot: true })) {
       return true;
     }
 
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -658,6 +658,30 @@ exports.group = {
     test.done();
   },
 
+  // See gh-3187
+  testIgnoreWithDot: function (test) {
+    var dir = __dirname + "/../examples/";
+    this.sinon.stub(process, "cwd").returns(dir);
+
+    this.sinon.stub(shjs, "cat")
+      .withArgs(sinon.match(/file\.js$/)).returns("This is not Javascript.")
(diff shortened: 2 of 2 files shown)


rwaldron:
@jugglinmike take a look at appveyor when you have a chance

jugglinmike:
@rwaldron I diagnosed the problem shortly after submitting this patch. It's a Node.js version 0.10 compatibility issue in one of our dependencies. I've offered the maintainers a fix here:

https://github.com/maxogden/extract-zip/pull/51

No response yet, though. Sorry for not mentioning that sooner!

rwaldron:
Should we hold off until that's resolved?

rwaldron:
(Node.js 0.10 is long unsupported by Node.js)

jugglinmike:
> Should we hold off until that's resolved?

Yeah, let's give it a few more days. Unfortunately, this is going to fail on
all builds until we resolve the problem or delete the "Windows/Node 0.10" job.

> (Node.js 0.10 is long unsupported by Node.js)

That occurred to me, too. As I wrote in the pull request description, I was
hoping that since supporting that runtime was trivial enough, maybe the
maintainers wouldn't mind taking it on.

jugglinmike:
@rwaldron Since you merged gh-3215, I re-started the build, and everything is green. Hooray!

rwaldron:
Excellent!

jugglinmike:
Thanks!

Kristinita:
**Type: Problem** :crying_cat_face:

It seems, that if user want to lint files with names beginning with a dot, he need manually add these files to the command as positional arguments. Example:

```shell
jshint . .pa11yci.js .simple-git-hooks.js
```

I can’t find the option or command-line argument like `"lint-dot-prefixed-files": true`. Templates as `!.*.js` in the file `.jshintignore` doesn’t help.

Thanks.

Issue #3194, opened 2017-09-25
Feature request: add intersection observer API globals to browser environment


cueedee:
http://caniuse.com/#feat=intersectionobserver

jugglinmike:
Sounds good to me. As near as I can tell, that specification introduces a single global binding: [`IntersectionObserver`](https://www.w3.org/TR/intersection-observer/#intersectionobserver). If so, this should be straightforward to implement. New contributors looking to get their bearings could use the recently-landed gh-3190 as a guide. I'd be happy to give more guidance, of course.

@cueedee Do you have any interest in helping out?

johnmryan:
I'd be interested in helping out if you're still looking for a contributor here

jugglinmike:
Resolved via gh-3194.

PR #3199, opened 2017-10-10
[[FEAT]] Add support for IntersectionObserver in global vars

johnmryan:
Adding support for IntersectionObserver global: https://github.com/jshint/jshint/issues/3194

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -162,6 +162,7 @@ exports.browser = {
   HTMLVideoElement     : false,
   history              : false,
   Image                : false,
+  IntersectionObserver : false,
   Intl                 : false,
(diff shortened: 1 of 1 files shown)


johnmryan:
`phantomjs` is not a fan of my commit according to `appveyor/pr`

jugglinmike:
The failure does not reflect a problem with this change set. It's a new issue with a module we depend on transitively, so addressing it properly requires some coordination between projects.

While we work that out, I'm happy to ignore the failure and accept this change. Merged to `master` at commit c7c5c42885aa618e384de3f6589793fd419de7c1. Thanks for your help!

Issue #3172, opened 2017-08-19
Whitespace other than spaces can break inline rule parsing


morganwdavis:
Consider this inline rule layout:
```
/* jshint
	strict:		implied,
	unused:		true,
	undef:		true,
	curly:		true,
	browser:	true,
	jquery:		true,
	esversion:	6
*/
```
If there is a tab between the `/*` and `jshint`, the entire rule is ignored (not discovered during parsing).  The layout above works as long as the only intervening whitespace is one or more spaces. Fortunately, the individual directives on subsequent lines can be preceded by tabs without trouble.

jugglinmike:
Confirmed. Thanks for the report!

For anyone looking to help out (might that be you, @morganwdavis?): The issue
originates in `lex.js`. The code for tokenizing a comment specifically looks
for comments that start with the character sequence `jshint` but also accepts a
single leading space, as in ` jshint`. I can't think of a reason why it should
be so specific, and I doubt that making it tolerate any number of spaces/tabs
would cause any problems in existing code (careful not to accept newlines,
though). The only "gotcha" is implementing this in an efficient way--remember
that this code path is hit whenever JSHint sees any comment!

morganwdavis:
Mike, my intent was to try this in Visual Studio Code after recently converting from SublimeText where I used jshint. I've since converted a large project over and switched to eslint which seems to suit me better. Fixing this bug will need to fall to someone else with more of an interest in this project.

johnmryan:
Will look into this.

nazarioa:
This should be closed on account of https://github.com/jshint/jshint/pull/3201

jugglinmike:
You're right, @nazarioa --thanks!

PR #3201, opened 2017-10-11
[[FIX]] Allow multiple spaces or tabs before inline rule

johnmryan:
In reference to: https://github.com/jshint/jshint/issues/3172

Also included my global vars addition from previous pull request, let me know if you'd like me to rip that out. 

Will work on some examples to throw in the unit testing.

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -438,7 +438,7 @@ Lexer.prototype = {
         // To handle rarer case when special word is separated from label by
         // multiple spaces or tabs
         var strIndex = body.indexOf(str);
-        if (!isSpecial && strIndex >= 0 && body.charAt(strIndex + str.length + 1) === " ") {
+        if (!isSpecial && strIndex >= 0 && body.charAt(strIndex + str.length) === " ") {
           var isAllWhitespace = body.substr(0, strIndex).trim().length === 0;
           if (isAllWhitespace) {
             isSpecial = true;
--- a/src/lex.js
+++ b/src/lex.js
@@ -435,6 +435,18 @@ Lexer.prototype = {
           body = body.substr(str.length + 1);
         }
 
+        // To handle rarer case when special word is separated from label by
+        // multiple spaces or tabs
+        var strIndex = body.indexOf(str);
+        if (!isSpecial && strIndex >= 0) {
+          var specialRegex = new RegExp('[\s\t]*' + str + '[\s\t]*');
+          if (specialRegex.test(body)) {
+            isSpecial = true;
+            label = label + " " + str;
+            body = body.substr(strIndex + str.length);
+          }
+        }
+
         if (!isSpecial) {
           return;
         }
--- a/src/lex.js
+++ b/src/lex.js
@@ -438,12 +438,12 @@ Lexer.prototype = {
         // To handle rarer case when special word is separated from label by
         // multiple spaces or tabs
         var strIndex = body.indexOf(str);
-        if (!isSpecial && strIndex >= 0) {
-          var specialRegex = new RegExp('[\s\t]*' + str + '[\s\t]*');
-          if (specialRegex.test(body)) {
+        if (!isSpecial && strIndex >= 0 && body.charAt(strIndex + str.length + 1) === " ") {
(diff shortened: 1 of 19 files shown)


johnmryan:
@jugglinmike can you take a look at these changes when you get a chance? I kept the previous conditionals in there because I figured this was a particularly rare case, so it shouldn't always hit this code path.

jugglinmike:
Thanks for the patch! I ran out of time this weekend, but I hope to have time to give this the attention it deserves by next weekend (if not sooner).

src/lex.js
```diff
@@ -435,6 +435,18 @@ Lexer.prototype = {
           body = body.substr(str.length + 1);
         }
 
+        // To handle rarer case when special word is separated from label by
+        // multiple spaces or tabs
+        var strIndex = body.indexOf(str);
+        if (!isSpecial && strIndex >= 0 && body.charAt(strIndex + str.length + 1) === " ") {
+          var specialSubstr = body.substr(0, strIndex + str.length);
+          var isAllWhitespace = (specialSubstr.length - specialSubstr.trim().length) === strIndex;
```
jugglinmike: I think this can be simplified by capturing just the preceding characters (`body.substr(0, strIndex)`) and making sure that string contains only whitespace (as in `.trim().length === 0`).

jugglinmike:
Thanks for taking the time to submit a patch, and thanks also for including test material. At the moment, those new files aren't being used to validate this change, though. To fix this, you'll need to add some code to the `tests/unit/parser.js` file. As of this writing, [line 373](https://github.com/jshint/jshint/blob/c7c5c42885aa618e384de3f6589793fd419de7c1/tests/unit/parser.js#L373) looks like an appropriate place for these. You can use the surrounding tests as a guide for how to get things running. That ought to satisfy our code coverage tool.

Also note that these tests don't actually demonstrate the new functionality. For instance:

    /*              jshint 
        browser: true
    */

    var dp = new DOMParser();
    console.log(dp);

JSHint is not expected to emit any errors or warnings for this input, even without your patch applied. That's because `browser: true` will only really help if `undef: true` is also set--in that case, it will suppress warnings about `DOMParser` being undefined.

I would recommend validating this change by using a simpler option. `unused` seems like a good choice. You can also put all the tests in the same file:

    function a() {
      //     jshint unused: true
      var x;
    }

    function b() {
      /*
       jshint unused: true */
      var x;
    }

    // etc. etc.

With the above approach, your tests should assert that JSHint warns about `x` being unused in each function. That will prove that JSHint is correctly recognizing the in-line directives.

Does that make sense?

johnmryan:
@jugglinmike , I have fixed the code per your suggestion but I am confused by your explanation of the `DOMParser` example. My understanding was that setting `browser: true` would suppress the warning because the `DOMParser` object name is known to the browser. I am pretty sure testing this with the current version of `jshint` suppressed the warnings, at least. 

This shows no warnings.
`/*              jshint 
    browser: true
*/

var dp = new DOMParser();
console.log(dp);`

This warns that `DOMParser` is not defined.
`var dp = new DOMParser();
console.log(dp);`

jugglinmike:
JSHint will only warn about undefined variables if the `undef` option is enabled. In your test, the `undef` option is not enabled, so JSHint is not expected to warn about the reference to `DOMParser` (regardless of the state of the `browser` option).

Now, depending on how you are interacting with those new test files, you may experience the warning we're discussing. For instance, if you invoke the JSHint CLI from the root of the JSHint project's source tree, then you will be linting the code with this project's configuration options (which happens to enable the `undef` option). This is not an accurate representation of how JSHint behaves by default.

This gets back to the other point I made in my previous comment: the new test files are not actually being used. If you follow the advice I shared about adding code to `test/unit/parser.js`, then you can avoid the kinds of variation that are causing confusion here.

johnmryan:
Alright @jugglinmike, should be better now. Definitely was running it from the project root, I was not aware of the default configuration, thanks for that.

jugglinmike:
Merged to `master` at commit efeb0f83da1d20aee28f60cdea0f2dd20ea06dd4. Thanks for following through with this, Jack!

PR #3207, opened 2017-10-28
[[DOCS]] Fix incorrect information about scopes in JavaScript

kpiotr:
Comment stated that JavaScript has only two real scopes.
This is incorrect as identifiers declared with let and const do have
block scope.

Maybe a little nitpicky but I think that as long as the information is there, it should be correct. It's also wrong on http://jshint.com/docs/options/

### DIFF
--- a/dist/jshint-rhino.js
+++ b/dist/jshint-rhino.js
@@ -16287,13 +16287,12 @@ exports.bool = {
     evil        : true,
 
     /**
-     * This option suppresses warnings about declaring variables inside of
-     * control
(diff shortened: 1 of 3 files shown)


jugglinmike:
Not nitpicky at all; your change is correct and helpful.

I've removed the changes to the files in `dist` because we generate those files for each release.  And as for the website; that content is derived from the source code, so your new text will be pushed there when we release the next version of JSHint.

Merged to `master` at commit c4370d9ee8b5367cfc0bf391618e7029852c78b8. Thanks for taking the time to fix this!

PR #3209, opened 2017-11-12
[[TEST]] Refactor negative syntax tests for "rest"

jugglinmike:
Validate JSHint's behavior for various invalid applications of the
"rest" operator in isolation. In the previous factoring, the independent
syntax tests interacted and caused JSHint to report errors that differ
from those it would report when encountering each form in isolation.
Although JSHint makes a best-effort approach to recovering from invalid
syntax, this behavior is not formalized and should therefor not be
asserted by unit tests.

This factoring avoids the undesirable interaction, allowing the tests to
document JSHint's behavior under more predictable conditions. It also
highlights a previously-existing bug in the errors reported (as
documented via in-line code comments).

---

@rwaldron Mind taking a look? This is pretty minor, but it will help me while rebasing the `v2.10.0` branch (this commit in particular 8b1bbf38cfe4690e810db6c7bbc591575abffbfe).

### DIFF
--- a/tests/unit/parser.js
+++ b/tests/unit/parser.js
@@ -7667,41 +7667,49 @@ exports.commaAfterRestParameter = function (test) {
 
 
 exports.extraRestOperator = function (test) {
-  var code = [
-    'function fn([a, b, ......c]) { }',
-    'function fn2([......c]) { }',
-    'function fn3(a, b, ......) { }',
-    'function fn4(......) { }',
-    'var [......a] = [1, 2, 3];',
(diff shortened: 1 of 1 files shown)


rwaldron:
> In the previous factoring, the independent syntax tests interacted and caused JSHint to report errors that differ from those it would report when encountering each form in isolation.

This is excellent!

Travis is failing in Node.js 0.10.x and 0.12.x because a nested dependency is using `const` :\

jugglinmike:
@rwaldron Since you merged gh-3215, I re-started the build, and everything is green. Hooray!

Issue #3211, opened 2017-11-15
List outer scoped variables of W083


flaviojs:
I was getting W083 with this code:
```
$.ajax({
  contentType: "application/json; charset=utf-8",
  type: "GET",
  url: api("/machines"),
  beforeSend: function() { $('#machines-spinner').removeClass('hidden'); },
  cache: false,
}).done(function (machines) {
  if (machines.length > 0) {
    var $machines = $('#machines').html('');
    for (var i = 0; i < machines.length; ++i) {
      var html = $('#template-machine').html().replace(/{{\w*}}/g, function (x) { // W083
        if (x === '{{id}}') return machines[i].id;
        if (x === '{{name}}') return machines[i].name;
        if (x === '{{targetoee}}') return machines[i].targetoee;
        if (x === '{{code}}') return machines[i].code;
        return x;
      });
      $machines.append($(html));
    }
  }
  else {
    $('#machines').text('No data.');
  }
  $('#machines-spinner').addClass('hidden');
}).fail(...);
```

I tried wrapping the code but it didn't remove the error:
```
...
      (function (machines, i, $machines) { // W083
        var html = $('#template-machine').html().replace(/{{\w*}}/g, function (x) {
          if (x === '{{id}}') return machines[i].id;
          if (x === '{{name}}') return machines[i].name
          if (x === '{{targetoee}}') return machines[i].targetoee
          if (x === '{{code}}') return machines[i].code;
          return x;
        });
        $machines.append($(html));
      })(machines, i, $machines);
...
```

Not knowing what else to try I gave up and added the `loopfunc` option.

I still don't know which "outer scoped variable" he was referring to.
I'd like jshint to list the offending outer scoped variable(s) with W086, even if it's only in `--verbose` mode.

jugglinmike:
Hi there, Flávio, thanks for the report! The problem is the reference to the `$` function. There are two ways to resolve this:

1. Include `$` in the parameters of the immediately-invoked function expression (and the corresponding argument list)
2. Configure JSHint to recognize `$` as an immutable global binding (i.e. `$:  false`). JSHint should correctly recognize that there is no risk for confusion when the reference cannot be modified

That said, I agree that debugging this warning is kind of difficult without the list of bindings. Implementing your request shouldn't require very much code. I'd be happy to help you write a patch--do you have any interest?

flaviojs:
I was expecting the `jquery` option to handle `$` since it defines it, seems I was wrong.

I'll try to make a pull request.

[...]

PR #3214, opened 2017-11-20
[[FEAT]] List outer scoped variables of W083

flaviojs:
Debugging W083 is kind of difficult without knowing which outer scoped variables it is referring to. This commit shows the variable names inside parenthesis at the end of the warning message.

References: #3211

-----

There was a generic "Application Error" when I signed the [Contributor's License Agreement](https://www.clahub.com/agreements/jshint/jshint), but when I opened it again it said I signed the CLA, so it should be fine.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3059,8 +3059,8 @@ var JSHINT = (function() {
       // If the function we just parsed accesses any non-local variables
       // trigger a warning. Otherwise, the function is safe even within
       // a loop.
-      if (f["(isCapturing)"]) {
-        warning("W083", token);
+      if (f["(outerMutables)"]) {
+        warning("W083", token, f["(outerMutables)"].join(", "));
       }
     }
 
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -315,7 +315,10 @@ var scopeManager = function(state, predefined, exported, declared) {
           isImmutable = labelType === "const" ||
             (labelType === null && _scopeStack[0]["(predefined)"][usedLabelName] === false);
           if (isUnstackingFunctionOuter && !isImmutable) {
-            state.funct["(isCapturing)"] = true;
+            if (!state.funct["(outerMutables)"]) {
+              state.funct["(outerMutables)"] = [];
+            }
+            state.funct["(outerMutables)"].push(usedLabelName);
           }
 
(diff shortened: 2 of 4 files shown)


jugglinmike:
Thanks for the patch, Flávio! And for moving so fast! Unfortunately, I'm not able to respond quite so quickly, but know that I've seen your submission, and I intend to give you some feedback within the next few days.

By the way: I can see that your signature was accepted, so we're all set on that front.

src/jshint.js
```diff
@@ -3060,7 +3060,7 @@ var JSHINT = (function() {
       // trigger a warning. Otherwise, the function is safe even within
       // a loop.
       if (f["(isCapturing)"]) {
-        warning("W083", token);
+        warning("W083", token, f["(isCapturing)"].join());
```
jugglinmike: Mind adding a space to the separator string? As in, ", "

src/scope-manager.js
```diff
@@ -315,7 +315,10 @@ var scopeManager = function(state, predefined, exported, declared) {
           isImmutable = labelType === "const" ||
             (labelType === null && _scopeStack[0]["(predefined)"][usedLabelName] === false);
           if (isUnstackingFunctionOuter && !isImmutable) {
-            state.funct["(isCapturing)"] = true;
+            if (!state.funct["(isCapturing)"]) {
+              state.funct["(isCapturing)"] = [];
```
jugglinmike: The property name "isCapturing" strongly suggests a boolean value. Since we're now implementing it as an array, a new name is in order. This is tricky to describe accurately and concisely, so I'm open to suggestion. My initial thought is "capturedMutables", but I'm not sold on that.

src/jshint.js
flaviojs: Ok.

src/scope-manager.js
flaviojs: "outerMutables"?

jugglinmike: Even better!

jugglinmike:
Squashed and merged to `master` at commit d03662c84c9eb032ab2163489989cde45b8f8565.

From initial report, to pull request, to landing the feature in `master`, you really went the distance on this. It's thanks to contributors like you that this project exists at all--we really appreciate it! :)

PR #3217, opened 2017-11-27
[[FIX]] Tolerate RegExp as `void` operand

jugglinmike:
Hey @rwaldron. I imagine that a patch like this would cause you to wonder about my priorities. Rest assured: I'm focused on async functions these days. I just stumbled on this corner case while trying to parse `await` correctly.


rwaldron:
LGTM

PR #3222, opened 2017-12-03
Add global objects for the browser: true option.

patrickdark:
* Add all missing interface global objects from CSSOM (https://drafts.csswg.org/cssom/).
* Add all missing element interface global objects from WHATWG HTML (https://html.spec.whatwg.org/multipage/indices.html#element-interfaces).
* Add all missing obsolete element interface global objects (i.e., `HTMLMarqueeElement`) from WHATWG HTML (https://html.spec.whatwg.org/multipage/obsolete.html#obsolete).
* Add the legacy `XMLDocument` interface global object from WHATWG DOM (https://dom.spec.whatwg.org/#xmldocument).
* Move the `HTMLElement` interface global object so that it appears in alphabetical order.
* <del>Remove the obsolete `HTMLIsIndexElement` interface global object from DOM2 HTML (https://www.w3.org/TR/2003/REC-DOM-Level-2-HTML-20030109/html.html#ID-85283003). The latest versions of Chrome, Edge, Firefox, and Safari don't support this object and it isn't listed in the WHATWG HTML spec's obsolete features section (https://html.spec.whatwg.org/multipage/obsolete.html#obsolete).</del>

<del>The little-used `isindex` element was removed from the Web platform because it created browser attack surface (i.e., allowing fingerprinting of the user and cross-site site scripting (XSS)) and created browser maintenance issues for little benefit.</del>

<del>WHATWG HTML GitHub Issue: https://github.com/whatwg/html/issues/1095
Bugzilla@Mozilla Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1266495
Chromium Code Review Issue: https://codereview.chromium.org/96653004/
WebKit Bugzilla Bug: https://bugs.webkit.org/show_bug.cgi?id=7139</del>

<del>I don't see a Microsoft EdgeHTML issue tracker page, but I've confirmed that the object doesn't exist in the latest version of Microsoft Edge (41.16299.15.0).</del>


src/vars.js
```diff
@@ -105,18 +116,25 @@ exports.browser = {
...
   HTMLCanvasElement    : false,
   HTMLCollection       : false,
+  HTMLDataElement      : false,
+  HTMLDataListElement  : false,
+  HTMLDetailsElement   : false,
+  HTMLDialogElement    : false,
   HTMLDirectoryElement : false,
   HTMLDivElement       : false,
   HTMLDListElement     : false,
+  HTMLElement          : false,
```
jugglinmike: The tidying is appreciated :)

src/vars.js
```diff
@@ -129,26 +147,32 @@ exports.browser = {
   HTMLIFrameElement    : false,
   HTMLImageElement     : false,
   HTMLInputElement     : false,
-  HTMLIsIndexElement   : false,
```
jugglinmike: This removal may reflect the state of modern browsers, but many of JSHint's consumers target legacy environments. That means this particular change constitutes a breaking change, so we'll have to defer it for the next major release. Could you instead add an in-line comment with the research you reported in the pull request description?

patrickdark: @jugglinmike I modified the pull request title and description. I hope those changes suffice.

Support for `isindex` in the most used browser—Google Chrome—was apparently pulled three years and nine months ago, so this seems like a low-risk change.

patrickdark: @jugglinmike I've changed the WHATWG HTML GitHub issue reference in the description from 1088 ("Remove `isindex` parsing special case") to 1095 ("Remove `<isindex>` and `<input name=isindex>`"). The latter more directly addresses the removal of `HTMLIsIndexElement`.

jugglinmike: That's more great research, thanks! Though it's good evidence for the low risk of removal, I still don't consider the benefit to be great enough to warrant it. Would you mind adding this back in with a comment describing your findings?

patrickdark: @jugglinmike Yes. I:

* restored `HTMLIsIndexElement` with an ECMAScript comment preceding it.
* updated the description with `del` elements.
* updated the pull request title.

jugglinmike:
Squashed and merged to `master` at commit badc7a445457c0a2c5b76570250262492214b98c. Thanks again, Patrick!

Issue #3230, opened 2017-12-23
Hangs when given a for loop with only the initial condition and an opening curly brace


outoftime:
* Version of JSHint being used: 2.9.5
* Input source code (simplified to only contain details necessary to demonstrate the problem):

  ```js
  for(var i=1){
  ```
* Configuration values: Defaults
* Description of expected behavior: Linting completes with a failing result
* Description of actual behavior: Linting hangs indefinitely, never returning a result and locking up the interpreter

jugglinmike:
Thanks for the report! Confirmed for JSHint at version 2.9.5 and `master` (badc7a4). Sorry you had to downgrade; we'll try to get this fixed for 2.9.6.

johnmryan:
Just took a quick look into this. I guess it is coming from https://github.com/jshint/jshint/commit/c083866... the previous implementation threw a `TypeError` because we were trying to check `null.type` on the return value. Now, we end up in an infinite loop in the `peek(p)` function because `j` never gets incremented in this case. I think we may be able to fix it by incrementing `j` before the `return lookahead[j - 1];`?

johnmryan:
If you think that seems feasible @jugglinmike, I can submit a PR for this.

johnmryan:
@jugglinmike I can work on a PR here unless you have context for an easy fix.

jugglinmike:
@johnmryan I do have that context:

It looks like we've already got this bug licked in the feature branch for the
next minor release (2.10.0). But we don't have a firm plan for that release, so
I've implemented a different fix that's suitable for our next patch
release--see gh-3231.

Sorry if this obviated any of your work, Jack. I usually try to be a little
more vocal about my work, but I've been offline since my last reply (and my
fix) in celebration of the Christmas holiday.

PR #3231, opened 2017-12-27
[[FIX]] Avoid infinite loop on invalid `for` stmt

jugglinmike:
Ensure that the special "end" token cancels parsing of the `for`
statement "head" in all cases.

---

Hey @rwaldon. The modified parsing logic has problems that extend beyond the
issue reported in gh-3230. As you know, we've refactored that code heavily in
the feature branch for the next minor release, so the problem is already fixed
there. But since we don't have a solid plan for 2.10.0 yet, I've implemented
this different solution so we can resolve the bug in our next patch release.
This will cause merge conflicts for the 2.10.0 feature branch, but I'm prepared
to resolve those.


rwaldron:
LGTM

Issue #2604, opened 2015-07-30
Support for async functions


jugglinmike:
Proposal: https://github.com/tc39/ecmascript-asyncawait

Advanced to stage 2 on July 30, 2015 https://github.com/tc39/ecma262

Depends on gh-2601

bterlson:
:+1:

lukeapage:
#2077 may be of use

rDr4g0n:
I commented on #2601, but this seems like the right place to comment specifically for async/await.

async/await is now [finished](https://github.com/tc39/proposals/blob/e8c03544751b071b973064ef7d437a34ac6752cd/finished-proposals.md), [merged into](https://github.com/tc39/ecma262/pull/692) the ES2017 specification, and shipping in chrome and ff.

Maybe it's time to dust off #2077 ?

Simobain:
Ok, so now that node 8 is live. Async/Await is officially supported. 
Will it be available soon on jsHint ?

Thanks,

lolmaus:
I'll just leave this here: http://eslint.org/

carldunham:
I'll just leave this here: https://github.com/eslint/eslint#what-about-experimental-features

johansyd:
You just lost my interest in jshint. Sorry, to slow.

SeregPie:
Node.js supports async/await. Every modern browser supports async/await. JSHint does not support it. This topic is 2 years old.

Thank you JSHint. It was good while it lasted.

Switched to ESLint.

Grafluxe:
Hi team, 

The async spec has reached [stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md), when can we expect to see support?

tomaszkunicki:
RIP Jshint

caitp:
That is not a helpful comment. Please submit a patch, rather than being negative.

Grafluxe:
I was a fan of jshint, but have now decided to move to eslint. The goal of ECMAs new release cycle — and the creation of transpilers — is to allow people to use the lastest-greatest features of JavaScript early. Async/await has been at stage 4 for over a month now (and baked into Node and many mainstream browsers for longer), but jshint has yet to add support.

I used jshint daily, so thank you for creating something that was more flexible than jslint. I now need a more powerful and up-to-date tool with support for the latest ES features as well experimental ones. 

All the best.

ddcovery:
[...]

PR #3234, opened 2018-01-08
Miscellaneous fixes

jugglinmike:
Hey @rwaldron,

Here are some more improvements in service of gh-2604. Note that I'm targeting `master`--these are all bugs that can be observed in the current release of JSHint. Who would have guessed implementing async functions would uncover so many problems?

I originally wrote these patches across a few distinct branches, but I decided that the work we'd have to do resolving conflict didn't justify the benefit of being able to land them independently. That said, if you'd like to put off some part of this and land the rest, I'm happy to factor it apart as necessary. Let me know!

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4042,6 +4042,7 @@ var JSHINT = (function() {
     advance(")", t);
     t = state.tokens.next;
     advance("{");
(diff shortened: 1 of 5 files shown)


rwaldron:
LGTM

PR #3237, opened 2018-01-15
[[CHORE]] Simplify lexing of literal values

jugglinmike:
Lex NullLiterals and BooleanLiterals as keywords, not Identifiers. This
simplifies the `scanIdentifier` function and makes its name more
accurately describe its behavior. It also obviates the need for two
token "types" that are internal to the lexer itself (as in either case,
the "type" attribute of the emitted token object is "(identifier)".

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -26,8 +26,6 @@ var Token = {
   StringLiteral: 4,
   Comment: 5,
   Keyword: 6,
-  NullLiteral: 7,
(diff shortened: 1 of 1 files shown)


rwaldron:
> Coverage decreased (-0.002%) to 98.752%

I feel like coverage machinery should reward you for removing lines of code

jugglinmike:
:)

PR #3243, opened 2018-01-22
[[FIX]] Restrict "name" of strict mode functions

jugglinmike:
@rwaldron this is another fix in support of gh-2604


rwaldron:
Excellent!

PR #3248, opened 2018-01-28
[[Docs]] Fix spellling, trys to tries.

ssochet:
Should this be Tries instead of Trys?

### DIFF
--- a/dist/jshint-rhino.js
+++ b/dist/jshint-rhino.js
@@ -916,7 +916,7 @@ exports.debuglog = function(set) {
 
 
 /**
- * Echos the value of a value. Trys to print the value out
+ * Echos the value of a value. Tries to print the value out
  * in the best way possible given the different types.
  *
  * @param {Object} obj The object to print out.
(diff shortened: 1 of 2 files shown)


jugglinmike:
The files in the `dist` directory are built automatically each time we release a new version of JSHint, and they include a large amount of source code from the project's dependencies. We don't maintain that code within this project.

So while it looks like that *is* a typo, and your patch corrects that, this is not a detail we can control here.

The good news is that it has been fixed in the "upstream" project. [Here's the typo on Node.js at version 0.12.6](https://github.com/nodejs/node/blob/v0.12.6-release/lib/util.js#L118), but it's no longer present in [the same file as it extists on the `master` branch](https://github.com/nodejs/node/blob/5c8ce90c2f47fc4fe699bbfb62f7004c4192d6e2/lib/util.js#L118).

So I'm going to close this pull request, but thanks for your interest in JSHint all the same!

ssochet:
Thank you for responding. I love JSHint, I use it all the time.

PR #3254, opened 2018-02-21
Update identifier data to ES2015+ and Unicode 11

mathiasbynens:
The data still excludes astral symbols although ES2015+ supports (some of) those in identifiers. This can be fixed in a follow-up commit that touches `lex.js` as well.


rwaldron:
@mathiasbynens there's a conflict in the package.json file, can you resolve this? Thanks!!

mathiasbynens:
@rwaldron Done!

rwaldron:
@mathiasbynens in the future, please use `git pull --rebase upstream master && git push origin HEAD -f` on your branch to rebase cleanly. Thanks!


@jugglinmike I'm +1 for this, can you review and merge when ready?

mathiasbynens:
@rwaldron I used GitHub’s “resolve conflict” UI, not knowing this was frowned upon. Noted for next time!

jugglinmike:
Thanks for the patch, Mathias! To me, the integration with `lex.js` is a crucial aspect of this fix. It doesn't seem like there is any benefit to merging this change set without that. I can pick up where you've left off, but that will take me some time (all of my development effort is focused on [async functions](https://github.com/jshint/jshint/issues/2604)). I'd appreciate it if you could carry this home. I don't know what that would entail, though--do you have the bandwidth for it?

mathiasbynens:
I do think these are separate changes, though. This patch updates the Unicode data to the latest version, future-proofing the generator script (i.e. when Unicode v11 is released, we can just update a dependency) and to ES2015+ semantics *except* for astral symbols. That’s much better than the status quo, and worthwhile all by itself IMHO.

The follow-up patch would make it fully conform to ES2015+ which supports some astral symbols in identifiers. I haven’t had time to look into this in detail yet, so I can’t really estimate how much work it would be.

jugglinmike:
I think I'm missing something important because I assumed that this patch does not actually change the behavior of JSHint. If it does, then the fact that it doesn't require changes to the Test262 "expectations" file suggests that we are missing coverage in Test262. Then again, we're using an older revision of that project, and it could be that the relevant tests have recently landed.

Can you give me a pointer for observing the effect of this change?

mathiasbynens:
@jugglinmike See `src/data-*.js`; e.g. [`var Ϳ;`](https://mothereff.in/js-variables#%CD%BF) now parses correctly.

jugglinmike:
@mathiasbynens I'm only just now returning to this; sorry for the delay! I could use your input on the best way to expose this new functionality. JSHint supports parsing program code in terms of outdated editions of ECMAScript. By enabling support unilaterally, we will be accepting the use of these new code points in projects that have been configured for ES5. In those contexts, the code points probably ought to trigger an error--would you agree?

In that case, though, I don't really know the best way to interpret Unicode versions generally. The spec reads, "A conforming implementation of ECMAScript must interpret source text input in conformance with the latest version of the Unicode Standard and ISO/IEC 10646," so we have a moving target on our hands. I'm inclined to think that without introducing a dedicated configuration parameter (i.e.  `unicodeVersion: 8` to match the existing `esVersion: 6`), we can't meaningfully restrict the character set.

So it might make sense to be as forgiving as possible. It's easy to make this decision today, but it's hard to renege if we change our minds later. Your perspective here will give me more confidence in a decision in either direction. Do you have a few moments to share it?

mathiasbynens:
There’s two separate issues here:

1. Extending the set of accepted characters, while sticking with the ES5 identifier grammar. ES5 only accepts characters in the range from U+0000 to U+FFFF in identifiers. By updating to the latest Unicode version, we extend the set of allowed characters within this range. That’s exactly what this patch does. It seems like a harmless change to me.

1. Implementing ES2015 identifier parsing. That is, extend the accepted character range to U+10FFFF and support `\u{XXXXXX}` syntax. This is likely a more complicated change, since the parser now needs to operate on code points rather than UCS-2/UTF-16 code units for identifiers. An example from <https://mathiasbynens.be/notes/javascript-identifiers-es6>:

    ```js
    // Invalid in ES5, but valid in ES2015:
    var 𐊧; // U+102A7 CARIAN LETTER A2
    var \u{102A7}; // U+102A7 CARIAN LETTER A2
    ```

Note that the set of accepted identifiers in ES2015 is a superset of the valid identifiers in ES5.

If you want to be strict about the different ES5/ES2015 semantics, you need two separate identifier parsers. This seems like overkill to me.

mathiasbynens:
`Ϳ` is U+037F GREEK CAPITAL LETTER YOT. It's a valid `Identifier` both per the old ES5 definition (because it's an `Uppercase_Letter`/`Lu` character) and per the new ES2015+ definition (because it's an `ID_Start` character).

It seems like you want the ES5 identifier regex to use an older version of the Unicode Standard. In that case, change these lines: https://github.com/jshint/jshint/pull/3254/commits/5995a9fe3b4b8d085447b387735466344677ee9d#diff-0b9174fac00b2f350070cebadda2dac4R22 Add a `getOld` function that uses e.g. `unicode-5.1.0` instead of using `unicode-11.0.0`. That would fix the issue.

mathiasbynens:
I've just pushed up a fix.

I used Unicode v5.2.0 (and not Unicode v5.1.0) because that's the minimum version `tests/unit/fixtures/identifiers.js` expects:

* Lines [16](https://github.com/jshint/jshint/blob/8df4a328c5019fd74e625c36ea18bbad82cc8dbf/tests/unit/fixtures/identifiers.js#L16) and [27](https://github.com/jshint/jshint/blob/8df4a328c5019fd74e625c36ea18bbad82cc8dbf/tests/unit/fixtures/identifiers.js#L27) use U+A6ED BAMUM LETTER FAAMAE, which was only added in Unicode v5.2.0.
* [Line 29](https://github.com/jshint/jshint/blob/8df4a328c5019fd74e625c36ea18bbad82cc8dbf/tests/unit/fixtures/identifiers.js#L29) uses U+ABE9 MEETEI MAYEK VOWEL SIGN CHEINAP, which was added in Unicode v5.2.0.

Feel free to choose any other Unicode version number for the ES5 identifier regex.

jugglinmike:
Wow, thank you Mathias--this is great! I made a couple alterations to commit fd962b5c48ab3c9057a55b67712c794f5ff20b02:

- Removed the built files in the `dist/` (we generate those at release time; deferring like this reduces the noise in patches)
- Reverted the Test262 update. We're locked to a very old revision of the test suite, but updating it should take place in a dedicated patch (because I'm afraid it will also introduce a lot of new entries into our whitelist file)

Here's the branch:

https://github.com/jshint/jshint/compare/master...jugglinmike:master-gh3254

Despite appearances, I am actually reluctant to modify others' contributions, so if you wouldn't mind one more back and forth: could you make sure that things still look good to you?

mathiasbynens:
I’m not sure which “Test262 update” you’re referring to — I don’t recall updating anything related to Test262 as part of this PR.

Other than that the changes LGTM.

mathiasbynens:
Oooh, now I see, I guess I accidentally updates the Test262 submodule. I didn’t mean to, so thanks for fixing that :)

mathiasbynens:
Thanks! Let’s get this merged :)

jugglinmike:
:tada: Thanks for all your help, @mathiasbynens!

PR #3264, opened 2018-03-23
Adds JSLint JSON license

renehamburger:
This reflects the actual license of JSHint as expressed in its package.json `(MIT AND JSON)` and at http://jshint.com/about/.


jugglinmike:
Thanks for your interest in JSHint! By combining both license texts into a
single file, this patch implies that all files in the source tree are governed
by the MIT Expat license and the JSON license. In fact, only the file currently
named `src/jshint.js` is governed by the JSON license. The way we communicate
this is through the traditional software licensing pattern of including the
appropriate license in the very beginning of the file to which it applies.

We draw attention to this unusual situation in two ways. For machines, we
reference both licenses via the SPDX license specification in the project's
`package.json` file. For humans, we include an explanation in the project's
`README.md` file.

This has been sufficient for folks for many years, so I'd like to avoid making
a change that could cause confusion.

renehamburger:
Thanks for the clarification!

PR #3272, opened 2018-04-23
Support for async functions

jugglinmike:
This patch isn't quite ready yet. A lot of people are eager to see support in JSHint, so I'm opening this pull request to help them check in on my progress.

Most of the challenge here relates to extending the JSHint parser. Whenever that work improves the correctness of JSHint in some other way, I am submitting the patch as a standalone change set. If possible, I've been targeting the `master` branch so the corrections are available sooner (i.e. in the next patch release of JSHint). When the patches can't be applied cleanly to `master`, I've been submitting them to the tracking branch for the next minor release of JSHint.

In all cases, my hope is to keep the changeset for async functions as focused as possible and to amortize the burden of code review (which @rwaldron has been shouldering heroically). What this also means is that not all of my efforts in service of async functions are visible from this pull request.  To help with that, I'll try to comment here with notes about such standalone fixes. In addition, I'll open a separate pull request to track the next minor release of JSHint (link forthcoming).


jugglinmike:
Brief update for those following along:

One of the big challenges here is the fact that in ES2017, `await` is sometimes a valid identifier (e.g. outside of async functions). There's precedence for this kind of "magic keyword" even in ES2015: `let`.  Unfortunately, JSHint has never implemented that correctly. Ever since the introduction of `esversion: 2015`, JSHint has unconditionally disallowed the use of `let` as an identifier.  So in keeping with my desire to limit the overall size of this patch, I've been working to fix `let` in a way that can be re-used here. [I submitted a patch for this last month against the `v2.10.0` feature branch](https://github.com/jshint/jshint/pull/3284).

Today, I made some additional progress on this branch, and this also included fixing some seemingly-unrelated problems:

- An early error we missed from ES2015: gh-3300
- A defect our implementation of the ES2016 restriction on the "use strict" directive for function code: gh-3301

jugglinmike:
Here are a couple more patches that will help with this feature (though they probably don't seem that way!):

- https://github.com/jshint/jshint/pull/3307
- https://github.com/jshint/jshint/pull/3312

jugglinmike:
There's been lots of progress in the `v2.10.0` branch since I opened this patch, so I just rebased. I've also introduced a new linting rule, `noreturnawait`. If possible, I'd like to land that as a separate commit for clarity (and I guess because I love to rebase things).

Simobain:
Hey @jugglinmike ! 
When will the v2.10 available ;) ?

Thanks,

oskarkrawczyk:
@jugglinmike thanks a lot for the work and progress on this!

Do you think you'll have more time to work on this patch and wrap it up by the end of the year? 

I'm asking because we get a lot of reports about `async/await` "not being supported by @jsfiddle", and the only thing I can say is: 🤷‍♂️ 

To be honest if you say that this is probably going to take a few more months, I'll probably will have to disable it for peace of mind, but obviously I wouldn't want to do that :)

jugglinmike:
@oskarkrawczyk I would love to have this implemented by the end of the year, but at this moment, I can't commit to that in good faith. I've been stalled by gh-1234, believe it or not, but I hope to get more clarity on that some time this week.

I'll share an update here as soon as I know more. It's nice to know that JSFiddle is still making use of JSHint, and it would be great to put an end to those complaints!

oskarkrawczyk:
@jugglinmike thanks for the update. That `gh-1234` looks really unfortunate, Doug's stand on the issue can be slightly upsetting.

If you have any info, bad or good, I'd appreciate if you could let me know (here or directly), would be good to plan.

Thanks again for the awesome work man!

jugglinmike:
Following some exciting progress on [the version 2.10.0 feature branch](https://github.com/jshint/jshint/pull/3273), this patch is finally in a review-worth state! I've rebased my work-in-progress and force-pushed the result (though if anyone is interested in the long, confused effort to get here, [the work-in-progress branch is available on my fork](https://github.com/jugglinmike/jshint/commits/async-functions-2019-01-20)).

Finishing this off entailed making a few more corrections to JSHint which are only indirectly related to async functions. Although I previously submitted those fixes separately (see notes above), I decided to lump them in here to streamline the review process.

Since ES2018 has been standardized since I started, this patch also includes support for asynchronous iteration (implemented with a dedicated commit). I also snuck a new linting rule in there: `noreturnawait`.

Once this lands, I think we'll be in good shape to release it as JSHint version 2.10.0!

@rwaldron and/or @caitp: Would you mind taking a look?

tifDev:
So now jshint supports async functions? Are this new changes in any documentation?

src/jshint.js
```diff
@@ -132,17 +132,13 @@ var JSHINT = (function() {
     return true;
   }
 
-  function isReserved(token) {
+  function isReserved(context, token) {
```
rwaldron: I was curious about `context` being added as the first parameter in the list, so I went looking for other examples and discovered that `function expression(rbp, context)` is the only place where the `context` bitfield is "not the first" parameter. I wonder if that function should be updated to match the signature, ie. `context` first?

rwaldron: I would expect that update to occur in a different changeset.

src/lex.js
```diff
@@ -577,7 +577,7 @@ Lexer.prototype = {
       "super", "return", "typeof", "delete",
       "switch", "export", "import", "default",
       "finally", "extends", "function", "continue",
-      "debugger", "instanceof", "true", "false", "null"
+      "debugger", "instanceof", "true", "false", "async", "await"
```
rwaldron: Did you intend to remove "null"?

jugglinmike: I did not. I've pushed up a "fixup" commit to address this. Strange that it didn't cause regressions; I'll look into that later.

src/jshint.js
jugglinmike: Totally!

rwaldron:
Merged into v2.10.0

PR #3273, opened 2018-04-23
Version 2.10.0 (in progress)

jugglinmike:
We have been hard at work on JSHint version 2.10.0 for over six months now. We're still working, but I'm opening this pull request today in order to give folks a better view into our progress.

Here's a high-level summary of what the next version contains and what we hope to implement before releasing:

- [x] New `esVersion`: 2016 
  - Exponentiation operator
  - Restrictions on the function form of the "use strict" directive
- [ ] New `esVersion`: 2017
  - Async functions
  - Trailing commas in parameter lists
- [x] `unstable` configuration
  - Generic support
  - Options
    - Object "rest" properties
- [x] New linting option: `leanswitch`

If you're interested in adding a new feature to JSHint, we encourage you to work from this branch. However, since we continue to publish bug fixes under the 2.9 release train, **we reserve the right to rebase this branch**. You should also be prepared to rebase your pull request as requested. (This is a more advanced workflow for git, and we'll be happy to assist you as necessary.) Those following along should also note that commits referenced on pages for merged pull requests may be present here under a different SHA.

PR #3274, opened 2018-04-26
[[FIX]] Improve handling of tab indentation.

### LINKED ISSUE #3151: Tab indentation breaks multiple rules

TzviPM:
By tracking the column number both with and without soft tabs,
we enable better messages for errors and warnings. This is especially
useful for editor linting integrations.

Fixes #3151


jugglinmike:
Thanks for the patch! This problem has been causing a lot of headaches, as gh-3151 demonstrates. I'm definitely interested in moving forward with this, but I think we need to reformat it a bit first.

The current behavior is somewhat surprising, but it is also poorly-documented.  At the moment, the documentation of the data reported by JSHint is limited to a single example on jshint.com, and that example does not even include the corresponding input text. Over the years, consumers have been forced to discover the feature and learn its behavior experimentally. So as surprising as we may find the behavior, it's possible that code out there has come to rely on it. Of course, anyone could be relying on any flaw in JSHint's implementation, and this alone isn't enough to stop us from fixing bugs. The extenuating circumstance here is the poor state of the documentation and the ambiguity of the expected behavior.

That's why I think we need to introduce this as a new feature rather than a bug fix. What do you think of this:

- We expose the new value via a new property on the report object named `column`
- We continue to expose the `character` attribute without change
- We document the entire report object, defining exactly what each property is intended to describe and labeling the `character` attribute as "deprecated"
- We merge this patch to [the `v2.10.0` feature branch](https://github.com/jshint/jshint/pull/3273) for inclusion in the next minor release of JSHint

Like the first draft you've submitted here, this approach fixes the problem for the folks that need it. On top of that, it avoids disrupting existing code, and it also gives us a clear path to the next major release, where the confusing `character` property can be removed entirely. (I kind of like the name "column" better anyway because "character" could describe a few things--a code point, its offset within the line, or even its offset within the entire input stream.)

So what do you think? Would you be up for that work? Depending on how you want to proceed, it may involve rebasing your branch. As noted in that "work-in-progress" pull request for version 2.10.0, I'd be happy to help you with that.

Arcanemagus:
I'd certainly be happy with moving to a different property in the result if it means being able to close the ~40 issues I've got open tracking this 😆.

cobexer:
Just fell over this issue again...

TzviPM:
Cleaning up my PRs and this doesn't look like it's going anywhere after 7 years 🤪

PR #3284, opened 2018-05-19
Improve handling of `let`

jugglinmike:
@rwaldron Since introducing support for `let` declarations, JSHint has
incorrectly disallowed use of the token `let` as an identifier outside of
strict mode. That's always kind of irked me, but I may be the only one (as far
as I know, we haven't received a bug report for this in the 5 years it's been a
problem). The number of Test262 failures this corrects is a bit underwhelming
:/

I also took the opportunity to move the parsing logic for Mozilla "`let`
expressions" out of the critical path of the parser. This is an internal
refactoring and not intended to change behavior. The change to error reporting
behavior is a side-effect of the second commit. As mentioned in the second
commit message, it only effects error messages and is therefore
backwards-compatible.

I'm finally getting to this because ES2017 introduces another "sort of
reserved" keyword: `async`. By introducing an extension to the Pratt parser
(the `useFud` method), this fix for `let` will simplify the patch we'll need to
introduce support for `async` functions correctly (that is: *without*
unconditionally reserving `async`). By the way: simplifying the changeset
necessary for async functions has been the motivation for many of my patches
since version 2.9.5.


src/jshint.js
```diff
@@ -4047,10 +4037,73 @@ var JSHINT = (function() {
...
+   * of the parser's state because this would allow JSHint to instruct users to
+   * enable the `moz` option where necessary. This is not possible because the
+   * language extension is not compatible with standard JavaScript. For
+   * example, the following program code may describe a "let block" or a
+   * function invocation:
+   *
+   *     let(x)
+   *     {
+   *       typeof x;
+   *     }
```
rwaldron: This is awesome.

tests/test262/expectations.txt
```diff
@@ -1021,7 +1017,6 @@ test/language/statements/let/global-use-before-initialization-in-declaration-sta
 test/language/statements/let/global-use-before-initialization-in-declaration-statement.js(strict mode)
 test/language/statements/let/global-use-before-initialization-in-prior-statement.js(default)
 test/language/statements/let/global-use-before-initialization-in-prior-statement.js(strict mode)
-test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-non-strict.js(default)
 test/language/statements/let/syntax/with-initialisers-in-statement-positions-label-statement.js(default)
 test/language/statements/let/syntax/with-initialisers-in-statement-positions-label-statement.js(strict mode)
 test/language/statements/let/syntax/without-initialisers-in-statement-positions-label-statement.js(default)
```
rwaldron: I'm adding this at the end, but applies to all changes in this file...


EXCELLENT.

rwaldron:
This is really smart—nice work!

PR #3287, opened 2018-06-10
[[FIX]] Remove warning W100

jugglinmike:
No justification could be found for this warning. See https://github.com/jshint/jshint/issues/2741

This should resolve gh-2741.


rwaldron:
I just want to acknowledge that I reviewed the decrease in coverage, but I don't see any issue that should prevent this from being merged.

Issue #2991, opened 2016-07-30
Object "spread" properties (stage 2 proposal) not supported


CoericK:
I got something like this

``` js
  args: {
    ...connectionArgs,
  },
```

And jshint its complaining about 
`xpected '}' to match '{' from line 167 and instead saw '...'.`

What should i add to my `.jshintrc`file to allow destructing?

jugglinmike:
JSHint does not currently support this proposed language feature. It recently reached "Stage 2" in the standardization process, though, so we are interested in adding support in a future version.

CoericK:
Oh thank you @jugglinmike, any idea about how can i make jshint ignore it at least?

jugglinmike:
Sure, you'll want to use [the in-line directives](http://jshint.com/docs/#inline-configuration) `jshint ignore:start` and `jshint ignore:end`:

```diff
  args: {
+   // jshint ignore:start
    ...connectionArgs,
+   // jshint ignore:end
  },
```

kevlarr:
Would be great to have JSHint support for this, since it's now Stage 3

derwaldgeist:
+1 for supporting this. It's already widely used in React/Redux

saidkholov:
+1

danorton2:
It's now at Stage 3: https://github.com/tc39/proposal-object-rest-spread

israbbani:
+1 for getting this done. Thanks guys!

dcantatore:
+1 would enjoy this very much

elmargasimov:
+1

leonardolessa:
+1

jugglinmike:
Support for this language feature has been merged to the `v2.10.0` feature branch (via gh-3150), so we expect it to be included in the next minor release of JSHint. I'll label this issue as "has PR" to better communicate this status.

iSkore:
Just a few days ago they announced:

>Support for this language feature has been merged to the v2.10.0 feature branch (via gh-3150), so we expect it to be included in the next minor release of JSHint. I'll label this issue as "has PR" to better communicate this status.

But this is how I fixed it:

Add *BOTH* `esnext` and `esversion` to your `.jshintrc` file:

```
...
"esnext": true,
"esversion": 6,
...
```

renphi:
+1 can't wait for v2.10.0 to be released. Good job! Is there any timeline for v2.10.0?

deadlyicon:
When I have both

```json
{
  "esnext": true,
[...]

PR #3290, opened 2018-06-12
[REFACTOR]: Warn instead of quitting when invalid options specified

caitp:
When the options "es3", "es5" or "esnext" are use, emit a warning that
the option is deprecated, and that you should use "esversion: <X>"
instead.

Previously, if you used "esnext" (or other deprecated options) in
addition to "esversion", jshint would refuse to valiate your code, and
quit instantly after reporting a weird error message that doesn't really
explain the change. The new error message, and warning-rather-than-quit
behaviour, should hopefully be easier for users to digest.

(Related to discussion on #2991)

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -189,16 +189,12 @@ var JSHINT = (function() {
    * Apply all linting options according to the status of the `state` object.
    */
   function applyOptions() {
-    var badESOpt = null;
     processenforceall();
 
     /**
      * TODO: Remove in JSHint 3
      */
-    badESOpt = state.inferEsVersion();
-    if (badESOpt) {
-      quit("E059", state.tokens.next, "esversion", badESOpt);
-    }
+    state.inferEsVersion();
 
     if (state.inES5()) {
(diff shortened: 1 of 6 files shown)


caitp:
Well, no, I don’t think insta-quitting is that great with deprecation.

App isn’t in ambiguous state, you always favour the non-ambiguous version if they’re both used, but the current case doesn’t even tell the user the feauture is deprecated, or inconsistent unless both flags are used.

There are alternatives, but I don’t think insta-quitting is the way to do this.

jugglinmike:
> App isn’t in ambiguous state, you always favour the non-ambiguous version if
> they’re both used

Trying to take the perspective of a user, I don't see a clear difference in
ambiguity between (for example) `es5: true` and `esversion: 6`. They both seem
equally descriptive.

> There are alternatives, but I don’t think insta-quitting is the way to do
> this.

My argument about parsing according to an assumption about the user's intent
didn't convince you, then. We should discuss those alternatives. Could you also
say more about why you are opposed to halting the parse?

PR #3283, opened 2018-05-19
[[CHORE]] Complete Lodash update

jugglinmike:
A previous commit modified the project's usage of Lodash in preparation
to update to a new major release of that library. That commit mistakenly
omitted the necessary modification for the `package.json` file. Complete
the dependency update by modifying the version specifier in that file.

PR #3291, opened 2018-06-14
Updated outdated dependencies

ryadiv:
Some libraries have been reported with security vulnerabilities, this update includes and an update for direct libraries use.


jugglinmike:
Thanks for the patch! This change has already been submitted via gh-3283

PR #3299, opened 2018-06-29
Modernizing jshint

jrgleason:
Removes JSCS and replaces with ESLINT per their site. I have to ignore two roles which we should fix later...

```
{
    "rules": {
        "no-control-regex": 0,
        "no-console": 0
    }
}
```


jrgleason:
Seems to work on 6 and higher

jrgleason:
@coveralls is wrong I will look into it to confirm but I might have decreased coverage by a few line but I also fixed linting (`npm run pretest`) which wasn't working...

```
45 code style errors found.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! jshint@2.9.5 pretest: `node ./bin/jshint src && jscs src`
npm ERR! Exit status 2
```

jrgleason:
Vs. Mine...

```
npm run pretest

> jshint@2.9.5 pretest /Users/jackiegleason/Code/jshint
> node ./bin/jshint src && eslint src
```

jrgleason:
The real issue with this is the loss of support for older versions. Can't you deprecate everything before 6 by now?

caitp:
> The real issue with this is the loss of support for older versions. Can't you deprecate everything before 6 by now?

it's hard to say --- I think we're probably OK to drop support for < Node LTS (or maybe the previous LTS release), but it would be good to make sure modern jshint continues to run out of the box on most distros, e.g. Ubuntu 18.10 currently includes Node 8~ (https://packages.ubuntu.com/cosmic/nodejs) --- but there may be good reasons to keep supporting ancient builds.

PR #3301, opened 2018-07-01
fixup! [[FEAT]] Enforce ES2016 restriction on USD

jugglinmike:
It turns out that the early error applies regardless of the strictness of the
surrounding code. So even though the second "use strict" directive in the
following example has no effect, the code is still syntactically invalid:

```js
"use strict";
function f(x=0) {
  "use strict";
}
```

I'm submitting this as a `fixup!` commit so we can merge it today and remember
to squash it when this branch is ready for `master`.


rwaldron:
@jugglinmike can you resolve the conflicts in `tests/test262/expectations.txt` and ping me for follow up review? Thanks!

jugglinmike:
Sure thing. Original version of this branch is available here:

https://github.com/jshint/jshint/compare/master...jugglinmike:fix-es2016-usd-error-prebase

rwaldron:
Excellent!

Issue #3304, opened 2018-07-12
Warn "misleading line break before" when array decomposition isn't done on declaration line ( with ASI )


paperrin:
JsHint version: 2.9.5

This code:
```
let a
if (true)
	({a} = {})
```
Gives result:
```
test.js|3 col 5 warning| Misleading line break before '('; readers may interpret this as an expression boundary
```

Same for this:
```
let a
if (true)
	[a] = []
```
Result:
```
test.js|3 col 5 warning| Misleading line break before '['; readers may interpret this as an expression boundary
```

With the config:
.jshintrc (json):
```
{
	"esversion": 6,
	"asi": true
}
```

Expected behavior:
This shouldn't give any warning and is perfectly fine.
If it is not, may I know why and how it's supposed to be done ? Thanks

I use this a lot, especially when creating variadic functions, and would prefer not to have to disable a warning, as I like having the most possible.

I could fix this doing:
```
let a
if (true) ({a} = {})
```
But I personnally find that to be really ugly, especially when there might be an else statement, which is the case when it's the most used for me..

jugglinmike:
Thanks for the report! I've submitted gh-3305 to fix this.

While we wait for that to be reivewed, merged, and published, you can also
supress this warning with the following change to your code:

```diff
 let a
-if (true)
+if (true) {
   ({a} = {})
+}
```

PR #3305, opened 2018-07-15
[[FIX]] Do not warn about non-ambiguous linebreaks

jugglinmike:
Resolve gh-3304. Commit message:

Warning W014 was implemented to alert users of potentially-confusing
formatting resulting from reliance on ASI.

The original heuristic was applied for every expression and simply
referenced token values. It assumed that when the previous token was a
closing parenthesis or bracket, this always signified a line of code
where the newline could be interpreted as a statement boundary, e.g.

    x = 3 * (2 + 1)
    [0]

However, this heuristic was susceptible to false positives, where the
closing punctuator would not be interpreted as the end of a statement:

    if (true)
      [x] = []

Re-implement the code that detects the warning condition from the
general "expression" function to the specific "infix" parsing logic
(used only when the "(" and "[" tokens have been identified as
components of an exiting expression).

(The first commit adds a few tests for existing behavior to ensure that the fix doesn't regress existing functionality.)

### DIFF
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -631,6 +631,21 @@ exports.safeasi = function (test) {
     .addError(10, 9, "Missing semicolon.")
     .test(src, { asi: true });
 
+  var afterBracket = [
+    'x = []',
+    '[1];',
+    'x[0]',
+    '(2);'
+  ];
+
+  TestRun(test, 'following bracket (asi: false)')
+    .test(afterBracket);
+
+  TestRun(test, 'following bracket (asi: true)')
+    .addError(2, 1, "Misleading line break before '['; readers may interpret this as an expression boundary.")
+    .addError(4, 1, "Misleading line break before '('; readers may interpret this as an expression boundary.")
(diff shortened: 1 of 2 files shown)


tests/unit/options.js
```diff
@@ -631,6 +631,39 @@ exports.safeasi = function (test) {
...
+    .test(afterBracket, { asi: true });
+
+  var asClause = [
+    'if (true)',
+    '  ({x} = {});',
+    'if (true)',
+    '  [x] = [0];',
+    'while (false)',
+    '  ({x} = {});',
+    'while (fale)',
```
rwaldron: I don't think it actually matters, but I figure I should at least call out this typo—in case you want to fix it?

jugglinmike: It probably doesn't matter, but I do want to fix it. Nice eye!

rwaldron:
Merged.

https://github.com/jshint/jshint/commit/ab3ab85e096476db7727aa55cfaad33b947288b7

PR #3307, opened 2018-07-16
[[FIX]] Correct restriction on function name

jugglinmike:
The ECMAScript specification details the following early error for
function-defining productions:

> - If the source code matching this production is strict mode code, it
>   is a Syntax Error if BindingIdentifier is present and the
>   StringValue of BindingIdentifier is "eval" or "arguments".

There were two bugs in the original implementation of this error:

- violations from function expressions with empty bodies would not be
  detected
- violations from function declarations contained within strict mode
  code would be reported twice

Correct the first issue by ensuring that the "(isStrict)" property is
accurately set for functions with empty bodies. Correct the second issue
by re-locating the error detection logic to within the processing logic
for expressions and declarations and inserting an additional condition
within the latter.

(This correction does not require modification to the Test262 whitelist
file, although that test suite does offer coverage for this behavior. At
the revision in use by JSHint, the relevant tests are expressed using
`eval` and therefore unusable by parsers. Those tests have since been
reformatted to correct this; they will be incorporated when JSHint next
updates its reference to Test262.)

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1764,6 +1764,9 @@ var JSHINT = (function() {
         metrics.statementCount += a.length;
 
         indent -= state.option.indent;
+      } else if (isfunc) {
+        // Ensure property is set for functions with empty bodies.
+        state.funct["(isStrict)"] = state.isStrict();
       }
 
       advance("}", t);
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3876,27 +3874,40 @@ var JSHINT = (function() {
...
       type: generator ? "generator" : null,
       ignoreLoopFunc: inblock // a declaration may already have warned
     });
+
+    // If the function declaration is strict because the surrounding code is
+    // strict, the invalid name will trigger E008 when the scope manager
+    // attempts to create a binding in the strict environment record. An error
+    // should only be signaled here when the function itself enables strict
+    // mode (the scope manager will not report an error because a declaration
+    // does not introduce a binding into the function's environment record).
```

PR #3312, opened 2018-07-22
[[FIX]] Relax restriction on asgnmnt to arguments

jugglinmike:
@rwaldron this is another another edge case in service of gh-3272

---

Previously, JSHint issued E031 ("Bad assignment.") when it encountered
assignment to a mapped `arguments` object. Not only is this message
vague, its implementation as an error incorrectly disallows
syntactically valid JavaScript.

Introduce a new warning which more clearly describes the rationale, and
emit the new warning in place of the existing error so that users may
opt out.

Although this change causes changes in messages reported by JSHint, it
is backwards-compatible because the prior behavior was implemented in
terms of an error. Existing users are therefore not reliant on the prior
message or code.


rwaldron:
👍

Issue #3318, opened 2018-08-03
Re-drop phantomjs dependency 


martinpitt:
https://github.com/jshint/jshint/pull/3215 moved phantomjs to an optionalDependency. This is rather evil, and conceptually completely unnecessary: it is only being used to test jshint itself, but it is in no way required for *users* of jshint.

This is dragging in an extra 66 MB download for `npm install` which is just dead weight. Note that phantomjs hasn't been maintained for a long time.

Please move it back to devDependency, which is the correct place. If someone wants to run tests with an ancient npm version, then they can just install phantomjs themselves IMHO. jshint shouln't penalize every user due to that.

Thanks for considering!

jugglinmike:
Thanks for taking the time to write this up! This problem has already been reported via gh-3316, so I'm going to close this as duplicate.

Issue #3316, opened 2018-08-01
jshint 2.9.6 node_modules folder uses 220MB disk space


xhronos:
An empty node project with jshint 2.9.5 uses about 6.8 MB disk space.
The node_modules folder using 2.9.6 (or ^2.9.5) consumes roughly **220 MB**.

Is this as expected?

(node.js version seems to have no effect)

jugglinmike:
Thanks for the report! A fix is on the way: https://github.com/jshint/jshint/pull/3319

You can expect a new release within a couple days.

eaviles:
Any updates on this @jugglinmike?

vorg:
@jugglinmike any blockers to quickly release 2.9.7?

alasdairhurst:
Waiting :)

jugglinmike:
[We've released the fix as JSHint 2.9.7.](https://github.com/jshint/jshint/releases/tag/2.9.7) My apologies for the delay.

PR #3260, opened 2018-03-11
[[CHORE]] Update to latest version of Lodash

jugglinmike:
Version 3 of the Lodash module contains a security vulnerability. That
vulnerability was patched in the 4.17.5 release.

Note that version 4 of Lodash removes support for Internet Explorer 6,
7, and 8. The JSHint project does not formally document compatibility
with web browsers, but the source is currently dependent on Array
prototype methods introduced in ES5. This means that updating Lodash
does not effect the set of browsers supported by this project.

[1] https://hackerone.com/reports/310443
[2] https://github.com/lodash/lodash/wiki/Changelog#v400

Signed-off-by: Mike Pennisi <mike@mikepennisi.com>

PR #3319, opened 2018-08-05
Reduce package size

jugglinmike:
We've received a couple reports (gh-3316 and gh-3318) that the installation size of JSHint increased drastically in the recently-released version 2.9.6.  This is easily verified using npm's `--production` flag:

    mkdir nodemods
    git checkout 2.9.5
    npm install --production
    mv node_modules nodemods/2.9.5
    git checkout -- package.json
    git checkout 2.9.6
    npm install --production
    mv node_modules nodemods/2.9.6
    du -h -d 0 nodemods/*

Output:

    9.9M    nodemods/2.9.5
    277M    nodemods/2.9.6

Two independent commits contributed to this 20-fold increase in side. This patch addresses each of them in turn. You can verify using the same technique:

    git checkout shrink-deps~
    npm install --production
    mv node_modules nodemods/2.9.6-phantom
    git checkout -- package.json
    git checkout shrink-deps
    npm install --production
    mv node_modules nodemods/2.9.6-phantom-unicode
    du -h -d 0 nodemods/*

Output:

    9.9M    nodemods/2.9.5
    277M    nodemods/2.9.6
    149M    nodemods/2.9.6-phantom
    18M     nodemods/2.9.6-phantom-unicode

The final size is still twice that of the prior release, but that is due to the upgrade to Lodash 4 (gh-3260 and gh-3283).

    9.9M    nodemods/2.9.6-phantom-unicode-lodash

Since that was motivated by security concerns, I'm more willing to accept the size increase. If this continues to be a concern for our consumers, we can consider instead relying on Lodash's sub-components.

Also of interest: the `remove-dev-dependencies.js` introduced here is not technically necessary on TravisCI because JSHint can now be successfully be installed in TravisCI on Node 0.8 and 0.10. That's surprising because this was all motivated by failures in TravisCI back in November. At first, I thought that maybe the maintainers of `requests` had changed their tune, but [no such luck](https://github.com/request/request/issues/2772). TravisCI now pre-installs PhantomJS, so the offending ES2015 code (used to fetch the binary) is no longer evaluated on any GNU/Linux build. However, we're not so lucky in the AppVeyor-provided Windows environment (and even if we were, we probably shouldn't rely on implicit aspects of the build environments). That's why I've written a script to manually remove the offending packages prior to installation in CI.

I'd like to release a new version (i.e. 2.9.7) for this changeset so that we can help our consumers avoid the inefficiency as soon as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: node_js
 after_success: npm run coverage && cat ./coverage/lcov.info | coveralls
+before_install:
+  # These packages cannot be installed in legacy versions of Node.js. Because
+  # they are only needed to verify correctness in a browser-like environment,
+  # they can be removed (and the tests skipped) without effecting coverage.
+  - if [[ $(node --version) != v6* ]] ; then
+      node scripts/remove-dev-dependencies.js phantom phantomjs-prebuilt;
+    fi
 script:
   - npm run pretest
   - npm run test-node
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,10 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - git submodule update --init --recursive
+  # These packages cannot be installed in legacy versions of Node.js. Because
+  # they are only needed to verify correctness in a browser-like environment,
+  # they can be removed (and the tests skipped) without effecting coverage.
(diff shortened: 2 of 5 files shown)


rwaldron:
Wow, nice work to solve this!

PR #3336, opened 2018-11-04
[[FIX]] Disallow fn declarations in stmt positions 

jugglinmike:
Two commits here. The first is a refactoring in support of the second. As a bonus, it also reduces global state.


jugglinmike:
Thanks for the help, @rwaldron!

(To merge this, Rick had to resolve a few conflicts in the Test262 "expectations" file. He pushed the result from his machine to the feature branch directly, so I'll follow up by closing this pull request.)

Issue #3340, opened 2019-01-02
W039 never triggered?


pamelafox:
I'm currently going through JSHint error messages to determine which of them ever show up in our use of JSHint on Khan Academy. 

I notice that W039 is only defined in the messages list, and never seems to be triggered from the parser.

We do still have a trigger for it in our old JSHint copy (https://github.com/Khan/live-editor/blob/master/external/jshint/jshint.js), but I'm thinking perhaps it went away in yours at some point?

Either that, or my git repo searching isn't yielding all results. 

Just an FYI in case it's easy dead code to remove.

pamelafox:
Same question regarding W105. Only in messages, not triggered from parser, and no evidence of a test for it.

jugglinmike:
git log's `-S` flag is perfect for questions like these:

    $ git log -SW039 --oneline
    925a706 v2.9.0
    08eb4e2 [[Fix]] Scope issues by refactoring
    adbaf51 No issue: track dist/ directory.
    83374ad No issue: Remove /stable/ subdirectories
    8683b08 Converted all messages to codes.
    6dc73c4 Extracted all warnings and error messages into a separate files.

The warning appears to have been unintentionally removed in commit 08eb4e2, and we subsequently released that removal in version 2.9.0. We apparently didn't have any tests for it (otherwise it wouldn't have passed continuous integration).

The story around W105 is a little more satisfying:

    $ git log -SW105 --oneline
    e38b8a0 2.5.0
    c09706c Remove 'nomen'
    adbaf51 No issue: track dist/ directory.
    83374ad No issue: Remove /stable/ subdirectories
    35b88fc Refactoring: more robust tokenizer, modules, etc.
    684b379 Extract regular expressions, lexer and some state variables.
    8683b08 Converted all messages to codes.
    6dc73c4 Extracted all warnings and error messages into a separate files.

That warning *did* have tests, and its removal was clearly intentional. Looks like Anton just forgot to remove the message.

The resolution is the same for both cases: remove the message. See gh-3341.

Thanks for the report!

pamelafox:
Okay, cool, thanks for the reply!

PR #3341, opened 2019-01-06
Remove unused warnings

jugglinmike:
This should resolve gh-3340


rwaldron:
Who doesn't love some good house keeping.

PR #3343, opened 2019-01-21
fixup! [[FEAT]] Implement support for ES8 trailing commas

jugglinmike:
The proposal's name threw me for a loop: I didn't realize it also added support for trailing commas in arguments lists. I'll fix this up into the referenced commit after we merge it.


rwaldron:
Merged locally, pushed to origin

PR #3344, opened 2019-01-21
[[FEAT]] Enable object rest/spread via esversion

jugglinmike:
The "Object rest/spread properties" feature was standardized in ES2018.
Update enabling mechanism to use a new `esversion` value dedicated to
that edition.


rwaldron:
Rebased, merged and pushed

PR #3348, opened 2019-01-28
[[FEAT]] Implement `regexpu` option

jugglinmike:
@caitp thought you might be interested in this since it's in a similar vein to [the `varstatement` option you implemented](https://github.com/jshint/jshint/pull/2234). It could probably use a better name, though.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -236,6 +236,15 @@ var JSHINT = (function() {
       }
     }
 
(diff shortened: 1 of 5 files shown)


rwaldron:
Conflicts resolved, merged into v2.10.0

Note that the warning number had to be changed from W146 to W147

PR #3349, opened 2019-01-28
[[FEAT]] Implement ES2019 RegExp "dotall"

jugglinmike:
Introduce a new warning to alert users of RegExp literals that include
the new flag without using the "dot" atom.


rwaldron:
Conflicts resolved, merged into v2.10.0

Note that the warning number had to be changed from W146 to W148

Issue #3354, opened 2019-02-05
2.10.0 marks variables as unused when they are not


bdwain:
I have a file that looks like this

```js

import {foo} from './foo';

class MyClass {
  foo(){
     return foo();
  }
}
```

and it is incorrectly marking foo as unused. This is a valid use since the call to `foo` is not referring to `this.foo` so it must be the imported `foo`

jugglinmike:
Confirmed. Thanks for the report!

jugglinmike:
Introduced via 6759ac902928c32e0c798134abe1fbd1353ad3a3

jugglinmike:
Further reduced:

    // jshint esversion: 6
    var m;
    class C {
      m() {
        void m;
      }
    }

This effects:

- methods
- static methods
- accessor methods
- genereate methods
- async methods
- async generator methods

This does *not* effect constructor methods, e.g.

    // jshint esversion: 6
    var constructor;
    class C {
      constructor() {
        void constructor;
      }
    }

jugglinmike:
Fix on the way: gh-3355

PR #3355, opened 2019-02-05
[[FIX]] Do not add cls method names to env. record

jugglinmike:
A recent refactoring of the class parsing logic [1] removed the function
name inference logic (i.e. integration with the `nameStack` object) in
favor of a similar codepath within the generic `doFunction` routine. By
providing the method name to that routine, the change had an unintended
side effect: method names were used to create new entries in the new
environment record. This was observable from user code in the presence
of JSHint's `unused` linting option: when set, methods would incorrectly
shadow variables that shared the same name.

Continue to rely on the `doFunction` method for name inference (limiting
duplication), but disable the extension of the environment record for
that case.

This is intended to resolve gh-3354.


rwaldron:
Cool bug and nice fix :D

Issue #3370, opened 2019-02-24
unexpected 'i' was used before it was declared, which is illegal for 'let' variables.


Yaffle:
```
  function f() {
    "use strict";
    for (let i = 1, j = i; j < 100; j += i, i = j - i) {
      console.log(i);
    }
  }
```

jugglinmike:
Thanks for the report! Here's a further-reduced example:

    // jshint esversion: 6
    for (let x, y = x; ;) void y;

This is a regression in version 2.10.0. It's limited to `for` statements; JSHint 2.10.1 continues to parse variable declarations correctly:

    // jshint esversion: 6
    let x, y = x;
    void y;

Version 2.10.0 introduced a lot of new features, so folks may not want to wait for a fix. One short-term workaround would be to declare the bindings outside of the `for` statement, adding an additional block scope if that's important. In terms of Viktor's original example:


```diff
 function f() {
   "use strict";
+  {
+    let i, j;
-    for (let i = 1, j = i; j < 100; j += i, i = j - i) {
+    for (i = 1, j = i; j < 100; j += i, i = j - i) {
       console.log(i);
     }
+  }
 }
```

That's far from ideal, so I'm prioritizing the fix!

Yaffle:
https://github.com/jshint/jshint/blob/master/src/jshint.js#L4337
noin is `2` here and so this block will not call `state.funct["(scope)"].initialize(t.id);`, and the error is reported from https://github.com/jshint/jshint/blob/master/src/scope-manager.js#L975 , I think.

Although, the code looks strange, it doesn't check if there is an initialization...
Seems, I cannot fix it.

jugglinmike:
Thanks for taking a closer look! I just submitted a patch to fix this: gh-3376; more detail is available there

PR #3376, opened 2019-03-03
Correct parsing of block binding declarations in C-style for loops

jugglinmike:
The first commit is intended to resolve gh-3370. The second fixes an as-yet unreported bug.


jugglinmike:
Thanks, @rwaldron!

PR #3388, opened 2019-04-15
Streamline "advance" function

jugglinmike:


src/jshint.js
```diff
@@ -2562,11 +2549,45 @@ var JSHINT = (function() {
...
+    var next = state.tokens.next;
+    this.arity = "unary";
+    this.right = expression(context, 150);
+
+    if (next.id === "-" || next.id === "--") {
+      warning("W006", this.right);
+    }
+
+    return this;
+  });
```
rwaldron: I'm so tempted to ask for some abstraction for all of these, but I think that I will just accept that sometimes it's nice to just have a clear view of the operations.

PR #3390, opened 2019-04-18
Coverage: add "istanbul ignore next" to lines that are not currently covered by tests

rwaldron:


rwaldron:
Rebased

src/jshint.js
```diff
@@ -3008,6 +3027,7 @@ var JSHINT = (function() {
         if (left.value.match(/^[A-Z]([A-Z0-9_$]*[a-z][A-Za-z0-9_$]*)?$/)) {
           if ("Array Number String Boolean Date Object Error Symbol".indexOf(left.value) === -1) {
             if (left.value === "Math") {
+              /* istanbul ignore next */
               warning("W063", left);
```
rwaldron: @jugglinmike I suggest that we update "W063" to be `"{a} is not a function."` and add `left.value` as a third parameter here. Then we can catch Atomics, JSON and Reflect

rwaldron:
> Coverage increased (+1.4%) to 99.978% 

Lol.

rwaldron:
@jugglinmike the additional changes are 👍

jugglinmike:
Oops, I wrote a comment offline, but I forgot to post it:

> @rwalrdon one of the Istanbul directives was slightly off. I corrected that and also reduced a bit of the repetition. Could you add `[[CHORE]]` to the commit message when you squash and merge?

Since you noticed and approved the new commits even without my comment, I'm happy to follow through on the merging.

jugglinmike:
Squashed and merged to `master` at commit 963d38d4c4f1aa31c373b64dbf08820f211b998b. Thanks, Rick!

PR #3332, opened 2018-10-06
Add support for typeof bigint === "bigint". Fixes #3331

kawanet:
Limitation: This PR fixes `Invalid typeof value 'bigint'`, but does NOT support `BigInt64Array` nor `BigUint64Array` still.

```
$ echo 'var a = "bigint" === typeof {};' | node bin/jshint -
stdin: line 1, col 18, Invalid typeof value 'bigint'
1 error

$ echo '/* jshint esnext: true */ var a = "bigint" === typeof {};' | node bin/jshint - && echo OK
OK
```

PR #3394, opened 2019-04-29
[[FEAT]] Add experimental support for BigInt

jugglinmike:
Following through with work originally started by @kawanet in gh-3332.

@rwaldron any thoughts for things we should lint for? Best to do that now since
any ideas we have after we initially release support will have to be disabled
by default.


jugglinmike:
@rwaldron This was originally reported as regressing coverage because I forgot
to update the coverage command to also use the new directory for experimental
tests. I've fixed that and force-pushed the result.

If you don't mind, could you use a merge commit for this one? The individual
commits are important for attribution purposes, but they aren't atomic as far
as the feature is concerned.

package.json
```diff
@@ -21,7 +21,7 @@
   "scripts": {
     "browser-test-server": "node tests/helpers/browser/server",
     "build": "node bin/build",
-    "coverage": "istanbul -- cover ./node_modules/nodeunit/bin/nodeunit tests/unit",
+    "coverage": "istanbul -- cover ./node_modules/nodeunit/bin/nodeunit tests/unit tests/unit/unstable",
```
rwaldron: Nice

rwaldron:
Looks like this didn't auto close :|

PR #3397, opened 2019-05-05
[[TEST]] Improve coverage

jugglinmike:
Add test for handling of invalid ImportsList (without the branch under
test, no error would be reported when ImportSpecifiers are not
delineated by a comma).


rwaldron:
Who doesn't love "Improve coverage"?!

PR #3400, opened 2019-05-19
Improve internal variable names

jugglinmike:
JSHint's "scopeManager" utility was initially implemented to refer to ECMAScript bindings using the term "label." Because ECMAScript uses the term "label" to describe a completely different language feature, this choice of names was somewhat confusing.

Modify all references to "JSHint labels" to use the more accurate term, "binding." Shorten the term "break label" to simply "label."

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -1579,7 +1579,7 @@ var JSHINT = (function() {
       return true;
     } else if (left.identifier && !isReserved(context, left) && !left.isMetaProperty &&
       left.value !== "eval" && left.value !== "arguments") {
-      if (state.funct["(scope)"].labeltype(left.value) === "exception") {
+      if (state.funct["(scope)"].bindingtype(left.value) === "exception") {
         warning("W022", left);
       }
       state.nameStack.set(left);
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -1894,7 +1894,7 @@ var JSHINT = (function() {
 
       hasOwnScope = true;
       state.funct["(scope)"].stack();
-      state.funct["(scope)"].block.addBreakLabel(t.value, { token: state.tokens.curr });
+      state.funct["(scope)"].block.addLabel(t.value, { token: state.tokens.curr });
```
rwaldron: Because it's not always for `break`!

src/jshint.js
```diff
@@ -2734,7 +2734,7 @@ var JSHINT = (function() {
       className = classNameToken.value;
       identifier(context);
       // unintialized, so that the 'extends' clause is parsed while the class is in TDZ
-      state.funct["(scope)"].addlabel(className, {
+      state.funct["(scope)"].addbinding(className, {
```
rwaldron: This has always been confusing.

src/scope-manager.js
```diff
@@ -36,9 +30,9 @@ var scopeManager = function(state, predefined, exported, declared) {
 
   function _newScope(type) {
     _current = {
-      "(labels)": Object.create(null),
+      "(bindings)": Object.create(null),
       "(usages)": Object.create(null),
-      "(breakLabels)": Object.create(null),
+      "(labels)": Object.create(null),
```
rwaldron: I can see clearly now.

PR #3418, opened 2019-07-21
[[FEAT]] Add support for ES2019 opt. catch param

jugglinmike:


src/jshint.js
```diff
@@ -707,13 +707,15 @@ var JSHINT = (function() {
...
           case "9":
+          case "10":
             state.option.moz = false;
             state.option.esversion = +val;
             break;
           case "2015":
           case "2016":
           case "2017":
           case "2018":
+          case "2019":
```
caitp: side note: I wonder if we should do something similar to babel, marking individual features as allowed/disallowed, and having modules containing presets based on spec versions. These numbers mean less each year in the new living standard model.

rwaldron:
@jugglinmike I only quickly looked at the failures and noted that they are definitely coming from the tests themselves

jugglinmike:
@rwaldron Right you are. I've pushed up a correction

src/jshint.js
jugglinmike: @caitp When you say "modules," do you mean "npm modules" (i.e. published separately) or "Node.js modules" (i.e. published within this project)?

rwaldron:
- Rebased fixup!
- Merged into v2.11.0: https://github.com/jshint/jshint/commit/ec71bc06e8c5b0db168958423e0cf30aa01726c4

PR #3419, opened 2019-07-28
Re-implement "comma fixes"

jugglinmike:
@rwaldron @caitp Writing something truly distinct was a little tough for this small patch, but the small size fortunately also makes originality less critical. I'm hoping that my discovery and correction of a previously-unreported bug lends some weight to the novelty of the new implementation.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3175,10 +3175,9 @@ var JSHINT = (function() {
       return;
     }
 
-    exprs[exprs.length - 1].parent = true;
-
     if (exprs.length > 1) {
-      ret = Object.create(state.syntax[","], { exprs: { value: exprs } });
+      ret = Object.create(state.syntax[","]);
+      ret.exprs = exprs;
 
       first = exprs[0];
       last = exprs[exprs.length - 1];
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -5227,7 +5244,7 @@ var JSHINT = (function() {
     if (!state.option.asi)
       nolinebreak(this);
 
-    if (state.tokens.next.id !== ";" && !state.tokens.next.reach &&
+    if (state.tokens.next.identifier &&
```
caitp: Does this change to ASI handling work properly if the next line begins with a pre-increment or pre-decrement?

caitp:
other than the ASI handling question, lgtm

src/jshint.js
rwaldron: @caitp I'm trying to think of an example, but maybe I'm misunderstanding you?

```js
var x = 0;
a: {
  break a
  --x
}
```

That doesn't seem right

caitp: I couldn't see that this was for stmt("break") from mobile, I guess it makes more sense now that I can see it

PR #3424, opened 2019-08-18
[[FIX]] Validate lone arrow function parameter

jugglinmike:
Previously, the identifier parsing logic returned early for a lone arrow
function parameter in order to avoid incorrectly extending the
environment record in which the associated arrow function was defined.
This unintentionally circumvented logic which enforced restrictions on
identifier names.

Refactor the parsing logic to enforce the restriction in all cases, but
only alter the environment record when the identifier is not a lone
arrow function parameter.


rwaldron:
Excellent!

PR #3427, opened 2019-09-15
Re-implement `yield` parsing improvements

jugglinmike:
@caitp @rwaldron This is the latest in the series of rewrites (the prior patches along these lines were 6759ac902928c32e0c798134abe1fbd1353ad3a3 and gh-3419).

Since the first commit intentionally regresses functionality, we should definitely use a merge commit to land this branch.

Commit messages:

>     [[FIX]] Improve parsing of `yield`
>
>     Re-implement the improvements reverted by the previous commit. Use
>     distinct logic to differentiate this new implementation, including the
>     following substantive improvements:
>
>     - Correct the reported location of warning W093 ("Did you mean to return
>       a conditional instead of an assignment?")
>     - Permit the use of `yield` as an identifier outside of strict mode and
>       generator function bodies
>     - Remove inaccurate restriction on `yield` within Mozilla generator
>       functions
>     - Forward the context when parsing computed property names
>
>     [[CHORE]] Revert `yield` expression improvements
>
>     JSHint's parsing of `yield` expressions is derived from code first
>     contributed by GitHub user "usrbincc." That user could not be reached to
>     sign JSHint's Contributor License Agreement. Revert the user's
>     contribution to the application logic (preserving the corresponding
>     MIT-licensed test code) so that it may be re-written under the terms of
>     the Contributor License Agreement.
>
>     Revert the changes introduced by
>     88c862df3dba9e2cfa1e44d4be909099d8306c97


jugglinmike:
I've rebased this against the `v2.11.0` branch so that we can trial it via a release candidate.

jugglinmike:
@rwaldron I'm thinking that since no one is actively working on the `v2.11.0` feature branch, I should rebase that on `master` before we merge this. What do you think?

rwaldron:
@jugglinmike I trust your intuition 👍

jugglinmike:
That's much appreciated, Rick. This is rebased and ready for review.

PR #3439, opened 2020-01-05
Refactor condition expression to remove unnecessary isString function

michabotti123456:


jugglinmike:
Thanks for bringing this to my attention, @michabotti123456

@rwaldron it looks like you wrote this patch back in April and pushed it upstream. Do you remember why you didn't submit it?

PR #3461, opened 2020-04-04
docs: Fix simple typo, overriden -> overridden

timgates42:
There is a small typo in tests/cli.js, tests/regression/libs/backbone.js.

Should read `overridden` rather than `overriden`.

### DIFF
--- a/tests/cli.js
+++ b/tests/cli.js
@@ -299,7 +299,7 @@ exports.group = {
     ]);
     test.ok(rep.reporter.args[0][0].length === 0);
 
-    // Test overriden, failed file
+    // Test overridden, failed file
     cli.interpret([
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch! We can forgo the CLA because this contribution is non-substantive.

There's one change I'd like to remove from this patch. `backbone.js` is a third-party library which we've included in this project for the purposes of regression testing. While the modification undeniably corrects a typo, it's not code we maintain here, so I'd rather not make any modifications to it. Can you revert that change?

timgates42:
Thanks @jugglinmike - I've hopefully done this now - let me know if any other corrections are necessary.

jugglinmike:
Thank you very much!

PR #3468, opened 2020-04-26
[[DOCS]] Consolidate information on workaround

jugglinmike:
Expand explanation on the special treatment of the `__proto__` property
and consistently reference the explanation from every location where it
is relevant.

### DIFF
--- a/src/scope-manager.js
+++ b/src/scope-manager.js
@@ -1,4 +1,20 @@
 "use strict";
+/**
+ * A note on `__proto__`:
+ *
+ * This file uses ordinary objects to track identifiers that are observed in
+ * the input source code. It creates these objects using `Object.create` so
+ * that the tracking objects have no prototype, allowing the `__proto__`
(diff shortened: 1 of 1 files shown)


src/scope-manager.js
```diff
@@ -1,4 +1,20 @@
...
+ * property to be used to store a value *without* triggering the invocation of
+ * the built-in `Object.prototype.__proto__` accessor method. Some environments
+ * (e.g. PhantomJS) do not implement the correct semantics for property
+ * enumeration. In those environments, methods like `Object.keys` and Lodash's
+ * `values` do not include the property name. This file includes a number of
+ * branches which ensure that JSHint behaves consistently in those
+ * environments. The branches must be ignored by the test coverage verification
+ * system because the workaround is not necessary in the environment where
+ * coverage is verified (i.e. Node.js).
+ */
```
rwaldron, line 17: 👍🏼

PR #3484, opened 2020-06-15
[[FIX]] Limit positions for class declarations

jugglinmike:
JSHint already includes logic to disallow `let` and `const` declarations
in certain statement positions. Extend the definition of `class` tokens
so that it applies the same restrictions to class declarations.


jugglinmike:
Totally. Changes like this are great because someone else has already done the hard work; we just have to opt more code in to it :)

PR #3486, opened 2020-06-20
Version 2.13.0

jugglinmike:
This is a tracking branch for the next minor release of JSHint. If you'd like
to implement a new feature, this is the place to submit it!

ES2020 features:

- [x] [Dynamic import](https://github.com/tc39/proposal-dynamic-import)
- [x] [`import.meta`](https://github.com/tc39/proposal-import-meta)
- [x] [BigInt](https://github.com/tc39/proposal-bigint)
- [x] [`globalThis`](https://github.com/tc39/proposal-global)
- [x] [Optional chaining](https://github.com/tc39/proposal-optional-chaining)
- [x] [Nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing)
- [x] [`export * as ns from "mod"`](https://github.com/tc39/ecma262/pull/1174)

PR #3490, opened 2020-07-02
[[FEAT]] Add support for `import.meta`

jugglinmike:
Tolerate parsing errors regarding escape sequences in IdentifierNames as
these reflect a pre-existing deficiency which will require a dedicated
patch to correct.


rwaldron:
@jugglinmike I've rebased this for you and merged it into v2.12.0

akvadrako:
It looks like this was never merged.

jugglinmike:
You're right, @akvadrako! I've cherry-picked it back into [the feature branch for version 2.13](https://github.com/jshint/jshint/pull/3486); that's not apparent on this page because GitHub isn't aware of the change, but you can verify by checking the commit history on that branch.

Thanks for keeping an eye out!

PR #3493, opened 2020-07-17
improved language to be more inclusive of less privileged groups (blm)

rgeerts:
wasn't real sure if one of the changes I made is a "breaking change" or not.  but most of this is just renaming of variables.

Thanks for considering.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -195,7 +195,7 @@ var JSHINT = (function() {
 
   function combine(dest, src) {
     Object.keys(src).forEach(function(name) {
-      if (_.has(JSHINT.blacklist, name)) return;
+      if (_.has(JSHINT.denylist, name)) return;
       dest[name] = src[name];
     });
   }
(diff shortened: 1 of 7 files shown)


src/jshint.js
```diff
@@ -6296,7 +6296,7 @@ var JSHINT = (function() {
     } else {
       JSHINT.errors = [];
       JSHINT.internals = [];
-      JSHINT.blacklist = {};
+      JSHINT.denylist = {};
```
jugglinmike, line 6299: This is a breaking change. We're not ready for a major release, but what we *can* do in the mean time is label `JSHINT.denylist` as "deprecated" and offer the more inclusive API as an alternative. So, `JSHINT.blacklist = JSHINT.denylist;`.

rwaldron:
ping @rgeerts

rgeerts:
> ping @rgeerts

sorry for the delay... I certainly agree with your suggestion, it will take me some time to understand more of what's going on to make this change

PR #3503, opened 2020-09-19
Simplify treatment of property names

jugglinmike:


rwaldron (approved):
Excellent refactoring!

PR #3504, opened 2020-09-21
Improve handling of exported bindings

jugglinmike:
It might not look like it, but this is in support of ES2020's new production, `export default as ns from '';` (tracking via gh-3486).


rwaldron (approved):
Love seeing the expected failures file shrinking.

PR #3507, opened 2020-10-02
Fixes markdown syntax

torn4dom4n:
I fix some markdown syntax.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -88,7 +88,7 @@ a file path to another configuration file, and the path should be relative to
 the current file.
 
 For example, you might define a `.jshintrc` file in the top-level directory of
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch! The "footnote" syntax is not recognized by the project website's Markdown processor, so it won't have the intended effect. Could you remove those changes?

torn4dom4n:
The `relicensing-2020.md` is not recognized but the changes easy to observe. I think it is necessary.

jugglinmike:
Please do. The website build system will not produce the intended markup from a file which uses that syntax.

torn4dom4n:
@jugglinmike I reverted changes in the `relicensing-2020.md`

jugglinmike:
Thanks!

PR #3513, opened 2020-10-25
[[FIX]] Improve declaration parsing

jugglinmike:
Add an internal test so the coverage analysis tool recognizes that the
modified branch is verified.


rwaldron (approved):
Excellent!

Issue #3495, opened 2020-07-28
Setting browser:true in .jshintrc does not cover some globals


rbruckheimer:
The globals "customElements" and "ResizeObserver" ought to be added to the exports.browser object in vars.js.

semmel:
Yes indeed, I need to tend an ever growing list of globals in my .jshintrc file:
```
"predef": [ "AbortController", "crypto", "DOMException", "NetworkInformation", "customElements", "URLSearchParams"]
```
It's a little sad. 😢

jugglinmike:
Sounds about right to me. This is a [good first issue](https://github.com/jshint/jshint/labels/good%20first%20issue)--@rbruckheimer has explained exactly what needs to be done.

JuanFML:
Hi I am new here I could add try to do this as my first PR, but do I only have to add "customElements" and "ResizeObserver", to  the exports.browser object in vars.js and set them as false? Or is something else needed first?

semmel:
@JuanFML please add all [DOM interfaces](https://html.spec.whatwg.org/multipage/indices.html#all-interfaces) and [Window globals](https://developer.mozilla.org/en-US/docs/Web/API/Window)

That would be great!

tusharkrbarman:
@jugglinmike I am new to Open source, Please assign this issue to me

PR #3522, opened 2020-12-17
[[FEAT]] Added window globals and DOM interfaces to browser

JuanFML:
Added DOM interfaces and Window globals into the exports.browser object in vars.js. In order to cover more globals regarding issue #3495 .


JuanFML:
Can somebody help me with the errors? I don't understand where the problem is.

jugglinmike:
Thanks for the patch! The build errors were due to an unrelated problem with
one of this project's transitive dependencies. [I just fixed
that](https://github.com/jshint/jshint/pull/3523), though, so things should
work as intended if you push another commit to this branch.

As for this patch, it introduces a good deal more bindings than what has been
requested in gh-3495. Some of them, like `toolbar`, are non-standard and
deprecated. Other bindings, like `WindowGlobalScope` are not defined by any
browser. JSHint ought to issue warnings when either of those kinds of bindings
are used, so they shouldn't be listed in the `vars.js` file.

In order to avoid confusion like this, we've taken to annotating new additions
with code comments which explain where the bindings have been standardized. If
you could do the same here, it would be more clear why each of the new bindings
is necessary.

jugglinmike:
@JuanFML It's been a few months since we heard from you. Are you still interested in persuing this patch?

JuanFML:
Hello @jugglinmike I am sorry I haven’t been able to give it some time, if anyone wants to finish it or should we close the PR?

Ashesh3:
@jugglinmike @JuanFML I would like to take this up and finish it. Before i proceed though, i would want to confirm how exactly you want the new additions to be annotated?

```
AudioTrack, // https://html.spec.whatwg.org/multipage/media.html#audiotrack
DOMParser, // https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#domparser
.....
```

Is it like above or

```
//Taken from https://html.spec.whatwg.org/multipage/indices.html#all-interfaces
AudioTrack,
DOMParser
....
```
Or like this? Or any other format. Please let me know and i would try to finish this PR.

jugglinmike:
The former approach seems good to me!

PR #3529, opened 2021-02-03
[[Fix]] Don't warn when RegExp() is used without 'new'.

lauriro:
https://262.ecma-international.org/#sec-regexp-constructor
...
function call RegExp(…) is equivalent
to the object creation expression new RegExp(…)
with the same arguments
...

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3046,14 +3046,17 @@ var JSHINT = (function() {
 
     if (left) {
       if (left.type === "(identifier)") {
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch--this is a good idea!

It's a bit more refactoring than I'd typically like to see in a bug fix, but I'm not particularly fond of how the code is currently written, so I'll take it :)

lauriro:
It was a bit more refactoring than I would liked to do but just extending the string triggered line length error.

.. and string.indexOf would give false-positives for `Arr()`, `Boo()`, and so on.

After pull request I found that `Function` is also missing.

PR #3531, opened 2021-02-08
[[FIX]] Don't warn when Function() is used without 'new'.

lauriro:
https://262.ecma-international.org/#sec-function-constructor
...
the function call Function(…) is equivalent to the object creation
expression new Function(…) with the same arguments
...


jugglinmike:
Very nice. Thanks!

Issue #3551, opened 2021-07-07
Optional chaining reports unorthodox function invocation error in specific case


ashwinbaliga:
Using the latest JSHint 2.13.0 that has optional chaining support (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), I see that a specific line wrongly reports an "unorthodox function invocation" error:

```
/* jshint esversion:11 */
var someChainingObj = {aNum:10,
                       aFunc: function() { return 500; },
                       anArr: ['val1', 'val2'] };

alert(someChainingObj?.aFunc?.()); // Reports unorthodox function invocation error
alert(someChainingObj.aFunc?.()); // Works fine

alert(someChainingObj?.anArr?.[1]); // Works fine
alert(someChainingObj.anArr?.[1]); // Works fine

alert(someChainingObj?.aNum?.isPrototypeOf); // Works fine
alert(someChainingObj.aNum?.isPrototypeOf); // Works fine

```

The error is only for the above mentioned line for "aFunc". It does not happen for arrays and for objects. I don't see anything wrong in the line that reports the above error.

jugglinmike:
Thanks for the report! The issue is the CallExpression, so we can futher reduce the problem like this:

```js
/* jshint expr: true, esversion: 11 */

Math?.min();
```

Working on a fix, now.

jugglinmike:
Fix under review here: gh-3552

jugglinmike:
[We released the fix in JSHint version 2.13.1.](https://jshint.com/blog/2021-08-10/release-2-13-1/)

PR #3552, opened 2021-07-11
[[FIX]] Allow invoking result of optional chaining

jugglinmike:
This should resolve gh-3551.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3081,6 +3081,21 @@ var JSHINT = (function() {
     return that;
   }, 160, true);
 
+
+  /**
+   * Determine if a CallExpression's "base" is a type of expression commonly
+   * used in this position.
+   *
+   * @param {token} token - token describing the "base" of the CallExpression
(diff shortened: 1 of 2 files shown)


rwaldron (approved):

ehoogeveen-medweb:
Does this also fix the warning `Expected an assignment or function call and instead saw an expression. (W030)` for expressions like `obj.func?.()` ("call this function if it exists")? It seems odd to me that `obj.func()` is allowed as an expression but `obj.func?.()` isn't.

jugglinmike:
@ehoogeveen-medweb That behavior isn't related to this fix. Your expectation seems logical to me, though, so I'll write a patch for that, too.

jugglinmike:
Here's the patch: gh-3557

jugglinmike:
[We released both fixes in JSHint version 2.13.1.](https://jshint.com/blog/2021-08-10/release-2-13-1/)

PR #3557, opened 2021-08-05
[[FIX]] Allow optional chaining call as satement

jugglinmike:
Take care to only disable warning W030 for the "function call" form of
the optional chaining feature; the warning is still relevant for
property access.

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -3070,6 +3070,7 @@ var JSHINT = (function() {
       that.left = left;
       advance();
       that.right = state.tokens.curr.led(context, left);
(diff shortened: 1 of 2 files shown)


src/jshint.js
```diff
@@ -3070,6 +3070,7 @@ var JSHINT = (function() {
       that.left = left;
       advance();
       that.right = state.tokens.curr.led(context, left);
+      that.exps = true;
```
rwaldron, line 3073: Lol, too easy.

rwaldron (approved):

PR #3562, opened 2021-08-18
[[FIX]] Limit "Too many Errors" (E043) to errors only

### LINKED ISSUE #3444: showing Error (E043) on "Too many errors", when those "Errors" are only warnings
There should be a separate counter for warnings and for errors.
And separate giving up actions and messages: "Too many Errors" and "Too many Warnings"

I came accross two cases where this is specially important.

I use `debug['something']` throughout the code. There is a reason for that and it's valid. I hid the warning. Not by jshint options. Now all I see is one error saying there is too many errors.
It was quite a time-waster issue finding out what was going on.

Yes, I tried // jshint -W069 instead and the error was not there anymore, probably warnings hidden by jshint options are not counted.
But it does not mean that the logic of a common counter is ok.
Many times jshint is not used directly and warnings could be hidden upstream.

A lot of scripts use jshint to check user-added scripts for errors before executing them in its environment.
They should be set to ignore all warnings or set the maxerr too high. Tamper Monkey seems to be doing that. But it is not something to be expected naturally.

There is a closed issue: https://github.com/jshint/jshint/issues/180,
but the proposed solutions were to remove the feature, or make it opt-in.

What I mean here is that the feature itself is not an issue, but joining counts of warnings and errors and reacting with one common action is.

Ashesh3:
As mentioned in #3444 , `E043` is triggered with "Too many Errors" even if there are just warnings in the checked code. This happens because the condition to trigger the said `E043` error is:

```
if (JSHINT.errors.length >= state.option.maxerr)
      quit("E043", t);
```
Which clearly checks for the total count of all the messages in `JSHINT.errors` array, The issue is that, this array not only contains errors but also warnings and informative messages. This makes the `E043` run even if there are no errors but only warnings exceeding the count defined in `maxerr` config.

This PR will modify the condition for running into `E043`, by ignoring all warnings and other messages and considering only errors.

This also updates the `maxerr` documentation to reflect the above change.

The tests changed to accommodate for the new behavior are:

1.  `core.js:testRawOnError`

    "Too many errors." is no longer being thrown because "Unnecessary semicolon" is a warning (`W032`)

2. `core.js:insideEval`
    Similar case of "Too many errors." not triggering (`maxerr` is set to 1 here) as "eval can be harmful." is a warning (`W061`)

3. `parser.js:restOperatorWithoutIdentifier`
    We run into "Too many errors." at `9:22` now instead of `8:30` because the warnings in between are no longer considered.

This PR closes #3444

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -424,9 +424,9 @@ var JSHINT = (function() {
     removeIgnoredMessages();
 
     var errors = JSHINT.errors.filter(function(e) { return /E\d{3}/.test(e.code); });
-    if (errors.length >= state.option.maxerr)
+    if (errors.length >= state.option.maxerr) {
       quit("E043", t);
-
+    }
     return w;
   }
 
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -423,7 +423,8 @@ var JSHINT = (function() {
 
     removeIgnoredMessages();
 
-    if (JSHINT.errors.length >= state.option.maxerr)
+    var errors = JSHINT.errors.filter(function(e) { return /E\d{3}/.test(e.code); });
+    if (errors.length >= state.option.maxerr)
       quit("E043", t);
 
     return w;
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -421,10 +421,19 @@ exports.argsInCatchReused = function (test) {
   test.done();
 };
 
-exports.testRawOnError = function (test) {
+exports.testRawOnWarning = function (test) {
   JSHINT(';', { maxerr: 1 });
   test.equal(JSHINT.errors[0].raw, 'Unnecessary semicolon.');
   test.equal(JSHINT.errors[1], null);
+
+  test.done();
+};
+
+exports.testRawOnError = function (test) {
+  JSHINT('@', { maxerr: 1 });
+  test.equal(JSHINT.errors[0].raw, 'Unexpected \'{a}\'.');
+  test.equal(JSHINT.errors[1].raw, 'Too many errors.');
+  test.equal(JSHINT.errors[2], null);
 
   test.done();
 };
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -424,8 +424,7 @@ exports.argsInCatchReused = function (test) {
 exports.testRawOnError = function (test) {
   JSHINT(';', { maxerr: 1 });
   test.equal(JSHINT.errors[0].raw, 'Unnecessary semicolon.');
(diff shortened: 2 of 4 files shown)


tests/unit/core.js
```diff
@@ -424,8 +424,7 @@ exports.argsInCatchReused = function (test) {
 exports.testRawOnError = function (test) {
   JSHINT(';', { maxerr: 1 });
   test.equal(JSHINT.errors[0].raw, 'Unnecessary semicolon.');
-  test.equal(JSHINT.errors[1].raw, 'Too many errors.');
-  test.equal(JSHINT.errors[2], null);
+  test.equal(JSHINT.errors[1], null);
```
jugglinmike, line 427: This change may be sufficient to cause the test to pass, but it subvert the original intent. Can you add another test which verifies that when `maxErr` is set to `1`, JSHint reports "Too many errors" after encountering a single error?

src/jshint.js
```diff
@@ -423,7 +423,8 @@ var JSHINT = (function() {
 
     removeIgnoredMessages();
 
-    if (JSHINT.errors.length >= state.option.maxerr)
+    var errors = JSHINT.errors.filter(function(e) { return /E\d{3}/.test(e.code); });
+    if (errors.length >= state.option.maxerr)
```
jugglinmike, line 427: Mind adding braces around the IfStatement body?

jugglinmike (changes requested):
Thanks for the patch!

tests/unit/core.js
Ashesh3: You seem right, I overlooked the original intent of this test. I believe replacing the semicolon warning with an actual error like "Unrecoverable syntax error" (if the input is `}`) would help the test retain it's original intent. 
Would replacing `;` with `}` here and adding back the test for "Too many errors." be sufficient?

src/jshint.js
Ashesh3: Yep. Done. I actually did read up your coding style in contribution guidelines but missed this one out because the previous if statement (before this commit) had no braces. My bad.

tests/unit/core.js
jugglinmike: Let's do both

Ashesh3: Modified the test to reflect it's original indented behavior and added another test to accommodate for this PR change. Let me know if it looks good!

jugglinmike (approved):
Looks good; thank you!

BillyWildHigh (commented):
gh pr checkout 3562

PR #3566, opened 2021-08-26
[[FIX]] change escape-sequence handler for double quotes (\")

### LINKED ISSUE #3315: `"` and `\"` are considered as different keys
JsHint version: 2.9.6

Code:
```
var x = {
  '"': 10,
  '\"': 20,
};
```

Configuration: default

Result: no warnings

Expected result: W075: Duplicate key

maayami:
Changes the handling of escape sequence case for `\"` in the **lex.js** file. Earlier, it converted `\"` into `\\\"`, which is not the correct js equivalent for the case as it should be `"` only. This was causing issue #3315 as `\" != "` that is why jshint considers them different strings.

Fixes and Closes #3315

### DIFF
--- a/src/lex.js
+++ b/src/lex.js
@@ -1115,7 +1115,6 @@ Lexer.prototype = {
       char = "\\\\";
       break;
     case "\"":
-      char = "\\\"";
       break;
     case "/":
       break;
--- a/src/lex.js
+++ b/src/lex.js
@@ -1114,8 +1114,6 @@ Lexer.prototype = {
     case "\\":
       char = "\\\\";
       break;
-    case "\"":
-      break;
     case "/":
       break;
     case "":
--- a/tests/unit/core.js
+++ b/tests/unit/core.js
@@ -2314,6 +2314,17 @@ exports.duplicateProto = function (test) {
 
   src = [
     "void {",
+    "  '\"': null,",
+    "  \"\\\"\": null",
+    "};"
+  ];
+
+  TestRun(test, "Duplicate keys (data)")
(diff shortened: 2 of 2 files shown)


src/lex.js
```diff
@@ -1115,7 +1115,6 @@ Lexer.prototype = {
       char = "\\\\";
       break;
     case "\"":
-      char = "\\\"";
```
jugglinmike, line 1118: Since this "case" clause no longer has any purpose, we can remove it entirely.

jugglinmike (changes requested):
Thanks for the patch! Could you add a unit test which demonstrates its correctness? The file `tests/unit/core.js` seems like a good place for this. We could use a test case that fails in the current release of JSHint but passes with this patch applied. That will prove that this patch has the intended effect, and it will help us avoid accidentally breaking the new and improved behavior in the future.

maayami:
Sure @jugglinmike  I'll add a unit test for this.
Also, could you please help me in running the jshint clone on the local development server.

src/lex.js
maayami: Okay will do that and will also check for any breaking changes after that.

maayami:
Hey @jugglinmike, I have added a unit test in the commit `849e08d` that fails for the latest version of JSHint but passes with this path applied.

src/lex.js
maayami: Done, I have made the changes suggested by you in the commit `
cf99223 `.

jugglinmike (approved):

jugglinmike:
Beautiful! Thanks for the help!

PR #3568, opened 2021-08-31
new identifiers generation file added

ayabhishek:
Issue number 3495 resolved. Created a coded sequence for new identifiers and browser settings covering the earlier uncovered globals. Used earlier approach too with new coded combination to complete the same.


jugglinmike:
I don't understand how this change is related to gh-3495; could you say more?

jugglinmike:
We can't move forward on this patch without more information. I'm going to close it to keep things organized, but I'll be happy to re-open whenever you have a moment to help me understand!

Issue #3567, opened 2021-08-29
URL should be a global when `node: true`


coolaj86:
```js
var url = new URL("https://example.com?" + resp.toJSON().body);
```

```txt
[jshint] URL is not defined
```

Node has `URL`, the same as browsers.

jugglinmike:
Sounds good to me! Care to send a patch our way?

coolaj86:
sent

jugglinmike:
Great! Now that your patch is merged, the next release of JSHint will recognize URL for Node.js environments.

PR #3570, opened 2021-09-12
bugfix: add URL for node in src/vars.js

coolaj86:
Re: https://github.com/jshint/jshint/issues/3567#issuecomment-917540451

See history in https://nodejs.org/api/url.html#url_class_url.


jugglinmike (approved):
Thank you!

PR #3581, opened 2021-11-06
Drop `exit` dependencie

lauriro:
According to nodejs/node-v0.x-archive#3584
this was fixed with nodejs/node-v0.x-archive@20176a9 in 2014.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
   "dependencies": {
     "cli": "~1.0.0",
     "console-browserify": "1.1.x",
-    "exit": "0.1.x",
     "htmlparser2": "3.8.x",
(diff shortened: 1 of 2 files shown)


jugglinmike:
Ah, very nice! This patch will interrupt JSHint's support for Node.js 0.10. That is admittedly a very old release of the platform, but I'm still reluctant to make a breaking change outside of a major release. No ETA on that yet, but given that removing the workaround module is a matter of housekeeping (rather than correctness), I think it's fine to delay this.

In the mean time, [I've proposed that the `node-exit` module document the current state of affairs](https://github.com/cowboy/node-exit/pull/16) so that others can benefit from your discovery.

PR #3582, opened 2021-11-14
Missing well-known globals

acarstoiu:
Hello,

I have no idea why this hasn't been noticed before, but some well-known globals are missing from `vars.js`. Here's what I've added:
- _BigInt_ for ES 11
- _Intl_ for node
- _specify_ and it's counterpart _xspecify_ for mocha
- _URLSearchParams_ for browser and node

This change is really easy to merge and I'm looking forward to seeing it in the next release.

### DIFF
--- a/src/vars.js
+++ b/src/vars.js
@@ -422,6 +422,7 @@ exports.browser = {
   TimeEvent            : false,
   top                  : false,
   URL                  : false,
+  URLSearchParams      : false,
   WebGLActiveInfo      : false,
   WebGLBuffer          : false,
   WebGLContextEvent    : false,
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks!

src/vars.js
```diff
@@ -505,21 +509,23 @@ exports.node = {
...
+  exports        : true,
+  process        : true,
+  setTimeout     : true,
+  clearTimeout   : true,
+  setInterval    : true,
+  clearInterval  : true,
+  setImmediate   : true, // v0.9.1+
+  clearImmediate : true, // v0.9.1+
+  URL            : true,  // v6.13.0+
+  URLSearchParams: true  // v6.13.0+
```
coolaj86, line 528: Not that it matters too much, but URLSearchParams was introduced in v10.0.0+

acarstoiu: Damn, I really checked the history section of [URLSearchParams](https://nodejs.org/api/url.html#class-urlsearchparams). I saw it was **introduced** in _v6.13.0_, but didn't notice that it became a **global variable** only in _v10.0.0_.

acarstoiu:
(!) **There's a problem:** I've just tested with the latest release _2.13.2_ and in spite of these lines
``` javascript
  11: {
    BigInt             : false
  }
```
it does not work, `BigInt` is reported as "not defined".

src/vars.js
jugglinmike: @coolaj86 Accurate documentation matters plenty! Thanks for checking our work :) Fixed via commit 06accfab74385e33ece831dc877cbafb20d9b515

PR #3588, opened 2021-12-15
fix: add URLSearchParams for node in src/vars.js

coolaj86:
Follow-up to https://github.com/jshint/jshint/pull/3570

See history at: https://nodejs.org/dist/latest-v16.x/docs/api/url.html#class-urlsearchparams

Also adjusted whitespace for consistency.


jugglinmike:
Thanks for the patch! [Another contributor spotted the omission a few weeks earlier](https://github.com/jshint/jshint/pull/3582), so I've resolved this issue by merging their patch.

PR #3596, opened 2021-12-31
[[DOCS]] Remove david-dm badges

adamchainz:
The site is down, from archive.org it seems since October or November: https://web.archive.org/web/*/david-dm.org . The [last snapshot](https://web.archive.org/web/20211002064243/https://david-dm.org/) shows the most announcement was 2016, so it seems the service has not been maintained in a while.


jugglinmike:
Thanks for the research and for taking the time to fix this!

Issue #3599, opened 2022-01-15
[security] upgrade shelljs version to 0.8.5


nodesocket:
There is a known high security impact exploit in the version of shelljs used https://github.com/jshint/jshint/blob/main/package.json#L49.

See the following [huntr.dev](https://huntr.dev/bounties/50996581-c08e-4eed-a90e-c0bac082679c/) for full details and the official [GitHub issue from shelljs](https://github.com/shelljs/shelljs/issues/1058).

Fix should just be pinning to version `0.8.5` and making sure the api did not change and break anything after the upgrade.

nfischer:
:wave: ShellJS maintainer here!

There's some significant API and behavior changes between 0.3.x and 0.8.5, so this particular project will need to exercise some caution when upgrading (but I still recommend that as the path forward). https://github.com/shelljs/shelljs/wiki/Breaking-Changes is a good place to start. This doesn't cover all the way back to 0.3.x, but it should help you get through the last few versions. IIRC there weren't too many major API changes between v0.7 and v0.8 (which is probably why I didn't write a migration guide for that version).

Let me know how I can help.

jugglinmike:
Thanks for the heads-up, Justin and Nate! This project's coupling with shelljs extends beyond the CLI code: [the automated tests perform a lot of mocking in order to validate behavior](https://github.com/jshint/jshint/blob/2a842ac5891916d93356ada81d7cd73918d4f443/tests/cli.js).

I haven't looked in to what it will take to upgrade versions yet, but this may be a good time to consider alternatives. JSHint's support for Rhino was broken years ago, partly because it differs significantly from what's available in Node.js. [Since there's still interest in using JSHint in that context](https://github.com/jshint/jshint/issues/3587), I want to cast a wide net for solutions.

nodesocket:
@jugglinmike I can appreciate upgrading shelljs breaking the automated tests and requiring some work and effort. However, @nfischer was nice enough to offer the [migration guide](https://github.com/shelljs/shelljs/wiki/Breaking-Changes) which should get us most of the way there.

GitHub is flagging `jshint` as having a security vulnerability because of its dependency on shelljs, so I'd argue it's worth the effort to upgrade shelljs to 0.8.5.

![Uploading Screen Shot 2022-01-17 at 6.41.12 PM.png…]()

nfischer:
[...]

PR #3603, opened 2022-01-18
Remove shelljs

jugglinmike:
Resolves gh-3599


jugglinmike:
@rwaldron TravisCI isn't running at the moment. I've contacted their support staff; let's hold off on merging this until I hear back. (I've verified all the Node.js API's referenced by this patch were available in 0.10, but I'd like to see the results to be certain.)

src/cli.js
```diff
@@ -90,7 +89,7 @@ function findConfig(file) {
   else if (envs) {
     home = path.normalize(path.join(envs, ".jshintrc"));
 
-    if (shjs.test("-e", home))
+    if (fsUtils.exists(home))
```
rwaldron, line 92: Why not use  `fs.existsSync()`?

jugglinmike: This method is backed by `fs.existsSync`. The reason not to use the Node.js method directly is so that we can more easily substitute in an implementation designed for the Rhino runtime (I'd like to restore support for Rhino one day).

jugglinmike:
@rwaldron CircleCI says it all passes in old Node.js! I had to merge `main` into this branch, though, so I'd like to rebase this branch before we merge it. (I'll also change `[[CHORE]] Remove shelljs` to `[[FIX]] Remove shelljs` so that it shows up in the change log.)

jugglinmike:
Thanks, @rwaldron. Since I didn't get the chance to update the commit message prior to merging, I've forcibly pushed [an amended version of the relevant commit](https://github.com/jshint/jshint/commit/eb4609a5ea3024ed54167c7b8406ce30dd940d84) to the project's main development branch. I typically try to avoid this, but I'm hoping that this comment (along with a note in the commit message itself) minimizes the disruption that generally comes from revising the history like this.

PR #3623, opened 2022-08-08
[[FIX]] Cannot read properties of undefined (reading value) when expr…

Lokya:
if write code on https://jshint.com like this: 
`function main() {
  const {...} ; // this

  return 'Hello,  World!'; 
}`
chrome：
Uncaught TypeError: Cannot read properties of undefined (reading 'value')

### DIFF
--- a/src/jshint.js
+++ b/src/jshint.js
@@ -4313,7 +4313,7 @@ var JSHINT = (function() {
             id = identifier(context);
           } else {
             expr = expression(context, 10);
-            error("E030", expr, expr.value);
+            error("E030", expr, expr && expr.value);
           }
         } else {
           id = identifier(context);


jugglinmike:
Thanks for the patch! As per [JSHint's contribution guidelines](https://jshint.com/contribute/), we'll need to add at least one automated test before we can merge this. Are you able to add a test?

Lokya:
> Thanks for the patch! As per [JSHint's contribution guidelines](https://jshint.com/contribute/), we'll need to add at least one automated test before we can merge this. Are you able to add a test?

Lokya:
sorry，wrong operation close this comment.
this is my automated test
[Coverage Status](https://coveralls.io/github/Lokya/jshint?branch=fix-exprValue)

Coverage remained the same at 100.0%

jugglinmike:
That appears to be the results of the project's automated test suite when run against your original patch. That's not what I had in mind.

In your write-up for this pull request, you shared a example of JavaScript source code which causes JSHint to crash. What we need is a new test which demonstrates what JSHint does when it receives code like that. This will look like a little bit of code added to one of the files in this repository's `tests/unit/` directory. That's where all the other unit tests are stored, so you can take a look around for hints about how to write a new test.

Issue #3632, opened 2022-11-06
W080 incorrect message for initializing const with undefined


qdirks:
```
const UNKNOWN = undefined; // W080 on this line
```
jshint advises that it's not necessary to initialize const variables with undefined. Taking the initializer away causes a syntax error. Thanks! :)

jugglinmike:
Thanks for the report! That's definitely a bug. I've submitted a patch in gh-3633.

(This also highlighted a kind of wonky error message, so I've suggested an improvement in gh-3634.)

jugglinmike:
We released the fix today in [version 2.13.6](https://jshint.com/blog/2022-11-11/release-2-13-6/)!

PR #3633, opened 2022-11-07
[[FIX]] Allow initializing const bindings to undef

jugglinmike:
Warning W080 is inappropriate for constant bindings because, unlike `var` and `let` declarations, `const` requires an initializer.

Relax the emission of W080 to allow initializing constant bindings with `undefined`. Add unit tests to verify that the warning continues to be emitted for constant bindings declared in ES2015 destructuring patterns.


jugglinmike:
This resolves gh-3632,

rwaldron (approved):

Issue #3648, opened 2023-08-08
No warnings about undefined variable when doing typeof yyy


dvhx:
```
function a() {
  return xxx;		// undefined variable, as expected
}

function b() {
  return typeof yyy;    // no warnings
}

a();
b();
```
Shouldn't it warn about yyy being undeclared?

PR #3656, opened 2024-10-03
Fix: typeof operator warning for undeclared variables

akshay-092:
#### Issue
In the current implementation of JSHint, the `typeof` operator can be used on undeclared variables without raising a warning, even when the `undef` option is enabled. This behavior can lead to potential issues when developers unknowingly reference variables that are not declared.

#### Solution
This PR adds a new warning (`W149`) to notify developers when they use the `typeof` operator with an undeclared variable. The change preserves the valid use case of `typeof` for unresolvable references by allowing the operation to return `'undefined'` without throwing an error, while still issuing a warning when the variable has not been declared in the current scope.

#### Key Changes:
- Modified the `prefix("typeof")` function to check if the operand is an undeclared variable and trigger a warning.
  
#### Example:
The following code will now trigger a warning:

```javascript
typeof undeclaredVar !== 'undefined';  // Warning: W149 - undeclared variable
```


jugglinmike:
Hi there, @akshay-092, thanks for the contribution! Could you say a little bit more about the kinds of issues the current behavior can cause?

akshay-092:
**Issue: No warnings about undefined variable when doing `typeof yyy` (#3648)**

Hi there! Thanks for the opportunity to contribute!

The main issue with the current behavior, where `typeof yyy` doesn't trigger a warning for undeclared variables, is that it can cause silent bugs. In JavaScript, using `typeof` on an undeclared variable is valid and returns `'undefined'`. However, this can lead to a situation where developers unknowingly use undeclared or misspelled variables, and because there's no warning, these issues can slip through unnoticed.

For example:
```javascript
if (typeof myVar === 'undefined') {
  console.log("myVar is undefined");
}
```
In this case, if `myVar` was never declared, `typeof` will return `'undefined'` without any error or warning. The developer might assume that `myVar` was declared elsewhere, leading to potential confusion or bugs. This is especially problematic in larger codebases where it's easy to overlook these silent failures.

By adding a warning in JSHint when `typeof` is used with an undeclared variable, it would help developers catch these errors early, ensuring that all variables are properly declared and reducing the risk of bugs caused by typos or incorrect assumptions about variable scope.

This change would help developers avoid subtle bugs and improve code quality by ensuring that undeclared variables don't go unnoticed in these situations.

jugglinmike:
I'm not sure that the pattern `typeof anyUndeclaredBinding` is itself a problem because the semantics of the `typeof` operator make it inherently safe, and because JSHint will warn about any subsequent references that require resolution (if it doesn't please file a bug about that!). Indeed, many people use `typeof` precisely for this semantic (for instance, to determine if some third-party code has executed and defined a global binding), so a warning like this would be somewhat disruptive for them.

Are the silent bugs you mentioned different from references to undeclared variables?

PR #3668, opened 2025-10-24
fix: add missing browser globals (customElements, ResizeObserver, etc.)

Nik1510:
Fixes #3522

Adds missing browser globals to `exports.browser` in `src/vars.js`, ensuring that enabling `browser: true` in .jshintrc correctly recognizes:

- customElements
- ResizeObserver
- AbortController
- DOMException
- crypto

Includes a test in `tests/unit/options.js` verifying that no warnings occur for these globals .


Nik1510:
This PR only updates core JSHint globals (customElements, ResizeObserver, etc.). All core and unit tests pass. The failing website tests are unrelated to this change and shouldn’t block merging.

